Issue #197, opened 2026-06-24
App Review follow-up: microphone permission wording + next keyboard button


Pivii:
## Context

Apple App Review follow-up for the first public App Store submission of Dictus iOS.

This issue tracks the current review blocker(s) and keeps the review history in one place, instead of scattering it across Telegram/App Store Connect messages.

Related issues:
- #189 — App Store listing for v1.7.1 / first submission prep
- #194 — Previous 5.1.1(iv) rejection about microphone permission button wording; closed after identifying the `Allow microphone` → `Continue` fix

### Previous review round — Guideline 5.1.1(iv)

Apple rejected the app because the microphone permission pre-prompt used wording that directed users toward granting permission:

> The app encourages or directs users to allow the app to access the microphone. Specifically: a custom message appears before the permission request, and to proceed users press an "Allow microphone" button. Use words like "Continue" or "Next" on the button instead.

Issue #194 identified the likely offending UI:

- `DictusApp/Onboarding/MicPermissionPage.swift`
- Button text: `Allow microphone`
- Required neutral replacement: `Continue` / `Next`

The earlier third-party AI / data-collection concern was not raised again after the privacy clarification that Dictus runs on-device.

### Current review round — latest Apple response

Apple acknowledged the previous response, but still reports the same microphone permission flow issue, and added a new keyboard-extension issue.

Full relevant Apple response:

> Hello,
>
> Thank you for your response and information provided. We appreciate your efforts to comply with the App Review Guidelines.
>
> Regarding Guideline 5.1.1, we still found that a custom message appears before the permission request, and to proceed users press a "Allow microphone" button.
>
> To resolve this issue, it would be appropriate to revise the permission request process in the app to not display messages before the permission request with inappropriate words on buttons. You may use words like "Continue" or "Next" on the button instead.
>
> Upon further review, we've identified additional issues that require your attention:
>
> Guideline 4.4.1 - Design - Extensions
>
> Issue Description
>
> Your keyboard extension does not provide a way for users to switch to another keyboard.
>
> Next Steps
>
> Please implement a next keyboard button that enables users to advance to another keyboard.
>
> To ask the system to switch to another keyboard, call the advanceToNextInputMode method of the UIInputViewController class. The system picks the appropriate "next" keyboard; there is no API to obtain a list of enabled keyboards or for picking a particular keyboard to switch to.
>
[...]

Issue #125, opened 2026-04-20
App is not yet properly optimized for iPad


Pivii:
## Summary

Dictus iOS is currently not properly optimized for iPad usage.

## User feedback

A first quick test on iPad suggests that:
- the app can be installed and launched
- onboarding appears to work correctly
- but the application as a whole does not yet feel properly adapted / optimized for iPad

## Current conclusion

iPad support appears to be at least partially functional, but the app should currently be treated as **not properly optimized for iPad**.

## Areas to investigate

- layout and sizing on larger screens
- onboarding flow on iPad
- settings and model management UI on iPad
- keyboard-related UX assumptions that may be iPhone-specific
- split view / multitasking behavior
- portrait vs landscape behavior
- whether some views are technically working but visually / ergonomically poor on iPad

## Why this matters

Even if iPad is not a priority today, users can already try installing Dictus there. So we should either:
- improve the iPad experience,
- or explicitly scope/support it later with a known limitation.

## Notes

Initial user impression:
- onboarding works well
- the rest of the app does not yet feel iPad-ready

Pivii:
> *This was generated by AI during triage.*

## Human Brief

**Category:** enhancement
**Summary:** Audit and polish the iPad experience across the whole app (excluding the keyboard extension, which is iPhone-shaped by design).

**Current behavior:**
The app installs and launches on iPad. Onboarding works. The rest of the app is iPhone-sized layouts stretched onto iPad with no iPad-specific design — feels visually and ergonomically unfinished.

**Desired behavior:**
The app should feel native on iPad: appropriate layouts on larger screens, sensible portrait + landscape behavior, no iPhone-only assumptions in settings / model management, working split-view / multitasking.

**Suggested scope (to be broken down before pickup):**
- Layout audit on iPad portrait + landscape (settings, model manager, onboarding, debug screens).
- Decide whether to support split view / Stage Manager / multitasking.
- Decide whether the keyboard extension itself needs iPad-specific layout (currently sized for iPhone).
- Decide whether to ship as an iPad-optimized binary or keep it as "iPhone app on iPad" with known limitations communicated in the App Store description.

**Acceptance criteria:**
- [ ] Audit document listing every screen and its iPad behavior (works / broken / acceptable but not polished).
- [ ] Product decision recorded: optimize fully, optimize selectively, or scope out with App Store messaging.
- [ ] If optimize: specific UI tasks broken into separate child issues with priorities.

**Why human, not agent:**
[...]

PR #198, opened 2026-07-02
fix(#197): next-keyboard globe (4.4.1) + Apple-matched key sizes

Pivii:
## Context

App Review follow-up for #197 (first App Store submission, v1.7.1).

**Guideline 4.4.1** — the custom keyboard offered no way to switch to another keyboard. On recent iPhones iOS draws its own globe below the keyboard (`needsInputModeSwitchKey == false`), which hid the problem; on older iPhones (home-button, e.g. SE/8) `needsInputModeSwitchKey == true` and there was no switch control at all. (App is iPhone-only — `TARGETED_DEVICE_FAMILY = 1` — so Apple tests on iPhone; native iPad is out of scope, tracked in #125.)

While in the keyboard, also fixed key-size inconsistencies vs Apple's keyboard reported during device testing.

## Changes (3 commits)

1. **`consistent special-key sizes`** — numbers/symbols row 3 totalled only 8.0 units (vs 10.0 on letter pages), stretching `#+=` and delete. Centered with spacers so the row totals 10.0.
2. **`next-keyboard button (4.4.1)`** — thread `UIInputViewController.needsInputModeSwitchKey` into the layout and prepend a `.keyboard` globe key (already wired to `advanceToNextInputMode()` + VoiceOver) left of 123/ABC **only when the system does not provide its own**. The space bar shrinks 1.0 to keep the row at 10 units. Recent iPhones are unchanged.
3. **`match Apple key widths`** — unified a single `bottomRow` for every page: emoji now present on all pages, 123/ABC and emoji share a small 1.5 width, only the space bar changes width — so nothing but the label moves on the 123↔ABC toggle. Punctuation keys on row 3 widened (1.2) to sit tighter like Apple.

Only `DictusKeyboard/KeyboardLayouts.swift` (layout data) and `DictusKeyboard/KeyboardViewController.swift` (pass the flag) are touched. Every row totals 10 units, with and without the globe.

## Testing

- Device (iPhone, recent): AZERTY/QWERTY typing, shift/caps, accent long-press, delete/word-delete, space + trackpad cursor, return, layer toggles ABC↔123↔#+=, emoji picker from both ABC and 123, dictation mic — all validated. No globe (system provides it) ✓ and nothing resizes on layer toggle ✓.
- Build: `DictusKeyboard` scheme builds clean (simulator).

## ⚠️ Residual risk before resubmission

The globe **render path** (`needsInputModeSwitchKey == true`) has not been visually confirmed on a device/simulator where the flag is true (all local testing was on a recent iPhone where it is false). The code follows Apple's documented rule exactly and reuses the pre-existing, proven `.keyboard` key, so confidence is high — but a quick check on an **iPhone SE (home-button) simulator** is recommended before resubmitting to App Review.

## Notes

- 5.1.1 mic-wording is already correct in build 18 (`7264b1d`, "Continue"); the earlier rejection matched build 17. No code change needed here — to be answered to Apple separately.
- iPad keyboard not displaying in the simulator is the iPhone-only-app compatibility mode, unrelated to this PR (see #125).

### DIFF
--- a/DictusKeyboard/KeyboardLayouts.swift
+++ b/DictusKeyboard/KeyboardLayouts.swift
@@ -18,7 +18,7 @@ enum KeyboardLayouts {
     // MARK: - Public API
 
     /// AZERTY layout (default for French).
-    static func azerty(lang: SupportedLanguage = .active) -> KeyboardDefinition {
+    static func azerty(lang: SupportedLanguage = .active, needsGlobe: Bool) -> KeyboardDefinition {
         return KeyboardDefinition(
             name: lang.displayName,
             locale: lang.rawValue,
@@ -26,16 +26,16 @@ enum KeyboardLayouts {
             returnName: lang.returnName,
             longPress: longPressData,
             layout: KeyboardDefinition.Layout(
-                normal: azertyNormal(lang: lang),
-                shifted: azertyShifted(lang: lang),
-                symbols1: numbersPage(lang: lang),
-                symbols2: symbolsPage(lang: lang)
+                normal: azertyNormal(lang: lang, needsGlobe: needsGlobe),
+                shifted: azertyShifted(lang: lang, needsGlobe: needsGlobe),
+                symbols1: numbersPage(lang: lang, needsGlobe: needsGlobe),
+                symbols2: symbolsPage(lang: lang, needsGlobe: needsGlobe)
             )
         )
     }
 
     /// QWERTY layout (default for English and Spanish).
-    static func qwerty(lang: SupportedLanguage = .active) -> KeyboardDefinition {
+    static func qwerty(lang: SupportedLanguage = .active, needsGlobe: Bool) -> KeyboardDefinition {
         return KeyboardDefinition(
             name: lang.displayName + (lang == .french ? " (QWERTY)" : ""),
             locale: lang.rawValue,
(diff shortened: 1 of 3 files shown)


Pivii:
## Review Hermes

Verdict: **changes requested**. I can’t submit a formal “request changes” review because GitHub considers this our own PR, so posting as a normal comment.

### Blocking

- **Globe key can advance multiple keyboards on one tap**

  Adding `.keyboard` to the layout activates the vendored `keyboardButtonExtraButton` path in `GiellaKeyboardView`. That overlay button is currently registered with:

  ```swift
  keyboardButtonExtraButton.addTarget(..., for: UIControl.Event.allEvents)
  ```

  Since the transparent `UIButton` sits over the globe cell, a single tap can fire `didTriggerKeyboardButton` for multiple control events (`touchDown`, `touchUpInside`, etc.). The result can be several calls to `advanceToNextInputMode()` from one user gesture, which risks skipping the intended next keyboard or making the App Review fix flaky on exactly the old-iPhone path this PR targets.

  Suggested fix: change the button target to a single event, typically `.touchUpInside` (or `.primaryActionTriggered` if you want accessibility/keyboard activation semantics), then test the `needsInputModeSwitchKey == true` path.

### Checked

- PR diff reviewed locally against `origin/main`.
- GitHub checks are green: **Build (iOS Simulator)** and **SwiftLint**.
- Row width totals verified manually: bottom rows and punctuation rows remain at 10 units with and without globe.
- No obvious security/secrets/debug leftovers in the diff.

After the button event fix, this should be safe to merge pending the recommended visual check on an iPhone SE/home-button simulator or device.

Pivii:
Fixed in edb57e0. Changed the `keyboardButtonExtraButton` registration from `UIControl.Event.allEvents` to `.touchUpInside`, so the globe now calls `advanceToNextInputMode()` exactly once per tap. Build stays green.

Still recommend the visual check on an iPhone SE / home-button simulator (`needsInputModeSwitchKey == true`) before resubmitting — will run it before submission.

Issue #141, opened 2026-05-02
Local LLM transcription polish — light corrections while staying faithful (free tier, distinct from #79 Smart Mode)


Pivii:
> *This was generated by AI during triage.*

## Summary

Add a local-LLM post-processing pass on top of raw STT output that performs **light, faithful polish only** — fixes obvious typos, missing punctuation, capitalization, missing accents, and similar low-risk cleanups — while preserving the user's actual words and meaning.

This is intended for the **free tier**, applied transparently after every transcription if enabled.

## Why this is distinct from #79 Smart Mode

| | This issue (free polish) | #79 Smart Mode (premium) |
| --- | --- | --- |
| Goal | Clean up STT output | Transform / rewrite text |
| Faithfulness | Must preserve user's exact words | Free to rephrase, restructure |
| Examples | `bonjour ca va` → `Bonjour, ça va ?` | `bonjour ca va` → `Hi, hope you're doing well!` |
| Tier | Free | Premium |
| Default | On (probably opt-out per language) | Off (user-triggered mode) |
| Latency target | Imperceptible (sub-200ms ideal) | Visible processing acceptable |

The faithfulness contract is the **defining boundary** between the two issues. If a polish change alters the user's meaning, words, or intent, it belongs in #79, not here.

## Why this matters now

Several open issues block on this capability:

- **#90 Parakeet language** — could use a local LLM to translate/force language post-transcription, since Parakeet auto-detects and ignores the language picker.
- **#115 Verbal punctuation** — STT models don't reliably interpret "comma" / "period" / "new line" as punctuation commands; a polish LLM could detect and apply these.
- **#109 German autocorrect** — community feedback explicitly suggested using local Apple Intelligence to post-correct German text where the keyboard's autocorrect is weak.

Without this layer, those issues remain blocked or require model-specific workarounds.

## Backend candidates (to evaluate)

- **Apple Foundation Models** (~3B, iOS 18.1+) — pre-loaded by the system, near-zero RAM cost when iOS already keeps them warm for notification summaries. Best latency. Limited to recent iOS versions and recent devices.
- **Open-source small models** (Phi-3 mini, Gemma 2 2B, Qwen2.5 0.5B, etc.) — broader device support but RAM/disk cost on top of WhisperKit/Parakeet. Need quantization (Q4_K_M or smaller).
- **MLX vs Core ML vs llama.cpp** — to compare for Dictus's existing model-loading patterns.

## Open product questions (to be decided before agent brief)

[...]

PR #228, opened 2026-07-25
Onboarding: replace layer choice with local transcription polish toggle

### LINKED ISSUE #213: Onboarding: replace layer choice with local transcription polish toggle
## Summary

Update the iOS onboarding flow to remove the current **language layer / mode choice** step (`1 / 2 / 3` or `ABC`) and replace it with a simple **Transcription Polish** toggle.

The product direction is now:

- keep Dictus focused on **keyboard-only mode** for now;
- avoid exposing layer/mode complexity during onboarding;
- introduce users early to the new local post-transcription polish capability;
- let users opt in/out from onboarding with a clear explanation of the latency/quality tradeoff.

## Product context

Pierre clarified that the post-transcription local LLM feature should be called **"transcription polish"** / **"polish de transcription"**.

This is the local Apple Foundation Models pass that lightly improves the raw transcription after STT. It should improve quality while remaining faithful to what the user said.

Related capability issue: #141.

## Current problem

The onboarding currently includes a choice around language/layer/mode (`1 / 2 / 3` or `ABC`). This is probably too complex and no longer aligned with the immediate product focus.

For now, Dictus should present itself primarily as a keyboard dictation app, not as a product where users must understand several layers/modes before they can start.

## Proposed UX change

Replace the layer choice step with a small section/toggle like:

```text
Polish transcription with a local model

Dictus can use Apple's local Foundation Model to lightly polish your transcription after speech recognition: punctuation, casing, accents and small corrections.

This can improve transcription quality, but may add a little processing time.

[ Toggle ] Enable transcription polish
```

French draft:

```text
Polir la transcription avec un modèle local

Dictus peut utiliser le modèle local d’Apple pour améliorer légèrement votre transcription après la reconnaissance vocale : ponctuation, majuscules, accents et petites corrections.

Cela peut améliorer la qualité du texte, mais peut ajouter un peu de temps de traitement.

[ Toggle ] Activer le polish de transcription
```

Potential shorter label if UI space is tight:

```text
Améliorer la transcription avec un modèle local
```

## Behavior

- If enabled in onboarding, persist the existing transcription polish preference.
- If disabled, keep raw transcription behavior.
- The setting should remain editable later in Settings.
- If Apple Foundation Models are unavailable on the device/OS, the onboarding should either:
  - hide the toggle; or
  - show it disabled with a short explanation.
- Do not introduce a model picker in onboarding.
- Do not introduce Smart Mode / rewrite UX here; this is faithful polish only.

## Copy guardrails

The copy should make these points clear:

- everything stays local/on-device;
- this is optional;
- this can improve output quality;
- this may add some processing time;
- this is not cloud AI and not a rewrite mode.

Avoid overpromising. Prefer “can improve” / “may add” over guaranteed claims.

## Acceptance criteria

- [ ] The old layer/mode choice step is removed from onboarding.
- [ ] Onboarding includes a transcription polish toggle or equivalent choice.
- [ ] The toggle uses clear copy explaining local Apple Foundation Models, quality improvement and possible added processing time.
- [ ] The selected value persists to the same preference used by the polish pipeline/settings.
- [ ] The setting remains changeable after onboarding.
- [ ] Unsupported devices/OS versions do not show a broken or misleading toggle.
- [ ] No cloud fallback is suggested or used.
- [ ] This remains distinct from Smart Mode / rewrite / premium transformations.

## Notes

This is an onboarding/product UX issue, not the core LLM polish implementation itself. The underlying local polish feature remains tracked in #141.

Pivii:
Closes #213

## Summary

Replaces the onboarding layer-choice step (ABC / 123 picker) with an opt-in **transcription polish** toggle, per the product direction in #213.

- **Removed** `ModeSelectionPage` from onboarding. The default layer keeps its default value and the picker remains available in Settings.
- **Added** `PolishTogglePage` at the same position in the flow (Welcome → Mic → Keyboard → **Polish** → Model download → Globe tutorial). It explains the local Apple Foundation Models pass (punctuation, casing, accents, small corrections), the quality/latency tradeoff, and that everything runs on-device and stays editable in Settings.
- The toggle writes the existing shared preference `SharedKeys.polishEnabled` (App Group), the same key read by the polish pipeline (#141) and the Settings toggle. Default stays off (opt-in).
- **Availability gating** reuses `PolishAvailability`:
  - `isToggleVisible == false` (device not eligible, pre-iOS 26, SDK-less build): the page is skipped entirely and the step indicator shows 5 dots instead of 6.
  - Fixable states (Apple Intelligence off, model still downloading): the toggle is shown with a short explanation caption, mirroring the Settings footer. No dead toggle.
- **Stale index handling:** a user who was mid-onboarding on the old flow resumes cleanly (index 3 shows the polish page when supported, otherwise falls back to model download).
- All new copy localized in **English + French** in `DictusApp/Localizable.xcstrings`; the two orphaned layer-choice strings were removed.

No changes to the polish pipeline, the Settings layer picker, or the keyboard extension.

## How to test

1. Delete the app (or reset onboarding) and launch DictusApp on an iOS 26 device/simulator.
2. Go through onboarding: after the keyboard setup step you should see "Polish transcription with a local model" instead of the old ABC/123 choice. Verify the FR copy with the device in French.
3. Enable the toggle, finish onboarding, then open Settings: the "Polish transcription" toggle should be ON. Toggle it off in Settings and confirm it stays editable.
4. Redo onboarding leaving the toggle OFF: the Settings toggle should be OFF (default).
5. On a device/simulator where Apple Intelligence is off or the model is still downloading, the page should show the toggle with a short explanation caption.
6. On a pre-iOS 26 simulator (if available), the polish page should be skipped and the step indicator should show 5 dots with no gap in the flow.
7. Kill the app mid-onboarding (e.g. on the polish page) and relaunch: onboarding should resume at the right step with no page-index glitches.

Build validated: `xcodebuild -scheme DictusApp` for iOS Simulator (iPhone 17) succeeds (after the known `patch-fluidaudio-swift5.sh` workaround).


Pivii:

**Verdict: ✅ solid.** I traced the page-index logic across all four scenarios (polish supported/unsupported × fresh start/stale resume), checked the preference write against the Settings toggle and the polish pipeline, validated the xcstrings and pbxproj, and checked for semantic conflicts with PR #227. No blocking defects. Two low-severity notes below — both verified against the code, neither requires changes before merge.

### Verified correct

- **Page-index logic** (`DictusApp/Onboarding/OnboardingView.swift:62-91`, `:135-138`): all paths coherent. Supported: 0→1→2→3(polish)→4→5, identity dot mapping, 6 dots. Unsupported: 2 advances to 4, `displayStep` shifts pages >3 down one, 5 dots, no gap. Stale persisted index 3 on an unsupported device hits the plain `case 3` (the `where showsPolishPage` case is checked first) and shows ModelDownloadPage advancing to 5, and `displayStep(3) == 3` lands on the same dot as page 4 — consistent. No off-by-one, no out-of-bounds.
- **Preference write** (`PolishTogglePage.swift:31-32`): same key (`SharedKeys.polishEnabled`), same App Group suite, same `false` default as `SettingsView.swift:44-45` and as `PolishCoordinator.swift:78` (`defaults.bool(forKey:)` → false when unset). Opt-in default off holds even if the user never touches the toggle.
- **xcstrings**: valid JSON; all 8 new keys have `fr` translations in `state: translated` (file convention is `sourceLanguage: en` with the key as the EN text — matches the 215 other keys); both orphaned layer-choice keys removed; FR copy matches the issue #213 draft; no em dashes.
- **pbxproj**: old `AA0000E1`/`AA1000E1` (ModeSelectionPage) refs fully removed; new `AA0000EA`/`AA1000EA` unique, in the DictusApp Sources phase and Onboarding group only.
- **PR #227 conflict check**: #227 touched only `DictusKeyboard/` + `scripts/` — disjoint from this diff, no semantic overlap.
- Conventions: no force unwraps, comments in English, no retained observers/timers; `dictusGlass(in:)` and the `dictus*` fonts all exist.

### Findings (low severity, non-blocking)

**1. `OnboardingView.swift:38-45` — the "computed once at init" comment is inaccurate SwiftUI semantics, and the invariant holds for a different reason than stated.** A `let` stored property on a View struct is re-evaluated every time the struct is re-initialized, and `DictusApp.swift:95` observes `scenePhase` at the root, so `OnboardingView` (fullScreenCover content) is re-inited at least on every background/foreground cycle — `showsPolishPage` is re-read from `PolishAvailability.isToggleVisible` each time. It happens to be safe today because visibility can't actually flip at runtime: the three dynamic states (`.available`, `.appleIntelligenceNotEnabled`, `.modelNotReady`) all map to visible, and the invisible states (`deviceNotEligible`/`osTooOld`/`sdkMissing`) are static (`PolishAvailability.swift:70-81`). Failure scenario is future-facing: if a later change ever maps a dynamic state to invisible (e.g. a new `@unknown` reason from `SystemLanguageModel` landing in `.other`), the flow silently changes shape mid-onboarding — `totalSteps` flips 6↔5 and a user sitting on page 3 has the polish page swapped for ModelDownloadPage under their finger. Suggest fixing the comment (and, if you want the stated guarantee, capturing the value in `@State` via `init`), but no behavior bug exists today.

**2. `PolishTogglePage.swift:122-133` — `availabilityNote` doesn't self-refresh while the page stays foreground.** `PolishAvailability.state` is read in `body` with no observation, so the caption only updates when something else triggers a re-render. The common path is covered: leaving to iOS Settings to enable Apple Intelligence and returning re-renders via the root's `scenePhase` observation. The uncovered path: the Apple Intelligence model finishes downloading while the user sits on this page — the "still downloading" caption persists until the next re-render. Harmless (the toggle still writes the preference and the pipeline checks availability at runtime), and Settings has the same characteristic, so this is parity — just noting it since it answers "what if availability changes while onboarding is on screen".

Issue #144, opened 2026-05-04
Whisper turbo model: slow load + concurrent transcription cancellations


Pivii:
## Summary

Selecting the Whisper turbo model (\`openai_whisper-large-v3_turbo_954MB\`) makes dictation unreliable on physical iPhone 17 Pro / iOS 26.3.1:

- Compilation step is very long (~2 minutes after download).
- Time between \`modelSelected\` and \`appWhisperKitLoaded\` is sometimes 10+ minutes.
- When dictation is triggered before the load completes, \`transcriptionStarted\` fires multiple times in the same instant on the same model, and all attempts fail with \`Swift.CancellationError\`.
- User-visible symptoms reported by reporter: dictation \"doesn't translate\" (stays in French), no audio output, perceived slowness, and intermittent silent failures with no UI error.

## Repro

- iOS 26.3.1, iPhone 17 Pro (physical)
- App build: \`1.6.0 (13)\`, develop branch with #142 patches (unrelated to this issue)
- Settings → Model → select \`openai_whisper-large-v3_turbo_954MB\`
- Tap mic in keyboard, dictate

### Slow compilation / load (session ~14:21–14:40)

\`\`\`
14:21:10  modelDownloadStarted    name=...turbo_954MB size=0MB
14:21:49  modelDownloadCompleted
14:21:49  modelCompilationStarted
14:23:48  modelCompilationCompleted duration=119553ms        ← ~2 minutes
14:25:41  modelSelected           name=...turbo_954MB
14:40:07  appWhisperKitLoaded     model=...turbo_954MB        ← ~15 min after select
\`\`\`

### Concurrent transcription + CancellationError (session 16:02:44)

\`\`\`
16:00:37  modelSelected           name=...turbo_954MB
16:01:29  liveActivityFailed      rejectedTransition error=transcribing->recording
16:01:31  liveActivityFailed      rejectedTransition error=transcribing->transcribing
16:02:06  liveActivityFailed      rejectedTransition error=transcribing->recording
16:02:09  liveActivityFailed      rejectedTransition error=transcribing->transcribing
16:02:44  transcriptionStarted    model=...turbo_954MB
16:02:44  transcriptionStarted    model=...turbo_954MB        ← 3× same second
16:02:44  transcriptionStarted    model=...turbo_954MB
16:02:44  transcriptionFailed     error=Swift.CancellationError 1
16:02:44  transcriptionFailed     error=Swift.CancellationError 1
[...]

PR #229, opened 2026-07-25
Unify model downloads: byte-level progress, stall detection, and resume for WhisperKit models

### LINKED ISSUE #210: Unify model downloads: extend ParakeetModelDownloader to WhisperKit models (MB counter, stall detection, resume)
## Summary

PR #209 (issue #207) introduced `ParakeetModelDownloader` for Parakeet models: real byte-level progress, "X MB of Y MB" counter, 30s stall detection with retry/backoff, resumable downloads, localized error messages. WhisperKit models (Small, Small Quantized, Medium, Turbo) still use `WhisperKit.download()` and get none of these improvements. This issue is about generalizing the downloader so both engines share the same download experience.

## Current state per engine

| Capability | Parakeet (ParakeetModelDownloader) | WhisperKit (WhisperKit.download → HubApi) |
|---|---|---|
| Progress moves continuously | Yes (byte-level, delegate-based downloadTask) | Yes (HubApi reports within-file progress: `fileProgress.completedUnitCount = 100 * fractionDownloaded`, HubApi.swift:680) |
| Progress weighting | Byte-weighted (accurate %) | **File-count weighted** — every file counts the same regardless of size, so the % is distorted (a 5 MB config file moves the bar as much as a 300 MB weight blob) |
| "X MB of Y MB" counter | Yes (`ModelManager.downloadByteInfo`) | **No** — HubApi only exposes a unitless `Progress` fraction, no byte totals |
| Stall detection | 30s idle timeout (`timeoutIntervalForRequest`), NSURLErrorTimedOut → localized error | **None** — default URLSession behavior, a dead connection can hang for a long time with no user feedback |
| Retry with backoff | 3 attempts/file, 2s/4s/8s | **None** at our level (whatever HubApi does internally) |
| Resume after failure | Yes — completed files are skipped (atomic per-file moves) | Partial — HubApi skips existing complete files on re-download, but there is no stall/error path that invites a retry |
| Localized error messages | Yes (stall / rate-limit / HTTP / incomplete) | Generic `error.localizedDescription` |

Note: WhisperKit does NOT have the #207 freeze bug — its progress moves continuously. This is a UX-consistency and robustness improvement, not an App Review blocker. That is why it was kept out of the 1.7.1 hotfix.

## Proposed approach

1. Rename/generalize `DictusApp/Models/ParakeetModelDownloader.swift` into `ModelRepoDownloader` parameterized by:
   - `repoPath` (HF repo id): `FluidInference/parakeet-tdt-0.6b-v3-coreml` or `argmaxinc/whisperkit-coreml`
   - file-selection rule: Parakeet = the 4 required `.mlmodelc` dirs + root `*.json`/`*.txt`; WhisperKit = everything under the variant subfolder `{identifier}/` (e.g. `openai_whisper-small/`) — check against what `WhisperKit.download(variant:)` actually filters (it supports glob-ish variant matching) before assuming
   - destination directory: Parakeet = `AsrModels.defaultCacheDirectory(for: .v3)`; WhisperKit = `Documents/huggingface/models/argmaxinc/whisperkit-coreml/{identifier}` (the exact layout `WhisperKitConfig(modelFolder:)` and `deleteModel`/`cleanupModelFiles` already expect — reuse their path construction)
   - final-verification rule (Parakeet = modelsExist mirror; WhisperKit = variant folder non-empty + key files present)
2. In `ModelManager.downloadWhisperKitModel`, replace the `WhisperKit.download` call with the downloader (keep `WhisperKitConfig` prewarm step unchanged, pointing `modelFolder` at the downloaded directory).
3. Populate `downloadByteInfo` for Whisper models too — `ModelLoadingOverlay` and `ModelDownloadPage` already display it whenever it is set; no UI change needed.
4. Reuse the same `modelDownloadProgress` / `modelDownloadStalled` PersistentLog events.

## Why not in the 1.7.1 hotfix

- The rejection (2.1(a)) was specifically the Parakeet 4% freeze; Whisper progress already moves.
- Replacing the WhisperKit download path touches every Whisper variant (incl. Turbo, which has its own compile fragility, see #144/#104) and deserves a TestFlight cycle.
- Devices with <6 GB RAM get Whisper Small in onboarding and would exercise this new path on first install — higher blast radius than the hotfix should carry.

## Acceptance criteria

- [ ] Whisper model downloads show byte-accurate % and "X MB of Y MB", matching Parakeet.
- [ ] Stall (no bytes for 30s) on a Whisper download surfaces the localized error + retry within ~100s; retry skips already-downloaded files.
- [ ] Prewarm/compile still works from the downloaded folder for all Whisper variants (Small, Small Quantized, Medium, Turbo).
- [ ] Model delete still removes the files (path parity with existing cleanup code).
- [ ] Onboarding on a <6 GB device (iPhone SE class / older iPad) shows the improved progress for Whisper Small.
- [ ] Regression: Parakeet download unchanged.

## Pointers

- `DictusApp/Models/ParakeetModelDownloader.swift` — the downloader to generalize (file-selection + progress math are pure `internal static` functions, deliberately kept testable).
- `DictusApp/Models/ModelManager.swift` — `downloadWhisperKitModel` (lines ~162-285), `downloadParakeetModel`, `cleanupModelFiles`/`deleteModel` for the exact WhisperKit on-disk layout.
- HubApi reference (why no byte totals): swift-transformers `Sources/Hub/HubApi.swift:664-694`.
- Issue #207 + PR #209 for the full root-cause analysis and validation recipe (incl. the iPad simulator automation used).

Pivii:
Closes #210

## Summary

Generalizes the Parakeet downloader from PR #209 so WhisperKit model downloads get the same UX: byte-accurate progress, "X MB of Y MB" counter, 30s stall detection with retry/backoff, and resume after failure.

- `ParakeetModelDownloader` is renamed to `ModelRepoDownloader` (git mv, pbxproj refs updated in place). Engine-specific knowledge is parameterized in a `Configuration`: repo path, directory-pattern file selection, root-metadata inclusion, and required-paths final verification. Two factories: `.parakeet()` (byte-for-byte the previous behavior) and `.whisperKit(variant:)`.
- `ModelManager.downloadWhisperKitModel` replaces `WhisperKit.download()` with the shared downloader. The prewarm step is unchanged and keeps pointing `WhisperKitConfig(modelFolder:)` at the downloaded variant folder.
- `downloadByteInfo` is now populated for Whisper models; `ModelLoadingOverlay` and `ModelDownloadPage` already render it, so no UI changes.
- Same PersistentLog events reused (`modelDownloadStarted` now logs the catalog size for Whisper too; `modelDownloadProgress` per decile; `modelDownloadStalled` on retries) via a shared `updateDownloadProgress` helper used by both engine paths.
- Error-path split for Whisper: a download-phase failure keeps files on disk (every file is complete thanks to atomic per-file moves) so retry resumes by skipping them, matching the Parakeet policy. A prewarm-phase failure still runs `cleanupModelFiles` (ANE E5-bundle corruption). The Settings "Retry" affordance still offers a full reset via `cleanupFailedModel`.
- Path parity: new `whisperKitRepoDirectory` helper is the single source of truth for `Documents/huggingface/models/argmaxinc/whisperkit-coreml`, now shared by download, `deleteModel`, and `cleanupModelFiles`.

## Why the layout stays compatible

- The downloader writes the exact HubApi snapshot layout (`.../whisperkit-coreml/{identifier}/...`), verified against `WhisperKit.download` / swift-transformers `HubApi.snapshot`.
- Catalog identifiers are exact top-level folder names in `argmaxinc/whisperkit-coreml`, so exact-folder selection matches what WhisperKit's `"*{variant}/*"` glob resolves to (checked for Small, Small Quantized, Medium, Turbo 954MB against the live repo tree).
- `DictationCoordinator.preloadActiveModel` loads with `download: true`; HubApi hash-verifies existing LFS files (sha256 vs etag) and skips them, so the freshly downloaded weights are NOT re-downloaded — only tiny non-LFS configs may be re-fetched once on first load while HubApi writes its own metadata.

## Validation done

- `xcodebuild -scheme DictusApp -destination 'platform=iOS Simulator,name=iPhone 17 Pro' build` succeeds (FluidAudio Swift 5 patch applied to the checkout).
- Pure selection functions exercised standalone with 15 assertions, including the `openai_whisper-small` vs `openai_whisper-small_216MB` prefix-collision case and Parakeet-parity regressions (root `parakeet_vocab.json` included, `README.md` excluded).
- Live dry-run of the tree walk for `openai_whisper-small` with the identical algorithm: 19 files, 486,487,465 bytes total, all required verification paths covered. The two `weight.bin` files are 99% of the payload — the exact case where HubApi's file-count-weighted progress was distorted.

## How to test

1. **Whisper Small download (main check)**: Settings > Models > download "Small". Expect the bar to advance continuously with a "X MB of 486 MB" counter next to the percentage (note: repo payload is ~486 MB even though the catalog label says ~250 MB — same payload as before, previously invisible). Prewarm then completes and the model transcribes.
2. **Other variants**: repeat for Small Quantized (~216 MB) and Medium (~1.5 GB payload). Turbo compile validation stays maintainer/TestFlight territory (#144).
3. **Stall simulation**: start a Whisper download, then cut connectivity mid-download (Airplane Mode, or Network Link Conditioner "100% loss"). Expect: after ~30s a `modelDownloadStalled` log + automatic retries (2s/4s/8s backoff), then the localized "The download stalled..." error within ~100s. Re-enable network, tap retry: the download resumes, skipping completed files (progress jumps ahead quickly).
4. **Delete parity**: delete the downloaded Whisper model from Settings; re-download works from scratch.
5. **Parakeet regression**: download Parakeet v3 on a >=6 GB device; behavior must be unchanged (MB counter, progress, prewarm).

## Remaining validation (not claimed here)

- Turbo (`openai_whisper-large-v3_turbo_954MB`) prewarm/compile on device (#144 fragility).
- Onboarding on a <6 GB device (iPhone SE class), which gets Whisper Small through this new path on first install.


Pivii:

**Verdict: ✅ solid** — no high or medium severity defects found. The core claims were independently re-verified (details at the bottom); findings below are low severity.

## Findings

### Low 1 — Resume skip is existence-only, no size check
`DictusApp/Models/ModelRepoDownloader.swift:184` — a file is treated as complete if it merely exists:
```swift
if FileManager.default.fileExists(atPath: destination.path) {
```
The atomic-move design does prevent half-written files from landing at the destination (verified: HTTP status is validated in `didFinishDownloadingTo` before the move, and URLSession's temp file lives outside the destination tree). But the tree listing already gives you `file.size`, and the skip ignores it. Concrete scenario: argmaxinc pushes a new revision of a variant between a failed attempt and the retry; the retry keeps the old-revision file (even if its size no longer matches the listing) next to new-revision files. Mitigations exist (HubApi hash-verifies LFS files on first load when online; a prewarm failure wipes the folder via `cleanupModelFiles`), so this is defense-in-depth, not a bug: comparing on-disk size to `file.size` and re-downloading on mismatch would be a one-line hardening. Same pattern as the Parakeet baseline, but the whisperkit-coreml repo is much more actively re-pushed than the Parakeet repo, so the window is slightly more real here.

### Low 2 — Parakeet comment now implies a phase split its code doesn't have
`DictusApp/Models/ModelManager.swift:431` — the Parakeet catch comment was reworded to "Deliberately NO cleanupModelFiles here **after a download failure**", mirroring the new WhisperKit `downloadPhaseCompleted` split. But the Parakeet path still has a single catch covering both phases: a FluidAudio compile/prewarm failure also skips cleanup there, unlike the WhisperKit path which now cleans on prewarm failure. That asymmetry is pre-existing behavior (not a regression from this PR), but the comment now reads as if Parakeet distinguishes the phases. Either make the comment honest about the asymmetry or (follow-up) give Parakeet the same `downloadPhaseCompleted` treatment.

### Info — stale progress tick can outlive the download (pre-existing pattern)
`ModelManager.swift:227-231` — progress callbacks hop to the main actor via unstructured `Task { @MainActor in ... }`. A tick enqueued just before completion could in principle run after lines 246-248 remove `downloadProgress`/`downloadByteInfo`/`lastLoggedDeciles`, re-adding stale entries during `.prewarming`. Main-actor jobs run in enqueue order in practice, and the same pattern shipped in the Parakeet path in #209, so this is informational only.

## What was verified (no issues)

- **Parakeet parity**: `.parakeet()` reproduces the pre-refactor behavior byte-for-byte — same patterns (`requiredModels` + trailing slash, sorted), same root `.json`/`.txt` inclusion, same required-paths set (4 mlmodelc dirs + vocab). `shouldRecurse`/`fetchTree`/retry/delegate code is untouched; `shouldInclude` restructuring is equivalent when `includesRootMetadata == true`.
- **Prefix collisions**: checked against the live repo tree (31 root entries, no `Link` pagination header). The trailing slash in `"\(variant)/"` correctly excludes `openai_whisper-small_216MB`, `openai_whisper-small.en`, and `openai_whisper-small.en_217MB` when downloading `openai_whisper-small`, in both `shouldInclude` and `shouldRecurse`.
- **Required paths**: `MelSpectrogram.mlmodelc`, `AudioEncoder.mlmodelc`, `TextDecoder.mlmodelc`, `config.json` confirmed present in all six catalog variants (including deprecated tiny/base) on the live repo.
- **Layout parity**: `whisperKitRepoDirectory` matches HubApi's `downloadBase` default (`Documents/huggingface`) + `models/{repo}` (HubApi.swift:120-121, 378-379) and `WhisperKit.download`'s return value (`modelFolder.appending(path: variantPath)` where `variantPath` resolves to the exact folder for our identifiers). `deleteModel` and `cleanupModelFiles` now share the helper — no orphaning.
- **Error-path split**: `downloadPhaseCompleted` is set only after the download phase, so download failures keep complete files (retry resumes) and prewarm failures clean up. Cancellation during the prewarm-wait poll cleans up a fully downloaded model, which matches the pre-PR behavior (cleanup on any failure).
- **Stall detection**: no timer to leak — `timeoutIntervalForRequest` is the idle detector; the per-file session is torn down via `defer { session.finishTasksAndInvalidate() }`, and the continuation is resumed exactly once from `didCompleteWithError`.
- **HubApi first-load claim**: confirmed in `HubFileDownloader.download` (HubApi.swift:561-579) — missing metadata + matching sha256 etag keeps LFS weights and only rewrites metadata; non-LFS configs are re-fetched once. Offline cold-load behavior is unchanged by this PR: `getFilenames` performed a network GET before this PR too.
- **Conventions**: no force unwraps in new code, English comments, pbxproj rename consistent (all 4 refs), no leftover `ParakeetModelDownloader` symbol references.

Issue #174, opened 2026-05-16
Downloaded model is not automatically selected after download/preparation completes


Pivii:
## Summary

After downloading and preparing a new transcription model, Dictus returns from the download/optimization/loading screen but does not automatically select the newly downloaded model.

## Observed behavior

1. User taps a model card for a model that is not installed yet.
2. Dictus correctly shows the download / optimization / model loading flow.
3. When that flow disappears, the downloaded model is **not selected**.
4. The user has to tap the same model card again.
5. This triggers another loading step, making the flow feel broken or redundant.

## Expected behavior

Once a user explicitly chooses to download a model and the download/preparation succeeds, Dictus should automatically make that model the active selected model.

The user should not need to tap the same model card a second time after completion.

## UX impact

This is confusing because the user's original intent was already clear: they selected that model. The app confirms the download/preparation flow, but then silently keeps another model selected, forcing a second manual action.

## Acceptance criteria

- After successful download + preparation, the downloaded model becomes selected automatically.
- The model card UI reflects the new selected state immediately.
- Starting a dictation afterward uses the newly downloaded model without requiring a second tap.
- If download/preparation fails, the previous selected model should remain unchanged and the error should be explicit.

## Related issues

Possibly related historically, but not obviously the same:

- #19 — Improve model download and preparation UX
- #25 — model downloaded during onboarding not recognized on Models page
- #8 — pre-compile CoreML models during onboarding and model download

Pivii:
Fix implemented in PR #209 (commit `6136c1c`), bundled with the #207 App Review hotfix. A freshly downloaded model now becomes the active one right after download + preparation succeeds, and the existing proactive RAM load (#144) applies to it — no second tap, no redundant reload. On failure the previously active model is untouched. Validated on iPad Air 11" (M3) simulator: with Whisper Small active, downloading Parakeet v3 flipped the active model automatically.

PR #230, opened 2026-07-25
Make model deletion discoverable with an explicit overflow menu

### LINKED ISSUE #193: Improve UX discoverability for deleting downloaded models
## Summary

Users can already delete downloaded models by swiping on a downloaded model row, but this interaction is not discoverable enough. A second user has now asked for a way to delete models in order to switch models later.

This issue tracks a UX improvement to make model deletion explicit and easier to find.

## User feedback

A TestFlight user wrote:

> You should include a way to delete models if I wanna switch it in the future

Pierre confirmed that deletion is currently possible via swipe, but the user did not discover it.

## Current behavior

- Downloaded models appear in the **Downloaded** section.
- Users can delete a model by swiping the row.
- There is no visible delete affordance, hint, or contextual action explaining this.

## Problem

Swipe-to-delete is a native iOS pattern, but it is easy to miss in this screen because:

- model cards look like rich custom cards, not standard editable list rows;
- users may not expect destructive actions on a model card;
- there is no edit mode, overflow menu, or hint text;
- users who want to switch models may think they are stuck with the downloaded model.

## Possible UX improvements

Evaluate one or more of these options:

1. Add an explicit overflow/menu button on downloaded model cards (`…`) with actions:
   - Set as active model
   - Delete model
2. Add a small hint under the Downloaded section title:
   - "Swipe left to delete a downloaded model"
3. Add an Edit button for the Downloaded section, exposing delete controls.
4. Add a delete/trash icon in the model detail or expanded state, if we introduce one.
5. Show an onboarding/toast hint after the first model download:
   - "You can delete downloaded models later by swiping left."

## Recommendation

Prefer an explicit overflow menu or visible delete affordance over relying only on swipe. The model rows are visually custom enough that users do not naturally infer iOS list swipe actions.

A small hint can be added as a quick fix, but the longer-term UX should make deletion visible without requiring users to know the gesture.

## Acceptance criteria

- [ ] Users can clearly discover how to delete a downloaded model without prior knowledge of swipe gestures.
- [ ] The delete action remains protected against accidental deletion with a confirmation dialog.
- [ ] The UI still makes it clear which model is currently active/recommended/downloaded.
- [ ] UX copy works in both English and French.
- [ ] Existing swipe-to-delete behavior can remain, but should not be the only discoverable path.

Pivii:
> [!IMPORTANT]
> **Stacked PR — merge AFTER #229.** This branch is based on `feature/210-unified-model-downloader` (PR #229) because both touch the model manager. Until #229 merges, the diff below includes #210's commits — that is expected. Once #229 lands in `develop`, this diff will shrink to the #193 changes only.

Closes #193

## Summary

TestFlight users could not discover that downloaded models are deletable: the only entry point was swipe-to-delete, and the rich custom cards do not signal that iOS list swipe actions exist. This PR adds an explicit, always-visible overflow button (`ellipsis.circle`) at the top-right of every downloaded model card:

- Tapping it opens a native `Menu` with a destructive **Delete Model** action (trash icon).
- The action funnels into the exact same confirmation alert and `deleteModel` path as swipe-to-delete — no new deletion logic.
- When deletion is blocked, the menu explains why with a disabled row instead of hiding the action:
  - active model → "Select another model to delete this one"
  - last downloaded model → "At least one model must stay on your device"
- Swipe-to-delete keeps working unchanged.
- The menu only appears on downloaded (`.ready`) cards in the Models tab; Available-section cards and the onboarding flow are untouched.
- New strings localized in English and French via `DictusApp/Localizable.xcstrings`, with an accessibility label ("Model options") on the ellipsis button.

## How to test

1. Open the app → **Models** tab with at least two models downloaded.
2. Each downloaded card shows an ellipsis button top-right. Tap it on a non-active model → **Delete Model** → confirmation alert appears → confirm → model is removed and reappears in Available.
3. Tap the ellipsis on the **active** model → the menu shows a greyed-out explanation, no delete offered.
4. With only one model downloaded, its menu shows the "at least one model" explanation.
5. Swipe left on a deletable card → the swipe Delete action still works and shows the same confirmation.
6. Switch the device language to French → menu items and explanations are translated.
7. Regression: card tap still selects a ready model / starts a download on an available one; download progress and the prep overlay behave as on `develop` + #229.


Pivii:

**Verdict: ⚠️ solid overall, 2 findings (1 medium UX-correctness, 1 low). Scope reviewed: only commit `51c1333` (diff vs `feature/210-unified-model-downloader`).**

The core design is sound and verified: the `Menu` overlay sits outside the `Button` label (correct — nested controls inside a Button label don't receive taps), its hit area is limited to its own ~44pt bounds (`.padding(12)` + `.contentShape`) so it cannot steal taps from the rest of the card; the menu is gated to `.ready` + non-nil handler, the Available section and previews never pass `onDeleteRequest` (and `ModelCardView` has no other call sites), downloading/prewarming cards can't show it, and the delete funnels into the exact same `modelToDelete`/`showDeleteAlert`/`deleteModel` path as swipe-to-delete with no duplicated state. `canDeleteModel` matches `ModelManagerView.canDelete` semantics. xcstrings: 4 new keys valid JSON, `sourceLanguage: en`, FR `translated`. No force unwraps, comments in English.

### Medium — blocked-reason priority is inverted for the single-model case; contradicts the PR test plan

`DictusApp/Views/ModelCardView.swift:109-123` — the branch order checks `isActive` before the last-model case. But a lone downloaded model is virtually always the active one: `ModelManager` auto-activates on download (`ModelManager.swift:307` and `:414`, issue #174) and reassigns `activeModel = downloadedModels.first` on delete (`:531`). So with exactly one model downloaded, the user sees **"Select another model to delete this one"** — an instruction that is impossible to follow (there is no other model to select; they'd have to download one). The "At least one model must stay on your device" row is effectively dead code, reachable only if `activeModel` is nil or stale.

Concrete failure: PR test-plan step 4 says *"With only one model downloaded, its menu shows the 'at least one model' explanation"* — the code shows the active-model message instead. Fix: check `downloadedModels.count <= 1` first (it's the stronger constraint — also the only one `ModelManager.deleteModel` actually enforces), then `isActive`.

### Low — fixed paddings don't scale with Dynamic Type; badges can slide under the button at accessibility sizes

`DictusApp/Views/ModelCardView.swift:125-130` and `:165` — the ellipsis glyph uses `.font(.title3)`, which scales with Dynamic Type (roughly 20pt → 40pt+ at accessibility sizes), but both the 12pt tap-target padding and the 32pt trailing reservation on the badge row are fixed. At default sizes the math works (row 1 clears 16+32 = 48pt from the trailing edge vs ~44pt of overlay). At AX sizes the overlay grows to ~64pt+ while the reservation stays 48pt, so the "Recommended"/engine badges can slide under the button and the enlarged menu hit area silently eats card taps in that corner. Consider `@ScaledMetric` for both values (or `relativeTo: .title3`).

### Observation (pre-existing, not introduced by this PR)

The confirmation alert (`ModelManagerView.swift:227-236`) does not re-validate `canDelete` at confirm time, and `ModelManager.deleteModel` only guards the last-model case — deleting the active model is silently allowed (it reassigns `activeModel`). The UI gating makes this unreachable in practice and swipe-to-delete has the same shape, so no action needed here; noting it because this PR adds a second entry point into that funnel.

Pivii:

Both review findings are addressed in commit 8bb26eb:

- **Medium (blocked-reason priority):** the overflow menu now checks the last-model case (`downloadedModels.count <= 1`) before `isActive`, so a single downloaded (auto-activated) model correctly shows "At least one model must stay on your device" instead of the impossible-to-follow "select another model" instruction. The multi-model active case still shows the "select another model" row.
- **Low (Dynamic Type):** the 12pt tap-target padding and the 32pt badge-row trailing reservation are now `@ScaledMetric(relativeTo: .title3)`, matching the ellipsis glyph's font so the reserved space grows with accessibility text sizes.

Build verified: `xcodebuild -scheme DictusApp` succeeds on iPhone 17 simulator.

Issue #235, opened 2026-07-27
Retry button on a failed model download moves the card instead of relaunching the download


Pivii:
> *This was generated by AI during triage.*

## Bug

When a model download fails (e.g. airplane mode mid-download), the error message shows correctly, but the retry flow is confusing:

1. The partially-downloaded model card appears in the **Downloaded** ("Téléchargés") section with an orange "Réessayer" retry arrow.
2. Tapping **Réessayer does NOT relaunch the download** — it moves the card back to the **Available** ("Disponibles") section.
3. The user must then tap the model a second time to actually start the download.

Observed by the maintainer on device on 2026-07-27 while validating PR #229 (screenshot: Parakeet v3 sitting in "Téléchargés" with the Réessayer button, airplane mode on). **Pre-existing behavior** — identical before PR #229 with the Parakeet downloader; NOT a regression of the unified downloader.

## Maintainer input (2026-07-27)

- **Locked:** tapping Réessayer must relaunch the download directly, in place — no two-tap dance.
- **Open (to decide at triage):** which section the partially-downloaded card should live in. Leaning acceptable to keep it in Downloaded since bytes ARE on disk, as long as retry works in place; alternatively move it to Available immediately on failure. Either way the state must be visually honest.

## Context for the fix

- Fix AFTER PRs #229 and #230 merge — both touch the same files (ModelCardView/ModelManagerView/ModelManager).
- PR #229 introduced resume (completed files are kept after a failure and skipped on retry), which makes in-place retry extra valuable: it resumes instead of restarting from zero.

## Acceptance criteria

- [ ] After a failed download, tapping the retry affordance immediately relaunches the download (resuming kept files).
- [ ] The card's section placement during/after a failed download follows the decision above and never shows a non-downloaded model as silently "Downloaded".
- [ ] Behavior identical for Parakeet and WhisperKit models.

Pivii:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Tapping the retry affordance on a failed model download must relaunch the download in place (resuming kept files) instead of wiping the partial download and moving the card.

**Triage finding (code trace, 2026-07-28):**
[...]

Issue #123, opened 2026-04-17
Audio route recovery breaks with AirPods / Bluetooth mic, causing wrong input and first-retry failure


Pivii:
## Summary

Dictus can get into a broken audio-route / input state when Bluetooth earbuds (AirPods) are involved, especially after listening to audio, taking a call, or recovering from a reset of the Dictus audio system.

In practice, Dictus may keep using the iPhone microphone instead of the AirPods microphone, and the first recording attempt after reset can fail with a microphone-unavailable style error before a second attempt works again.

## User-observed behavior

Scenario reported:
- Dictus had already been used successfully earlier
- user then put on earbuds / AirPods to listen to audio
- user wanted to reply with voice input
- Dictus did **not** use the earbuds microphone and appeared to keep using the phone microphone instead
- user had to stop Dictus audio (Dynamic Island / force quit) to recover normal behavior
- after that, the next cold start failed once with a **microphone unavailable** style error
- a second cold start / retry then worked, though with additional glitches

## Why this matters

This breaks a very important real-world use case:
- listening through earbuds
- then immediately replying using voice

It also suggests the app is not correctly tracking whether the active audio input route is still valid / preferred.

## Evidence from logs

The attached session logs point to a few relevant signals:

### 1. Broken / stale engine state after recovery attempt
- `dictationFailed error=Zombie engine: 0 samples after 2s, force-restarting`
- followed by:
- `engineWarmUpFailed context=forceRestart error=(OSStatus 560557684)`

This suggests the engine can enter a bad state where recording is "started" but no usable input is actually flowing.

### 2. First retry fails, second retry works
Logs show a first recovery attempt that ends in failure, then a later retry that succeeds.
That matches the user experience of:
- first attempt -> microphone unavailable / failure
- second attempt -> recording works again

### 3. Route / readiness mismatch suspicion
The user report strongly suggests the selected microphone route is wrong after Bluetooth / call transitions:
- earbuds are connected and in use
- Dictus does not actually use the earbuds mic
- stopping / resetting Dictus changes the outcome

So this may be a route-selection / route-recovery issue rather than a pure permission issue.

## Hypothesis

Possible causes include one or more of:
- Dictus not re-evaluating preferred input correctly after Bluetooth audio route changes
- stale `AVAudioSession` input route after listening / call interruption
- audio engine surviving in a logically warm state while actual input is no longer valid
- failed recovery path after interruption, leading to zombie-engine behavior
[...]

PR #251, opened 2026-07-28
fix(models): load Whisper from the local folder on the dictation path (#249)

### LINKED ISSUE #249: Cold-start dictation re-downloads the Whisper model instead of loading it locally — every Whisper model is unusable offline

## Bug

Cold-start dictation initialises WhisperKit **by model name with downloads enabled and no local model folder**, so it resolves the model against the Hugging Face hub over the network — even though the model is already downloaded on the device. When that request fails, dictation fails.

Captured on device 2026-07-28, iOS 26.5.2, app 1.8.0 (23), `openai_whisper-medium`.

## Evidence

Audio capture was **completely healthy** — this is not a microphone or audio-route problem:

```
19:46:20  audioSessionConfigured category=playAndRecord
19:46:20  audioRouteChanged reason=routeConfigurationChange details=inputs=MicrophoneBuiltIn
19:46:20  audioEngineStarted
19:46:21  waveformSnapshot count=30 min=0.023 max=0.922 spread=0.900     ← real signal
19:46:25  engineCollectResult sampleCount=83252 engineRunning=true        ← ~5.2s captured
```

The failure is entirely on the model path:

```
19:46:26  engineWarmUpFailed context=init-preload
          error=Model not found. Please check the model or repo name and try again.
          Error: httpStatusCode(504)
19:46:26  dictationFailed error=Cold start engine load: Model not found. …
          Error: httpStatusCode(504)
19:46:26  statusChanged from=transcribing to=failed
```

`httpStatusCode(504)` is a gateway timeout from the remote hub. The second attempt ~6 seconds later succeeded (`transcriptionCompleted duration=2016ms words=16`) because the engine was already resolved and running.

## Root cause

Two WhisperKit initialisation paths exist and they disagree.

- The **model management** path builds its config with an explicit local `modelFolder` and `download: false` — fully offline, correct.
- The **cold-start dictation** path builds its config with only the model *name*, no `modelFolder`, and `download: true` — so WhisperKit resolves the identifier against the hub and can reach the network.

The local path is already known: model management resolves it as `Documents/huggingface/models/argmaxinc/whisperkit-coreml/{identifier}` and treats it as the single source of truth for download, delete and cleanup (#210). The dictation path simply does not use it.

## Why this matters

1. **It breaks a core promise.** Dictus is positioned as fully offline. A cold-start dictation currently depends on a third-party service being reachable and healthy. A flaky connection, airplane mode, or an outage on the hub's side breaks dictation on a device where the model is already present.
2. **It is very likely the real cause of the long-standing "first attempt fails, second attempt works" pattern**, which has been attributed to audio-route recovery (#123). At least this occurrence has nothing to do with audio.
3. **The user-facing error is wrong and untranslated.** The user is shown *"Model not found. Please check the model or repo name and try again"* — an English, developer-facing string — for what is actually a transient network failure. Users read this as a microphone or installation problem.

Worth noting for scope: what leaves the device here is model-identifier metadata to the hub, not user audio or transcriptions. It is a reliability and product-promise problem, not a user-data leak.

---

## Agent Brief

**Category:** bug
**Summary:** Cold-start dictation resolves the Whisper model over the network instead of loading the already-downloaded local copy, so a hub outage breaks dictation.

**Current behavior:**
The dictation engine initialisation constructs its WhisperKit configuration from the model identifier alone, with downloads enabled and no local model folder. WhisperKit therefore performs hub resolution over the network. On a `504` the initialisation throws, dictation reports failure, and the user sees an untranslated developer-facing message. A retry usually succeeds because the resolved engine is reused.

**Desired behavior:**
When the selected model is already present on disk, cold-start dictation loads it **entirely locally** — no network request — using the same local repository path that model management already treats as the source of truth. Dictation must work in airplane mode with a downloaded model.

When the model is **not** present locally, dictation must not silently attempt a download on the hot path. It should fail fast with a clear, localised, actionable message directing the user to the model manager. Downloading remains the model manager's responsibility.

**Key interfaces:**
- The dictation engine initialisation — must supply the local model folder and disable downloading, matching the configuration model management already uses.
- The local repository path resolution currently owned by model management — should become reachable from the dictation path rather than duplicated. Keep one source of truth (#210).
- Error surfacing to the keyboard via the shared last-error key — a model-availability failure and a network failure must reach the user as distinct, localised messages, and neither should be phrased as a microphone problem.

**Acceptance criteria:**
- [ ] With the model downloaded and the device in airplane mode, a cold-start dictation completes successfully.
- [ ] No network request is issued during cold-start dictation when the model is present locally. Verify by observation, not by assumption.
- [ ] With the model absent locally, dictation fails immediately with a localised message pointing at the model manager, and does not attempt a download.
- [ ] No English developer-facing string can reach the user on this path.
- [ ] Both Whisper and Parakeet selections are checked — confirm whether the Parakeet path has the same split and report it rather than silently changing it.
- [ ] The local path stays consistent with what model management uses for download and delete; deleting a model still makes dictation report it as missing.

**Out of scope:**
- Audio session, route handling and the `#123` investigation — the capture above shows audio was healthy.
- Changing the model download UI or the downloader itself (#210, #235).
- Adding an offline/online indicator to the UI.

**One decision to confirm before implementing:** first-run behaviour, where no model has ever been downloaded. This brief assumes dictation should fail fast and defer to the model manager rather than download on the hot path. If the existing onboarding already guarantees a model before dictation is reachable, state that in the PR instead of adding a redundant guard.

## Related

- #123 — audio-route issue; the "first attempt fails" pattern attributed there may partly belong here
- #210 — unified model downloads, established the canonical local repo path
- #144 — Whisper turbo unreliability, separate cause

Pivii:
Fixes #249.

## The bug

Cold-start dictation built its `WhisperKitConfig` from the model **name** alone, with `download: true` and no `modelFolder`. `WhisperKit.setupModels` only skips the network when `modelFolder` is set; otherwise it resolves the identifier against the Hugging Face hub and downloads. So the dictation path tried to download a model that was already on disk, and every Whisper model was unusable offline. The failure surfaced to the user as WhisperKit's own untranslated string, `Model not found. Please check the model or repo name and try again.`

The model-manager prewarm path was already correct (`modelFolder` + `download: false`). The two paths simply disagreed.

## What changed

**`DictusCore/Sources/DictusCore/WhisperModelRepository.swift` (new)** — the canonical local path, `Documents/huggingface/models/argmaxinc/whisperkit-coreml/{identifier}`, extracted from `ModelManager` (where #210 established it) so download, delete, cleanup and dictation all resolve it in one place instead of duplicating it. It exposes:

- `modelFolderURL(for:)` — the path whether or not it exists, for delete/cleanup.
- `installedModelFolderURL(for:)` — the path only when the folder actually holds a compiled `.mlmodelc` bundle, so a partial or wiped download is never handed to WhisperKit.
- `cachedTokenizerRepositoryNames()` — diagnostic only, see "Honest limits" below.

`ModelManager.whisperKitRepoDirectory` now delegates to it, so download, delete and cleanup keep using exactly the same path as the loader. Deleting a model still makes dictation report it as missing (covered by a test).

**`DictusApp/DictationCoordinator.swift`** — `ensureWhisperKitEngineReady` resolves the installed folder first and loads with `modelFolder` + `download: false`. If the model is not installed it throws immediately, before creating the init task, and never downloads. Two `diagnosticProbe` log lines were added (`localModelResolved` / `localModelMissing`) so the local path is observable in an exported log.

**`SpeechModelError`** (in `DictusApp/Audio/SpeechModelProtocol.swift`) — a `LocalizedError` with two cases, `modelNotInstalled` and `engineLoadFailed`. `errorDescription` is localised (FR + EN, added to `DictusApp/Localizable.xcstrings`); `diagnosticDescription` keeps the English technical detail for `PersistentLog`. Every engine-load failure on this path, including the one an awaiting concurrent caller receives, is wrapped, so no WhisperKit or Core ML English string can reach `handleError` and the keyboard's banner. The sibling `"No model downloaded. Open Dictus to download a model."` was also wrapped in `String(localized:)` — it was reaching users in English regardless of device language.

**`WhisperKitEngine.prepare(modelIdentifier:)`** carried a second copy of the same bug. It is currently dead code (the coordinator injects its instance via `setWhisperKit`), but it is fixed the same way so it cannot regress.

**Refactor, worth flagging for review**: `ensureEngineReady`, `ensureWhisperKitEngineReady` and `ensureParakeetReady` moved verbatim into a `private extension DictationCoordinator` at the bottom of the file. Only the WhisperKit one changed; the other two are byte-identical apart from dropping the now-redundant `private` modifier. The reason is not stylistic: `DictationCoordinator`'s class body was already a baselined **error**-severity `type_body_length` violation at 620 lines, and the fix pushed it to 644. The move brings it to 525 — a warning, and one baseline entry fewer than before.

## First-run behaviour (the decision the brief asked about)

**No redundant guard was added, and onboarding does guarantee a model.** `ModelDownloadPage` is step 5 of 6 in `OnboardingView`, which is a `fullScreenCover` with `interactiveDismissDisabled()` and no swipe navigation. Its "Continue" button only renders once `downloadComplete == true`; there is no skip path. So dictation is unreachable before a model exists.

That said, the new guard is **not** redundant, for a different reason: `SharedKeys.modelReady` is App Group bookkeeping, not disk truth. It stays `true` after iOS purges files, after a partial download, or after any divergence between the persisted list and what is on disk. The guard checks the files WhisperKit will actually read, which is the check that was missing.

## Parakeet finding (checked, not changed)

**Parakeet does not have the same split, and I changed nothing there.** Both `ModelManager.downloadParakeetModel` and `DictationCoordinator.ensureParakeetReady` go through the single call `ParakeetEngine.prepare` → `AsrModels.downloadAndLoad(version: .v3)`, which resolves the FluidAudio cache locally and skips the download when the files are present. One path, no divergence — which is exactly why it already worked offline in the maintainer's airplane-mode log.

One asymmetry worth recording but deliberately left alone: if Parakeet models are *absent*, `downloadAndLoad` would attempt a download on the dictation hot path, which is the behaviour this PR removes for Whisper. It is out of scope here, it is not the reported bug, and touching it risks the one engine that currently works offline. Worth a follow-up issue if you want the two engines symmetric.

## How I verified

- `swift test` in `DictusCore`: **435 tests, 0 failures**, including 10 new `WhisperModelRepositoryTests` covering path construction, the quantized-identifier case, complete vs partial download detection, and delete-makes-it-missing.
- `xcodebuild build -scheme DictusApp -destination 'generic/platform=iOS Simulator'`: **BUILD SUCCEEDED** (the only warnings in the touched file are pre-existing `#available(iOS 17)` redundancy notes).
- `swiftlint lint --strict --no-cache`: **exit 0** with the regenerated baseline. No rule disabled; the baseline shrank from 46 to 45 entries.
- **No-network claim, source-level**: I read WhisperKit's `setupModels` in the resolved checkout. When `modelFolder` is non-nil it assigns the URL and returns — the `else if download` branch that calls `HubApi` is not reached. `loadModels` then reads Core ML bundles from that path. That is a static reading of the dependency, not a runtime observation.

**I did not run this on a device and did not observe network traffic.** I could not verify the airplane-mode criterion empirically. The `localModelResolved` log line was added specifically so the on-device test can confirm the local path was taken.

## Honest limits — please check this during device validation

WhisperKit loads the **tokenizer separately from the model**. It is not in the variant folder (I confirmed against the HF API: `openai_whisper-small/` ships only the three `.mlmodelc` bundles, `config.json` and `generation_config.json`). `ModelUtilities.loadTokenizer` looks for `tokenizer.json` under `Documents/huggingface/models/openai/whisper-*` and **falls back to downloading from the hub** if it is not there.

In practice it should be cached: the online prewarm that immediately follows every download loads with `load: true`, which populates that cache. But this is the one remaining way the path could touch the network, and I could not test it. Mapping a variant identifier to its tokenizer repo lives inside WhisperKit and I deliberately did not duplicate it, so there is no guard for this — instead the `localModelResolved` log line reports `tokenizerCache=` with whatever is cached under `models/openai/`. If airplane mode still fails after this PR, that field will say so immediately, and the follow-up is to pre-seed or bundle the tokenizer.

## Out of scope

- Audio session and route handling (#123). Not touched.
- The downloader and model-manager UI (#210, #235). Not touched.
- The Parakeet hot-path download asymmetry described above.
- Any offline/online indicator in the UI.

## Suggested device test (airplane mode, per the issue)

1. Online: download and select `openai_whisper-small_216MB`, confirm it transcribes.
2. Airplane mode on, force-quit the app, dictate from the keyboard. Expect success, and `localModelResolved` in the exported log.
3. Repeat with a full variant (`openai_whisper-medium`) — both were confirmed affected.
4. Switch to Parakeet, still offline, dictate. Expect no change from today.
5. Delete a Whisper model, select it if the UI allows, dictate. Expect the French message "Ce modèle n'est pas installé…" and no download attempt.

## Summary by CodeRabbit

* **New Features**
  * Added localized, user-facing error messaging for speech-model preparation and engine load failures.
  * Added robust local model repository and tokenizer cache discovery.
* **Bug Fixes**
  * Dictation now fails fast when a required speech model isn’t installed locally, instead of attempting to download it.
  * Improved handling of incomplete or unavailable model installations.
* **Tests**
  * Added automated coverage for local model resolution, installation status (including partial downloads), model removal, and tokenizer cache detection.


Pivii:

## Device validation protocol

The PR's own verification is source-level only — no device run, no observed network traffic. Everything below closes that gap. **The residual risk is the tokenizer, not the model.**

### Why the tokenizer is the thing to test

`WhisperKitConfig.download` gates model resolution, **not** tokenizer loading. `ModelUtilities.loadTokenizer` searches local folders for `tokenizer.json` and falls through to a hub download when it is missing *or when local loading throws*. So `download: false` does not by itself make WhisperKit fully offline.

Two facts that shape the tests:

1. **Tokenizers are per-variant, not shared.** The variant→repo mapping is one-to-one: `.small` → `openai/whisper-small`, `.medium` → `openai/whisper-medium`, `.largev3` → `openai/whisper-large-v3`, with separate `.en` entries. Each installed size needs its own tokenizer cached. Quantized and non-quantized of the same size *do* share one, since the variant is detected from model dimensions rather than the identifier.
2. **App code never deletes tokenizers.** Tokenizers live in `Documents/huggingface/models/openai/whisper-*`, a sibling of `argmaxinc/whisperkit-coreml`. Both `deleteModel` and `cleanupModelFiles` are scoped to `argmaxinc/…/{identifier}`. Being under `Documents` rather than `Caches`, they are also not purged by iOS.

The expected mechanism is that the online prewarm following each download populates the tokenizer for that variant. **This is unverified.** The new `localModelResolved` log line reports `tokenizerCache=`, so every test below is self-diagnosing: if a run fails offline, that field says whether the tokenizer was present.

Read it with: `grep "localModelResolved\|localModelMissing" <log>`

---

### A. Core fix — does it work at all

- [ ] **A1.** Airplane mode, `whisper-medium` installed, force-quit, dictate. Expected: works. This is the exact reproduction from #249.
- [ ] **A2.** Same with quantized `whisper-small`. Both variants were confirmed broken before the fix.
- [ ] **A3.** Confirm `localModelResolved … download=false` appears in the log, and that no `downloadError` appears anywhere.
- [ ] **A4.** Back online, dictate normally. Confirms nothing regressed on the happy path.

### B. Model switching — the tokenizer concern

Each step **offline**, force-quitting between switches so the cold-start path is exercised rather than a warm in-memory engine.

- [ ] **B1.** Install small **and** medium while online. Then airplane mode: switch small → medium → small → medium, dictating after each switch. Expected: all succeed, since each download prewarmed its own tokenizer.
- [ ] **B2.** Install a size never used before (e.g. `tiny` or `base`) **while online**, then go offline and dictate with it. Verifies the download+prewarm really does cache that variant's tokenizer.
- [ ] **B3.** The adversarial case: install a new size **while offline** (should be refused by the downloader), then confirm dictation still works with the previously installed models. The failure must stay contained to the missing model.
- [ ] **B4.** Record `tokenizerCache=` at each step and note whether the list grows as expected. If a switch fails, this field is the answer.

### C. Delete and reinstall

- [ ] **C1.** Delete a model, reinstall it **offline**. Expected: refused with a clear localised message pointing at the model manager — not a raw WhisperKit string, not a silent hot-path download.
- [ ] **C2.** Delete a model, reinstall **online**, then dictate offline. Confirms reinstall restores a fully working offline state.
- [ ] **C3.** Delete a model and dictate offline with a *different* still-installed model. Expected: unaffected — deletion is scoped and must not disturb the others.

### D. Error paths and localisation

- [ ] **D1.** No English developer-facing string reaches the keyboard on any path above. Specifically, WhisperKit's `"Model not found. Please check the model or repo name and try again."` must never surface again.
- [ ] **D2.** The missing-model message is in French on a French device and points at the model manager.
- [ ] **D3.** Cold start where the keyboard fires while the launch preload is already in flight — the concurrent-await branch was also wrapped in this PR and needs at least one real exercise.

### E. Regression — do not break what worked

- [ ] **E1.** Parakeet offline, cold start. It already worked; the PR must not have disturbed it.
- [ ] **E2.** Parakeet ↔ Whisper switching, offline, dictating after each switch.
- [ ] **E3.** Normal online dictation on both engines.

---

### Review focus, separate from testing

The largest part of the diff in `DictationCoordinator` (~309 lines) is **not** the fix — it is three methods moved into a private extension to get back under the baselined `type_body_length` limit. The author states the move is verbatim and CI is green, but it is the change least related to the bug, in the file that drives dictation. That is where a reviewer's attention is worth spending.

### Known follow-up, deliberately out of scope

If Parakeet models are **absent**, `AsrModels.downloadAndLoad` would download on the dictation hot path — the behaviour this PR removes for Whisper. Not a regression and not urgent, but worth its own issue if symmetry is wanted.

Issue #202, opened 2026-07-10
DictusKeyboard exceeds its jetsam memory limit (~90 MB) and gets killed


Pivii:
## Summary
On a physical iPad Pro 11\" 1st gen (A12X, 4 GB, iPadOS 26, iPhone-compat mode), the `DictusKeyboard` extension is killed by jetsam **~4 s after launch** for exceeding its per-process memory limit. iOS then falls back to another keyboard (observed as "Dictus shows a few keys then switches away"). It relaunches and is killed again identically.

This was surfaced while diagnosing #197 via live device logs (`idevicesyslog` / Console.app). The #197 next-keyboard globe fix is **not** the cause — the kill is a pure `per-process-limit` memory kill, independent of the layout.

## Evidence (device kernel log)
```
runningboardd  [DictusKeyboard:6637] Memory Limits: active 90 inactive 90
...~4 s later...
kernel  memorystatus: DictusKeyboard [6637] exceeded mem limit: ActiveHard 90 MB (fatal)
kernel  killing_specific_process pid 6637 [DictusKeyboard] (per-process-limit 100 4s) 92161KB
SpringBoard  Process exited: domain:jetsam(1) code:per-process-limit(7)
ReportSystemMemory  Process DictusKeyboard [6637] killed by jetsam reason per-process-limit
```
Footprint at kill: **~92 MB** against a **90 MB** hard limit. Reproduced twice back-to-back (pids 6637, 6639).

## Why it matters
- `CLAUDE.md` targets **~50 MB** for the keyboard extension; we are at ~90 MB, i.e. right at (over) the ceiling.
- The limit is device-dependent. This iPad grants 90 MB; some contexts grant less (~50–60 MB). A keyboard that needs ~90 MB is fragile and can be killed on iPhone too under memory pressure.
- Even though the app is iPhone-only (not the App Review target device), this is a real robustness bug.

## Suspected contributors (to investigate/profile)
- SwiftUI `UIHostingController` (toolbar + recording overlay + emoji picker) overhead.
- Vendored keyboard cache (`Cache loaded with 6210 pre-cached in CacheData`).
- Spell-correction / prediction dictionaries loaded in-extension (SymSpell ~8 MB, see Phase 25 trie migration).
- Emoji dataset.

## Next steps
1. Profile the extension with Instruments (Allocations / VM Tracker) on device to attribute the ~90 MB.
[...]

Issue #138, opened 2026-04-30
Intermittent missed spacebar input while typing on iOS keyboard


Pivii:
## Summary
A beta user reports intermittent missed spacebar input on the iOS keyboard. While typing normally, spaces sometimes do not get inserted, causing multiple words to be concatenated.

## User report
> Sur iOS j'ai l'impression que cavientsurtoutd'unelatenceala frappe sur le clavier (voilà ce que ça donne haha)

Expected text was likely close to:
> Sur iOS j'ai l'impression que ça vient surtout d'une latence à la frappe sur le clavier

Observed problematic segment:
> cavientsurtoutd'unelatenceala frappe

## Observed behavior
- Space presses appear to be intermittently ignored while typing.
- The issue produces merged words such as `cavientsurtoutd'unelatenceala`.
- At least one later space was inserted correctly before `frappe`, so the spacebar is not completely broken.

## Expected behavior
Pressing the spacebar should reliably insert a single space, even during fast typing or when the keyboard is under load.

## Current interpretation
This looks more consistent with input latency / dropped taps than a simple isolated mis-tap:
- several consecutive word boundaries are missing, not just one;
- the user explicitly describes a feeling of latency while typing;
- some spaces still work, suggesting an intermittent responsiveness problem rather than a permanently non-functional key.

Possible areas to investigate:
- touch handling latency on the spacebar / bottom row;
- whether rapid typing queues or drops key tap events;
- whether autocorrect / text update work blocks the main thread around space insertion;
- whether the spacebar has a smaller effective hit target or edge dead zone than expected;
- whether the keyboard extension is doing expensive work synchronously during key input.

## Reproduction status
Not yet confirmed reproducible. Based on one user report with an example typed sentence.

## Suggested test cases
1. Type a French sentence quickly with many short words, e.g. `ça vient surtout d'une latence à la frappe`.
2. Repeat while the keyboard has just been opened / after switching apps.
3. Compare normal typing speed vs fast typing.
4. Add temporary logging around spacebar touch down/up and text insertion to distinguish:
   - tap not received;
   - tap received but insertion delayed;
   - tap received but insertion skipped / overwritten.

## Impact
Medium: core keyboard typing reliability issue, but currently intermittent and not yet reproduced consistently.

Pivii:
Additional user report confirming the same symptom:

[...]

Issue #117, opened 2026-04-14
Audit and align keyboard key geometry with Apple system keyboard


Pivii:
## Problem

Multiple beta testers report making more typing errors on the Dictus keyboard than on Apple's system keyboard, with the subjective feeling that the layout is \"not quite the same\" as Apple's. A typical quote:

> Vu que le layout et la disposition des touches n'est pas tout à fait le même que celui d'Apple, j'ai du mal à taper du texte dessus et je fais beaucoup d'erreurs.

This is now a repeated pattern across testers, so it's likely a real geometric discrepancy rather than pure habituation.

## Hypothesis

Our key heights, row spacing, horizontal padding, and/or vertical position within the keyboard extension frame do not match Apple's system keyboard exactly. Even small offsets (~2pt) can shift users' muscle memory and increase typo rate.

Current values (`DictusKeyboard/KeyboardMetrics.swift`):
- `keyHeight`: compact=40, standard=43, large=46
- `rowSpacing`: compact=9, standard=11, large=12
- `keySpacing`: compact=5, standard=6, large=6
- `rowSidePadding`: compact=3, standard=4, large=5

Apple's published HIG keyboard height is documented but key-level geometry is not. We need empirical measurements.

## Plan

1. **Measure Apple's keyboard** on reference devices (iPhone SE, iPhone 15/16, iPhone Plus):
   - Take screenshots of Messages + system keyboard at identical state (lowercase, no suggestions)
   - Measure in pixels: key width, key height, row spacing, horizontal row padding, top/bottom padding of keyboard frame
   - Convert to points using device scale
2. **Measure Dictus keyboard** at identical state, same devices
3. **Compute deltas** per device class — identify which constants are off
4. **Overlay comparison** — put Apple + Dictus screenshots side by side, align origin, visualize gaps
5. **Update `KeyboardMetrics.swift`** with corrected values
6. **Validate on device** — have testers type the same text on both keyboards, count errors before/after

## Acceptance

- Each key's center position matches Apple's within ~1pt on each reference device
- Tester feedback improves on next beta (subjective)

## Notes

- Related to #56 (dead zones, closed) and the incremental keyboard rebuild approach
- Do not change key visual style as part of this — only geometry

## Priority

High — this is now reported by multiple testers and directly impacts typing accuracy / perceived quality of the keyboard.

Pivii:
## Round 1 fix shipped (vertical alignment) — round 2 abandoned

### What was done

[...]

PR #253, opened 2026-07-28
fix(keyboard): extend bottom-row hit area below the visible keys (refs #138)

Pivii:
Fixes the bottom-row hit-target bug re-diagnosed in #138.

## What was wrong

The key grid (`GiellaKeyboardView`) is pinned flush to the input view's bottom edge and its `UICollectionView` fills it exactly, so bottom-row cell frames terminate at that edge. A touch landing below it fails UIKit's default `point(inside:with:)`, is never delivered to `touchesBegan`, and resolves to no key at all. The existing `clampedPoint` helper cannot compensate — it only runs on touches the view already accepted. Apple's keyboard keeps responding lower than ours, so a user with Apple muscle memory taps a spot Dictus silently ignores and the space is dropped.

## What changed

Only **where touches are accepted** — nothing rendered moves or resizes.

- **`GiellaKeyboardView.point(inside:with:)`** accepts points up to a tolerance below the grid. Horizontal behaviour and the top edge are untouched. `clampedPoint` then maps the accepted point onto the bottom row (its `y` is clamped to `height - 4`), so there is **no parallel resolution path**: `handleTouches`, `touchesMoved`, key repeat and `touchesFoundLongpress` all keep using the same helper, which is why long-press and swipe-to-alternative keep working on the bottom row.
- **`KeyboardInputView.point(inside:with:)`** accepts the same band. Hit testing is hierarchical: the grid is flush with the input view's bottom, so a point below the grid is also below the input view — without this override UIKit never descends into the grid and the grid's own override would never be consulted.
- **The globe key's invisible tap button** is extended downward on the bottom row (`hitFrame(forCellFrame:inSection:)`). The globe is handled by a transparent `UIButton` over its cell (it needs the raw touch stream for `advanceToNextInputMode()`), so it would otherwise be the one bottom-row key without tolerance. The button has a clear background and no content — a taller frame renders nothing. The `keyboardButtonFrame` setter still only assigns a frame, so the #202 layout-loop guard is unaffected.

### The #117 trap

No geometry was changed. `computeKeyboardHeight()`, the `bottomPadding = 0` constant and `sizeForItemAt` are untouched, so nothing divides extra height across the four rows. Grid height, per-key cell height, key visuals and the toolbar are byte-for-byte the same code path as before.

## The tolerance value — read this before merging

`KeyboardHitArea.bottomRowDownwardTolerance = 12.0` (new file `DictusKeyboard/TouchHandling/KeyboardHitArea.swift`).

**This is a provisional starting point, not a measurement.** I had no physical device, and I did not compare anything against Apple's keyboard. The number comes from constraints only:

- it must stay under the home-indicator inset iOS reserves below the input view (~34pt portrait, ~21pt landscape) so the band can never reach the host app, and so it still fits in landscape;
- each bottom-row cell already carries `Theme.keyVerticalMargin` (5.5pt portrait, 3.5pt landscape) of hit area under the visible key glyph, so 12pt adds to that rather than replacing it — roughly 17.5pt of total tolerance below the visible spacebar in portrait;
- it stays well under half a cell height, so a deliberate home-indicator swipe is not systematically captured by the keyboard.

**To tune:** change that one constant and nothing else. `0` restores the exact pre-fix behaviour.

## Instrumentation for taking the measurement

Two `diagnosticProbe` entries (Settings → Logs), following the existing `PersistentLog` pattern:

1. `component=GiellaKeyboardView action=bottomRowHitArea` — emitted once per distinct geometry (open, rotate, page change):
   - `input=` / `inputSafeBottom=` — input view bounds + its bottom safe-area inset
   - `bottomRowWin=` / `spaceWin=` — bottom-row and spacebar cell frames in **window** coordinates (`x,y,wxh`)
   - `hitBandMaxYWin=` — the lowest `y` the keyboard now accepts, in window coordinates
   - `winH=` / `winSafeBottom=` — window height and inset. `winH - hitBandMaxYWin` is the room left below the hit area before the screen edge.
2. `component=GiellaKeyboardView action=bottomRowBelowBoundsTouch` — one line per touch-down that lands **below** the grid, with `overshoot=` (how far below). These are exactly the taps that were dropped before this PR. Type a session normally, then set the constant just above the overshoots that are real taps.

Suggested calibration: with the current build, tap Apple's spacebar as low as it still responds, then the same spot on Dictus; `hitBandMaxYWin` vs `winH` tells you how much headroom is left to grow the band.

## What I could NOT verify (no device available)

- **No measurement against Apple's keyboard was performed.** The acceptance criterion "measured hit-area delta documented with raw numbers" is not met by this PR — per the scope adjustment in #138 it is now a maintainer step.
- Not verified on device: that a tap below the visible spacebar now inserts a space; the French fast-typing test sentence; long-press / swipe-to-alternative / globe on the bottom row; portrait vs landscape; a device without a home indicator.
- **Structural unknown worth checking first:** whether the band is reachable at all depends on iOS actually placing our input view above the home-indicator strip and hit-testing that strip into it. If `inputSafeBottom`/`winSafeBottom` in the probe show the input view is already flush with the screen bottom, the band is off-screen on that device and the tolerance will do nothing — that would contradict the analysis and should be reported rather than papered over. The probe was designed to answer exactly this.
- Before/after screenshot comparison at identical scale — not done. The geometry argument above is a code argument, not a visual verification.

## Also noted

- While the keyboard is hidden (recording overlay), the input view now accepts the band itself and swallows those touches instead of letting them fall through. Only affects the home-indicator strip; worth a glance during device testing.
- iPad: out of scope, but the same tolerance applies there. Nothing suspicious found in the iPad path; the globe extension is bounded to the last row, which is correct for both layouts.

## Checks

- `xcodebuild -scheme DictusKeyboard` builds (iOS Simulator, Xcode 26.4.1). FluidAudio needed the usual `scripts/patch-fluidaudio-swift5.sh` treatment in my derived data.
- `swiftlint lint` clean — **`.swiftlint-baseline.json` NOT regenerated** (no new violations), so no conflict with the other open PRs.
- DictusCore tests: 425 passed, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard touch handling near the bottom row, making taps just below the key grid more reliable.
  * Expanded the forgiving touch area for the bottom-row keys, including the globe/language key, without changing the keyboard’s visible layout.
  * Preserved existing touch behavior outside the supported bottom-edge area.


Pivii:

## Device validation — read the probe BEFORE testing anything else

The PR flags a structural unknown that determines whether this change can work at all. **Resolve it first**, because everything downstream is meaningless if the answer is no.

### Step 0 — is the band even on screen?

Open the keyboard, then read `action=bottomRowHitArea` in Settings → Logs.

Look at `winH - hitBandMaxYWin`. That is the headroom remaining below the accepted band.

- **Positive** → the band sits inside the strip iOS reserves under the input view. Proceed.
- **Zero or negative** → the input view is already flush with the screen bottom, the band is off-screen, and the tolerance can do nothing on this device. **That would contradict the diagnosis in #138 and must be reported rather than worked around.** In that case the missed taps have another cause and the issue needs reopening for analysis, not a bigger constant.

### Step 1 — measure, then set the constant

Type normally for a minute or two, including the fast French sentence from the issue. Each tap landing below the grid emits `action=bottomRowBelowBoundsTouch` with `overshoot=`.

Those lines are precisely the taps that were silently dropped before. Their distribution is the measurement: set `KeyboardHitArea.bottomRowDownwardTolerance` to cover the bulk of them, staying under the headroom from Step 0. The shipped `12.0` is a constraint-derived starting point, explicitly **not** a measurement.

Then compare against Apple's keyboard: tap progressively lower on the spacebar in both keyboards and note where each stops responding.

### Step 2 — the functional checks

- [ ] A tap below the visible spacebar inserts a space.
- [ ] The full bottom row benefits, not the spacebar alone — check `123`, emoji, return.
- [ ] **The globe key specifically.** It is handled by a separate transparent button that bypasses the normal touch path, and it got its own fix here. It is the most likely thing to have broken.
- [ ] Long-press and swipe-to-alternative still work on the bottom row.
- [ ] Fast typing of `ça vient surtout d'une latence à la frappe` produces correctly separated words, repeatedly.
- [ ] Landscape, where the reserved inset is smaller (~21pt vs ~34pt).
- [ ] A device without a home indicator, if one is available.

### Step 3 — the risk this change introduces

**Not in the PR's own list, and it is the one that could regress something that works today.**

The band extends acceptance into the strip where the system home-indicator gesture lives. Verify explicitly:

- [ ] Swipe up from the very bottom of the screen with the keyboard open — going home still works, and is not swallowed or delayed.
- [ ] Swipe up to reach the app switcher with the keyboard open.
- [ ] No new delay on bottom-row key presses. This keyboard has prior history with window-level gesture recognisers introducing latency on edge keys, so a regression here would be familiar territory.

If any of these degrade, lowering the constant is the first lever — the mechanism scales down cleanly, and `0` restores exact pre-fix behaviour.

### Step 4 — nothing rendered moved

The "no geometry change" claim is a code argument, not a visual check — the author was explicit that no before/after screenshots were taken.

- [ ] Before/after screenshots at identical scale: key heights and grid position are unchanged.
- [ ] Cross-check `computeKbHeight` in the logs against a pre-change build — `grid=` and `total=` must be identical.
- [ ] Side by side against Apple's keyboard, visible key height still matches (#117 did not return).

Pivii:
Closing without merging. Device measurement shows the mechanism cannot work, and the underlying premise is false.

## The measurement

Two device sessions on `9fbc13a`, the second with ten deliberate taps low on the spacebar. Across every sample from the `bottomRowHitArea` probe, in both the settled and unsettled layout states:

```
hitBandMaxYWin=944.0   winH=932.0    ->  -12.0
hitBandMaxYWin=288.0   winH=276.0    ->  -12.0
```

The gap is always exactly `-12.0`, the tolerance itself. The bottom of the key grid coincides with the bottom of the keyboard's own window, so the accepted band lies entirely outside that window. UIKit begins hit testing at the window and never routes an exterior point to it, so the `point(inside:)` overrides in `GiellaKeyboardView` and `KeyboardInputView` are never consulted for those points.

`inputSafeBottom=0.0` and `winSafeBottom=0.0` in every sample: there is no reserved strip below the input view to extend into.

This is the structural unknown the PR description flagged as capable of contradicting the diagnosis. It did.

## Direct proof, not inference

`bottomRowBelowBoundsTouch` fires from `touchesBegan` for any touch landing below the grid. If the band worked, a low tap would be accepted, `touchesBegan` would run, and the probe would emit.

**It emitted zero times across both sessions**, including the one with deliberate low taps. No additional touch was ever accepted. The three changes in this PR are inert code; the constant could be 12, 30 or 0 with identical user-visible behaviour.

## Why no version of this can work

The maintainer reproduced the underlying symptom and tested it against competing third-party keyboards: they behave identically. Apple's keyboard responds lower because it is not an app extension and is not bound by the same window. This is an iOS platform limitation for keyboard extensions, not a defect in our hit testing. See #138.

## What was learned, and what is kept

The instrumentation did its job: three log lines disproved the theory before anyone spent days tuning an inert constant. That value is now spent, so the probes are not being merged — keeping them would add permanent log volume for a question that is answered, which works against #255.

The branch `fix/138-bottom-row-hit-area` is left in place should the instrumentation ever be needed again.

Caveat on the conclusion, stated for the record: measured on one device in portrait only. No landscape, no device without a home indicator, no iPad. Given that competing keyboards show the same limit, another configuration granting space below the window is unlikely but not excluded.

PR #254, opened 2026-07-28
feat(keyboard): reflect model-loading state in the mic button (#250)

### LINKED ISSUE #250: Mic button stays enabled while the model is loading — user taps, gets a red error instead of seeing it is not ready

## Not a bug — a reactive UX where a preventive one is possible

The keyboard already refuses mic taps while the model is loading, and it already tells the user why. The problem is purely *when* it tells them: the mic button looks fully available, the user taps it, and only then does a red error appear. The user learns the state by failing.

Reproduced deliberately by the maintainer on 2026-07-28 (iOS 26.5.2, 1.8.0 (23), `openai_whisper-medium`): force-quit Dictus, then immediately try to dictate from the keyboard.

Message shown: **"Modèle en cours de chargement. Ouvrez Dictus."** in red, on the keyboard toolbar, after the tap.

## What the log shows

```
20:05:29  modelLoadStateChanged from=ready to=loading reason=init-preload
20:05:36  dictationDeferred reason=keyboard refused — model load in flight
20:05:38  dictationDeferred reason=keyboard refused — model load in flight
20:05:39  keyboardMicTapped                                  ← accepted
20:05:44  modelLoadStateChanged from=loading to=ready reason=cold-start-success
20:05:46  transcriptionCompleted duration=1845ms words=3
```

The refusal window spanned roughly ten seconds in this capture. Two taps were rejected before one was accepted. Everything worked as designed — the guard did its job.

**This is model load latency, not a defect.** It is pronounced on `whisper-medium`; the maintainer notes Parakeet loads fast enough that the window is barely noticeable. There is nothing to "fix" about the loading itself here.

## Why the change is cheap

The mechanism is already in place and does not need inventing:

- `ModelLoadState` (`idle` / `loading` / `ready`) already lives in DictusCore and is written to App Group UserDefaults precisely so the keyboard extension can refuse mic taps during a load. It was introduced for #144.
- The keyboard **already reads it** — that is what produces the refusal above.
- A disabled mic-button presentation already exists in the keyboard, currently used for the missing-Full-Access case, with a tap-to-explain affordance.

So the state is available and the visual pattern exists. What is missing is only wiring the button's *appearance* to a state the keyboard already consults before the tap.

## Design considerations to resolve

**Avoid flicker.** On fast paths (Parakeet, warm app) the loading window can be very short. A button that greys out for a couple hundred milliseconds on every cold start reads as instability, not as feedback. A short delay before showing the not-ready state is likely needed, so that only genuinely slow loads surface it.

**Do not create a permanently dead button.** The shared state is written by the app. If the app is force-quit part-way through a load, the last written value can remain `loading` with no running process to move it on. A disabled button driven naively by that value could stay disabled indefinitely, which is far worse than today's behaviour. The not-ready presentation needs a bound after which it stops trusting a stale `loading` value.

**Keep the tap useful.** Even when not ready, tapping should still explain and offer the action that resolves it, rather than doing nothing. The existing message already names the right action — opening Dictus lets the load proceed.

**Do not silently drop the existing guard.** The refusal path stays; this only changes what the user sees *before* they reach it.

---

## Agent Brief

**Category:** enhancement
**Summary:** Reflect model-loading state in the keyboard mic button before the tap, instead of only rejecting the tap afterwards.

**Current behavior:**
While the active model is loading, the keyboard's mic button renders in its normal, fully available state. Tapping it is refused and a red message appears on the toolbar. The user discovers the state only by acting on it, and the red styling frames a normal transient condition as an error.

**Desired behavior:**
When the shared model-load state indicates a load in flight and that load is slow enough to be worth surfacing, the mic button presents a visually distinct not-ready state. Tapping it still explains the situation and points at opening the app. Once the state becomes ready, the button returns to normal without requiring the keyboard to be dismissed and reopened.

**Key interfaces:**
- `ModelLoadState` in DictusCore and its App Group key — read-only here. Do not change its semantics; other consumers depend on it (#144).
- The keyboard's existing mic-tap refusal path — unchanged; this adds presentation ahead of it, it does not replace the guard.
- The existing disabled mic-button presentation used for the missing-Full-Access case — reuse or generalise it rather than introducing a second, divergent disabled style.

**Acceptance criteria:**
- [ ] With `whisper-medium` selected: force-quit the app, open the keyboard, and the mic button shows the not-ready state instead of looking available.
- [ ] Tapping the not-ready button explains the state and offers the resolving action; it never silently does nothing.
- [ ] The button returns to normal automatically once the model is ready, with the keyboard left open throughout.
- [ ] On a fast load (Parakeet, or a warm app) the not-ready state does not flash. State the chosen threshold and the reasoning in the PR.
- [ ] A stale `loading` value left behind by a force-quit cannot disable the button indefinitely. Verify by force-quitting mid-load and confirming the button recovers.
- [ ] The existing refusal path and its message still function; this change does not remove the guard.
- [ ] Full Access disabled still shows its own message, not the model-loading one.

**Out of scope:**
- Reducing model load time, prewarm strategy, or model choice.
- Anything in #249 (network resolution of the model) — separate cause, separate issue.
- Redesigning the keyboard toolbar generally.

## Related

- #144 — introduced `ModelLoadState` and the mic-tap refusal this builds on
- #249 — a genuinely broken model load path; distinct from this normal loading window
- #123 — audio route; unrelated

Pivii:
Closes #250.

While the transcription model loads, the keyboard's mic button looked fully available. The user tapped, the tap was refused, and a red message appeared. The state is now visible **before** the tap.

## What changed

**DictusCore**
- `ModelLoadStateStore` — single owner of the App Group contract for `ModelLoadState`. It writes the state together with a new `modelLoadStateChangedAt` timestamp, and reads the pair back. The tri-state semantics from #144 are untouched: `loading` still means "mic taps MUST be refused".
- New Darwin notification `modelLoadStateChanged`, posted by the app on every real transition, so the keyboard hears about a load starting or finishing while it is on screen.
- `MicAvailabilityPolicy` — the pure, time-injectable decision (`available` / `modelLoading`). All the risk of this feature lives in this one function, which is why it is pure and unit-tested.

**DictusApp**
- `setModelLoadState` now persists through the store (state + timestamp) and posts the Darwin ping. Its guard, logging and in-process `NotificationCenter` post are unchanged.

**DictusKeyboard**
- `KeyboardState.micAvailability` is derived from the shared state, refreshed on the Darwin ping, on keyboard appearance, on a refused tap, and by a 0.25s timer.
- `ToolbarView` now has **one** "mic unavailable" presentation, used by both the missing-Full-Access case and the new model-loading case, rather than a second divergent disabled style. The model-loading variant keeps the same dimming but stays tappable, so the tap still explains the wait.
- The refusal message is informational (secondary colour) rather than red — a model still loading is a normal transient wait, not a failure — and it now auto-clears after 3s like the error messages already did.

The refusal guard in `startRecording()` is untouched and still authoritative. This adds presentation ahead of it; it does not replace it.

## Flicker threshold: 0.6s

The not-ready state is only shown once a load has already been running for **0.6s**.

Reasoning: it has to be long enough to swallow the fast paths — Parakeet and a warm app resolve well inside it, and it also absorbs the few milliseconds of cross-process UserDefaults propagation lag — while staying under the ~1s mark at which a wait stops feeling instantaneous. A load that resolves faster than that never needed announcing. A load that outlasts it (whisper-medium was a ~10s window in the issue's capture) is surfaced essentially as soon as the user looks at the keyboard, because the timestamp lets us count from when the *app* started loading, not from when the keyboard happened to look.

I have not measured the Parakeet load window myself — 0.6s is chosen from the perception argument plus the ~10s figure in the issue, and it is a one-line constant if it feels wrong on device.

## Bounding the stale `loading` value

Two layers, because this is the failure mode that would be worse than the status quo:

1. **The timestamp does the real work.** `modelLoadStateChangedAt` is written next to the state, so the keyboard can compute the actual age of a `loading` value. A load stranded by a force-quit an hour ago ages out immediately and never dims the button — even on a keyboard process that has just launched and has no memory of it.
2. **A 30s cutoff as the backstop**, for the case where no timestamp is stored (a `loading` value written by a build older than this PR). Past 30s the policy stops trusting the value and the button returns to normal. 30s is 3x the slowest window observed in the issue. A genuinely slower load (a first-run turbo compile can take minutes) simply loses the pre-tap hint and behaves exactly as it did before this PR — the tap is still refused and still explains itself. That is the safe direction to be wrong in.

The same 30s bound also caps the re-evaluation timer's lifetime, so the keyboard never leaves a timer running because of a value nobody will update.

## Verification

Automated, on my machine:
- `swift test` in `DictusCore` — 439 tests pass, including 14 new `MicAvailabilityPolicyTests` covering: fast load never surfaces, slow load surfaces at the threshold, the ~10s whisper-medium window stays surfaced, the cutoff, a value stranded an hour ago, unknown age, and a backwards device clock.
- `xcodebuild build -scheme DictusApp -destination 'generic/platform=iOS Simulator'` — **BUILD SUCCEEDED**.
- `swiftlint lint --no-cache` — **0 violations**. `.swiftlint-baseline.json` is **not** touched, so there is no conflict with #251. This constrained the change: the two new statements in `DictationCoordinator` are a net-zero line change inside that class body, because the baseline matches on the exact line count in the violation text.

Against the acceptance criteria:

| Criterion | Status |
|---|---|
| whisper-medium, force-quit, keyboard shows not-ready | **Not verified — needs a device.** Logic verified by unit test; visual behaviour is yours to judge. |
| Tapping the not-ready button explains and offers the action | Verified by code path (unchanged guard → message naming "open Dictus"), not on device. |
| Returns to normal automatically, keyboard left open | Verified by code path (Darwin ping + 0.25s timer re-read). **Not observed on device.** |
| No flash on a fast load | Threshold reasoned + unit-tested. **The actual feel on Parakeet is unverified.** |
| A stale `loading` cannot disable the button indefinitely | Unit-tested for both the timestamped and the untimestamped path. **Not reproduced on device.** |
| The existing refusal path still functions | Guard is byte-for-byte unchanged apart from the message type. |
| Full Access disabled still shows its own message | Unchanged branch; it now renders through the shared dimmed-mic builder, same visual result (`.disabled(true)`, `opacity 0.4`). |

**I have no physical device and ran no simulator UI session.** Everything above the table is compiler-, linter- and unit-test-verified; everything about how it *looks and feels* is unverified and needs the on-device pass.

Two things worth a specific look on device:
1. The dimmed mic keeps `AnimatedMicButton`'s idle glow animation under the 0.4 opacity. It may read as "pulsing while loading" (good) or as "dimmed but still animating" (odd) — I could not judge that without seeing it.
2. Whether 0.6s is the right threshold on the Parakeet path.

## Summary by CodeRabbit

* **New Features**
  * Added dimmed microphone feedback during model loading, including delayed surfacing and automatic recovery when loading becomes stale.
  * Added structured (typed) info/error status messages and updated toolbar rendering to use them.
  * Keyboard now refreshes mic availability promptly as model load state changes.

* **Bug Fixes**
  * Improved resilience for interrupted/stalled model loads, including aging behavior for preexisting states.
  * Prevented older status-message timers from clearing newer messages.

* **Tests**
  * Added coverage for mic availability timing and staleness logic.


Pivii:

## Review notes and device validation

The policy is pure, time-injectable and unit-tested, with both risky constants documented in place — that is the right shape for this feature. Three things worth attention before merge.

### 1. Merge order: land #251 first

This PR deliberately keeps `DictationCoordinator`'s class body at a net-zero line change so `.swiftlint-baseline.json` stays untouched — the baseline matches on exact line counts. That constraint is real but **temporary**: PR #251 restructures the same class, moving three methods into an extension and taking the body from 644 to 525 lines, which regenerates the baseline and very likely removes that violation entry entirely.

So merge **#251 first**. Afterwards the net-zero constraint here becomes moot and the file is no longer carrying a hidden invariant. Merging this one first would leave a fragile line-count dependency in place for whoever edits that class next.

Applies to #253 as well — it also leaves the baseline untouched. Sequence: **#251 → #253 → #254**, updating each branch in turn.

### 2. Is 30s the right stale cutoff? Worth checking, not assuming

`staleLoadCutoff = 30` is justified as 3× the ~10s whisper-medium load from the #250 capture. That reasoning holds for the loads observed **on this path** — cold-start loads in the captured logs resolve in seconds.

But two data points sit near or beyond it:

- A `modelCompilationCompleted duration=21870ms` was observed for whisper-small in a device log from 2026-07-28 — 21.8s, already most of the budget.
- #144 documents Whisper turbo compiles taking roughly two minutes.

**The question to answer is which flows actually write `ModelLoadState = .loading`.** If compilation during download/prewarm goes through the model manager and never sets it, 30s is comfortable. If it does set it, the cutoff is far too short: on a slow compile the button would silently revert to normal mid-load, which is precisely the case this PR exists to fix.

The failure is bounded either way — it degrades to pre-#250 behaviour, not to something broken — but it is worth resolving rather than leaving to chance. One constant either way.

### 3. Device validation

Everything about how this *looks* is unverified — the author was explicit that no device or simulator UI session was run.

- [ ] `whisper-medium`, force-quit, open the keyboard: the mic shows the not-ready state **before** any tap.
- [ ] Tapping it explains the wait and names the resolving action. It never silently does nothing.
- [ ] The button returns to normal on its own once the model is ready, keyboard left open throughout.
- [ ] **Parakeet: no flash.** The 0.6s threshold is a reasoned choice, not a measurement — the author did not time the Parakeet window. If it flickers, that constant is the knob.
- [ ] Force-quit mid-load, then reopen the keyboard: the button must not be stuck dimmed. This is the trap the timestamp is there to prevent.
- [ ] Full Access disabled still shows **its own** message, not the model-loading one — the two now share a dimmed-button builder.
- [ ] The refusal message is now informational rather than red, and auto-clears after 3s. Check the tone reads as "not yet" rather than "error".
- [ ] The dimmed mic keeps `AnimatedMicButton`'s idle glow under 0.4 opacity — flagged by the author as possibly reading as an odd pulse. Judgement call, needs eyes.

### Incidental finding worth its own issue

`scripts/patch-fluidaudio-swift5.sh` only patches checkouts under the default `~/Library/Developer/Xcode/DerivedData`, so a build using `-derivedDataPath` is missed and fails on the known Xcode Swift-5-mode issue. This bit the author, who patched by hand. It will bite any agent or CI job using a custom derived-data path. Unrelated to this PR; the script was already hardened once in #212.

Pivii:
## Measured: the 30s stale cutoff is too short. Do not merge as-is.

The review note above asked which flows actually write `ModelLoadState = .loading`, and whether 30s is enough. Device logs from 2026-07-29 answer both. The answer is worse than the 21.8s figure that prompted the question.

**Two flows write `.loading`, and both exceed the cutoff:**

```
09:52:29  modelLoadStateChanged from=ready to=loading  reason=selectModel-proactive
09:52:57  modelLoadStateChanged from=loading to=ready  reason=selectModel-proactive-success
          -> 28s   (whisper-medium)

10:57:58  modelLoadStateChanged from=ready to=loading  reason=selectModel-proactive
10:58:35  modelLoadStateChanged from=loading to=ready  reason=selectModel-proactive-success
          -> 37s   (whisper-medium, second session)

10:55:31  modelLoadStateChanged from=ready to=loading  reason=init-preload
10:57:17  modelLoadStateChanged from=loading to=ready  reason=init-preload-success
          -> 106s  (Parakeet, first launch after install)
```

Builds `1.8.0 (24)` at `79391f1` and `63f0e71`. Three samples, three different durations, all at or beyond `staleLoadCutoff = 30`.

**Consequence as written:** on a load slower than 30s the mic button silently returns to its normal appearance mid-load, which is precisely the situation this PR exists to fix. The degradation is bounded (it falls back to pre-#250 behaviour, the tap is still refused and still explains itself), so nothing breaks, but the feature stops working exactly when it is most needed.

**Why simply raising the constant is probably the wrong fix.** 106s is not an outlier: it is a first-launch preload, and #144 documents Whisper turbo compiles taking roughly two minutes. A cutoff generous enough to cover those is generous enough that a genuinely stranded `loading` value (force-quit mid-load) would dim the button for minutes. The cutoff's whole purpose is to bound that case.

Worth considering instead: an app-liveness signal rather than a fixed delay. The keyboard already learns about the app through Darwin notifications, so "is the app still working on it" is answerable more directly than "has enough time passed". The timestamp mechanism this PR introduces stays useful either way, it is the fixed 30s bound that does not survive contact with the measurements.

The 0.6s surface delay is unaffected by any of this and is not in question here.

Recorded now so the measurement is not lost. To be decided at triage, before merge.

Pivii:
Closing in favor of #262. This PR implements the fixed 30-second stale-loading cutoff, but device measurements and the existing cold-start flow show that the problem should be solved at the interaction boundary: allow cold-start taps to open Dictus and keep the request pending while the model prepares. The PR branch remains available for reference.

Issue #256, opened 2026-07-29
Idle warm-state release is only armed after a dictation, so the audio engine can run for hours


Pivii:
## Summary

`UnifiedAudioEngine.scheduleIdleRelease()` is called from exactly one place: `collectSamples()`. That runs only at the end of a completed recording. Every other path that leaves the engine warm and idle arms no timer at all, so the warm state is never released and the engine keeps running until the process dies.

Observed on a device: **the audio engine ran continuously for 8 hours 18 minutes overnight**, with no dictation in that window. This defeats the entire purpose of #106 Phase B and carries a battery cost the user never opted into.

It also causes #257, because the standby Live Activity is dismissed by the `dictusWarmStateReleased` notification that this release posts. No release means the pill stays up until ActivityKit's 8-hour ceiling ends it.

## Evidence

Device log, build `1.8.0 (23)` at `3dff5d8@feature/239-auto-polish-prompt`:

```
28/07 21:37:06  appLaunched                                  engineRunning=false
28/07 21:37:06  liveActivityStarted    id=7239D025-...
28/07 21:37:06  engineWarmUpAttempt    context=didBecomeActive
28/07 21:37:12  engineWarmUpSuccess    context=unifiedEngine-warmUp
28/07 21:37:21  appDidEnterBackground
                        (no dictation for the next 8h18m)
29/07 05:55:27  engineStateSnapshot    engineRunning=true sessionConfigured=true
29/07 05:55:27  engineWarmUpSuccess    context=didBecomeActive-already-running
```

The engine was warmed by `didBecomeActive`, not by a recording, so `collectSamples()` never ran and no timer was ever armed.

The contrast is in the same log. `warmStateReleased` appears exactly twice all evening, and both follow a real dictation:

```
28/07 20:25:27  warmStateReleased  idleSeconds=602
28/07 20:38:09  warmStateReleased  idleSeconds=600
```

The mechanism works correctly when it is armed. It is simply not armed on any other path.

## Second path: the timer does not survive suspension

A separate device log, build `1.8.0 (24)` at `79391f1`, shows the timer armed and still not firing:

```
09:53:50  engineCollectResult          <- timer armed here, due ~10:03:50
09:55:35  appDidEnterBackground
             (56 minutes of no log output at all)
10:51:56  engineWarmUpAttempt  context=didBecomeActive
10:51:56  engineWarmUpSuccess  context=unifiedEngine-warmUp    <- cold start, not "already running"
```

[...]

PR #259, opened 2026-07-29
fix(live-activity): reconcile against ActivityKit's activityState (#257)

### LINKED ISSUE #257: Live Activity recovery never checks activityState, so a system-ended activity is treated as healthy forever
## Summary

`LiveActivityManager` never reads `Activity.activityState`. Not once: `grep -rn "activityState" --include="*.swift" .` returns nothing across the whole repo. Recovery decisions are made from the manager's own `currentPhase` plus a check that the activity id still appears in `Activity.activities`.

Neither of those distinguishes a live activity from one the **system has already ended**. ActivityKit ends every Live Activity after 8 hours. The ended activity stays in `Activity.activities` (its state only becomes `.dismissed` when the UI is actually removed, which can be 4 more hours). So the manager sees the id, concludes the activity is healthy, and never recreates it.

The user-visible result: the Dynamic Island disappears and never comes back. Reinstalling the app is the only recovery the maintainer found.

## Evidence

Device log, build `1.8.0 (23)` at `3dff5d8@feature/239-auto-polish-prompt`:

```
28/07 21:37:06  liveActivityStarted           id=7239D025-EF14-4B81-85BD-BA1472730A73
                        (activity runs continuously, see #256)
29/07 05:55:27  liveActivityStandbySkipped    reason=ensureAlive-healthy-standby isEnabled=true areActivitiesEnabled=true
29/07 05:55:34  liveActivityStandbySkipped    reason=alreadyStandby
29/07 05:55:39  liveActivityStandbySkipped    reason=ensureAlive-healthy-standby
```

**8 hours 18 minutes and 21 seconds** between the activity starting and the first recovery attempt. ActivityKit's hard ceiling is 8 hours, so the system had already ended it. Three consecutive recovery attempts, three skips, no recreation. The Dynamic Island was gone from the screen throughout.

We also know the id was still present in the system list: had it been absent, the first branch of `ensureActivityAlive()` would have logged `liveActivityFailed context=ensureAlive error=activity <id> gone from system`. It did not. The activity was listed and `.ended`, which is exactly the state the code cannot observe.

## Root cause

`LiveActivityManager.ensureActivityAlive()`:

```swift
// Detect externally-ended activities (e.g., StopStandbyIntent bypasses manager).
if let current = currentActivity,
   !Activity<DictusLiveActivityAttributes>.activities.contains(where: { $0.id == current.id }) {
    // treat as dead, reset to .idle
}

guard currentActivity == nil || currentPhase == .idle else {
    logStandbySkipped(reason: "ensureAlive-healthy-\(currentPhase.rawValue)", ...)
    return
}
```

The first check catches an activity that has left the list entirely. It cannot catch one that is still listed but no longer live. The second guard then short-circuits on `currentPhase == .standby`, which is the manager's own belief and is never reconciled against the system.

Relevant ActivityKit semantics, from Apple's documentation:

- *"A Live Activity can remain active for up to eight hours, after which the system automatically ends it and removes it from the Dynamic Island. It may persist on the Lock Screen for an additional four hours."*
- `ActivityState.ended`: *"The Live Activity is visible, but a person, the app, or the system ended it, and it won't update its content anymore."*
- With the default dismissal policy, *"The ActivityState doesn't change to ActivityState.dismissed until a person or the system removes the Live Activity user interface."*

## Relationship to #256

These are two faces of one incident and should be read together, but they are separate fixes in separate files.

#256 is what lets a standby activity run long enough to hit the 8-hour ceiling in the first place: the idle release that would have dismissed it is never armed unless a dictation completed. **Fixing #256 makes this bug far harder to reach**, because a standby pill would normally be dismissed after about 10 minutes of idle.

This issue remains worth fixing regardless, as defence in depth. A user who genuinely dictates across an 8-hour span, or any future path that keeps an activity alive, would hit the same wall. And the underlying flaw, trusting local belief over system truth, is the kind that resurfaces.

#256 also describes a third path this issue is the natural home for: when the app process is killed while a standby activity is up, the activity outlives the process and nothing can end it.

## What needs to change

1. **Treat any activity whose `activityState != .active` as absent.** That single check, applied in `ensureActivityAlive()` before the healthy-phase guard, closes the reported bug.
2. **Reconcile `currentPhase` against system truth, not only against itself.** The healthy-phase guard should not be reachable while the underlying activity is `.ended`, `.dismissed` or `.stale`.
3. **Log `activityState` in the existing standby-skipped probe.** `logStandbySkipped` already records `isEnabled` and `areActivitiesEnabled`. Adding the activity's real state makes this class of failure self-diagnosing from an exported log, the way `localModelResolved` did for #249. Worth doing even before the fix lands.

## Acceptance criteria

- [ ] An activity ended outside the manager (system ceiling, user swipe, Stop intent) is detected on the next `ensureActivityAlive()` and a fresh standby activity is created.
- [ ] `ensureAlive-healthy-standby` is never logged while the underlying activity is not `.active`.
- [ ] The standby-skipped probe records the real `activityState`.
- [ ] Unit coverage for the decision: given a non-active state, the activity must be treated as absent. The 8-hour wait must not be required to test the fix.
- [ ] Device check: end the activity by swiping it away from the Lock Screen, foreground the app, and confirm the pill returns.

## Open decisions

- **Should a system-ended activity be recreated immediately?** Recreating it starts a fresh 8-hour window, so an app left open indefinitely would recreate a standby pill forever. That may be undesirable, and ActivityKit applies its own rate limits to repeated requests. The conservative alternative is to clear the stale reference so that the *next dictation* can bootstrap cleanly, without immediately putting a new pill on screen. This decision should be made before implementing, since it changes the acceptance criteria above.
- Whether `.stale` should be treated as absent or merely refreshed with an update.

## Not in scope

- The idle release that would have prevented this, which is #256.
- Any redesign of the standby pattern itself.

## How this was found

During the #249 device validation. Reported by the maintainer as "the Dynamic Island disappeared overnight and only a reinstall brings it back".

Pivii:
Closes #257

## Root cause, in my own words

`LiveActivityManager` decided "is my Live Activity still alive?" by asking whether its id still appears in `Activity.activities`. That question cannot detect the failure it was written to detect.

ActivityKit ends every Live Activity after 8 hours. Ending it does **not** remove it from `Activity.activities` — with the default dismissal policy the state stays `.ended` (still visible, never updating again) until the UI is actually removed, which can be four more hours. So the id was still there, the presence check passed, `currentPhase` was still `.standby` because nothing ever contradicted it, and `ensureActivityAlive()` short-circuited on its healthy-phase guard. Three recovery attempts, three skips, no recreation. The manager's belief was never reconciled against the system, and the one value that could have told it the truth — `activityState` — was read nowhere in the repo (`grep -rn "activityState" --include="*.swift" .` returned nothing).

I checked the issue's claims against the code and they hold. Two things it does not mention, both of which the fix would have been incomplete without:

1. **The bug is not confined to `ensureActivityAlive()`.** `startStandbyActivity()`'s orphan-adoption path (`systemActivities.first`) had the same flaw: it would adopt an `.ended` activity as a healthy orphan and return `.orphanRecovered`, whose `activityAvailable` is `true`. The bootstrap path the maintainer's decision relies on would have handed the same corpse straight back.
2. **`ensureActivityAlive()` alone does not reach the reported user.** Its only caller is `didBecomeActive`. A keyboard dictation never foregrounds the app, so for a user who dictates without opening Dictus — the normal Dictus flow — the new check would never have run. `transitionToRecording()` also skips its auto-bootstrap because `currentPhase` is `.standby`, not `.idle`, so every `activity.update()` goes to a dead activity and is silently dropped. Reconciliation therefore runs at three entry points, not one.

## What changed

**`DictusCore/LiveActivityLiveness.swift` (new)** — a plain enum mirroring `ActivityState`, with a `decision` property (`keep` / `refresh` / `treatAsAbsent`). It mirrors rather than uses `ActivityState` because ActivityKit is iOS-only while DictusCore's test target runs on macOS; mirroring is what makes the rule testable in milliseconds instead of 8 hours. It has an `unknown` case because `ActivityState` is not frozen.

**`LiveActivityManager`**
- `liveness(of:)` maps `ActivityState` at the boundary, with `@unknown default`.
- `discardActivityIfNotLive(context:)` — on a non-live activity: cancels the auto-dismiss task and watchdog, clears `currentActivity`, forces the phase to `.idle`, logs `liveActivityEnded reason=<context>-notLive-<state>`, and calls `activity.end(dismissalPolicy: .immediate)` so it actually leaves `Activity.activities` instead of lingering in `.ended`. It deliberately does **not** call `startStandbyActivity()`.
- Called from `ensureActivityAlive()` (which returns immediately after clearing, so no pill appears at that moment), `startStandbyActivity()` (before its `alreadyStandby` fast path, which would otherwise report a corpse as healthy standby on every background transition), and `transitionToRecording()` (so the auto-bootstrap can actually fire).
- `adoptableSystemActivities()` ends non-live system activities and excludes them from orphan adoption.
- `ensureAlive-healthy-<phase>` is now only reachable when the activity is `.active`; the dead and stale cases return earlier under their own reasons (`ensureAlive-clearedNotLive`, `ensureAlive-refreshed-<phase>`).

**`logStandbySkipped` / `LogEvent.liveActivityStandbySkipped`** now carry the real `activityState`. It defaults to reading the held activity's own state, because every existing skip site skips *because of* the activity it holds; only the dead path passes it explicitly, having just cleared the reference. The line becomes `reason=… isEnabled=… areActivitiesEnabled=… activityState=ended`. The #257 signature — both booleans `true` while the activity was already dead — is no longer indistinguishable from a healthy skip.

### I verified the bootstrap path really works from `.idle`

As instructed, not taken on faith. From `.idle` with `currentActivity == nil`, `transitionToRecording()` logs `idle -> recording-bootstrap` and calls `startStandbyActivity()`, which passes the enabled checks, finds no adoptable orphan (the dead one having been ended and filtered out), and reaches `Activity.request`. One caveat, pre-existing and unchanged here: `Activity.request` fails from the background, which is why `startStandbyActivity()` is also called on the `.background` scene transition. So after a clearing, the pill realistically returns on the next background transition rather than instantly.

## How I resolved the `.stale` question

**`.stale` is refreshed with a content update, not treated as absent.** This turned out not to be a close call, and treating it as absent would have been a severe immediate regression.

`.stale` means the content is past its `staleDate` — the activity is still visible and still accepts updates. Dictus sets `staleDate` to `now + 30s` (`staleInterval`) and **nothing updates a standby activity** while it sits in standby. A perfectly healthy standby pill is therefore `.stale` from 30 seconds after it appears, permanently. Following the issue description's "treat any activity whose `activityState != .active` as absent" literally would have torn down every standby pill half a minute after creating it.

So the issue description is wrong on that point, and I have said so in the code comment rather than leaving it implicit. Refreshing costs one ActivityKit update per app foreground and clears the stale mark.

`unknown` also maps to `refresh`: an unrecognised future state is not evidence of death, and tearing down a live activity is the worse error; an update on a dead activity is a harmless no-op and the next reconciliation will see `.ended`. **That one is a judgement call**, unlike the `.stale` decision.

## Verified, with real numbers

- `cd DictusCore && swift test` — **445 tests, 0 failures** (435 before; 10 new in `LiveActivityLivenessTests`).
- `swiftlint lint --strict --no-cache` — **0 violations, 0 serious in 146 files**.
- `xcodebuild build -scheme DictusApp -destination 'generic/platform=iOS Simulator'` — **BUILD SUCCEEDED**.

### SwiftLint baseline: regenerated, and exactly what changed

`.swiftlint-baseline.json` **is** modified here. I tried to avoid it and could not; evidence rather than assertion:

Baseline entries for length rules embed the line count in their match key. `LiveActivityManager.swift` carries a pre-existing `type_body_length` entry reading "currently spans 491 lines". I probed it empirically: adding **two** code lines to that class body re-reports the pre-existing warning as new and `--strict` fails. Any non-zero change to the body forces regeneration, so "keep the change small enough" is not reachable — only "exactly zero" would be, which is not compatible with adding real logic. `.swiftlint.yml`'s own comment prescribes regeneration for precisely this case.

I did shrink the blast radius: `LogEvent`'s enum body also tripped its baselined entry, so I kept the new case and its message on single lines (`line_length` is disabled in this repo), returning that body to zero net growth. Its entry is untouched.

Semantic diff, old vs new:

- **45 entries before, 45 after.**
- **Exactly one entry changed**, and only its count:
  - removed: `type_body_length` / `DictusApp/LiveActivityManager.swift` / `class LiveActivityManager {` / "currently spans **491** lines"
  - added: same rule, same file, same declaration / "currently spans **558** lines"
- Same rule identifiers before and after (`cyclomatic_complexity`, `file_length`, `for_where`, `force_unwrapping`, `function_body_length`, `function_parameter_count`, `large_tuple`, `nesting`, `opening_brace`, `type_body_length`).
- All entries remain `warning`. **No rule was disabled, no violation suppressed, no entry added or removed.**

⚠️ **Conflict warning:** the PR for #256 also regenerates this baseline (for its own `UnifiedAudioEngine` count). The file is a single line of JSON whose key ordering is not stable across runs, so the two will conflict as a whole-line conflict. Whichever merges second should **regenerate after merging** rather than hand-merge the JSON, then re-run `swiftlint lint --strict` to confirm 0 violations.

## What I could NOT verify

Stated plainly, because it matters more than the green checks above:

- **The actual reported scenario is unverifiable by me.** I have no physical device, and reproducing it as reported means waiting out ActivityKit's 8-hour ceiling. Nothing here has been observed fixing the overnight symptom.
- **No behaviour involving a real `ActivityState` was executed at all.** Live Activities do not run in unit tests, and `LiveActivityManager` is not unit-testable (a `@MainActor` singleton wrapping ActivityKit). The 10 new tests cover the *decision* — non-active means absent, `.stale` means refresh — and the log payload. The `ActivityState` → `LiveActivityLiveness` mapping and every call site in the manager are covered by **compilation and review only**.
- **The Dynamic Island was never rendered.** No simulator run, no visual check; `generic/platform=iOS Simulator` is a build, not an execution.
- **The timing of `end(dismissalPolicy: .immediate)` is assumed, not measured.** I assume the activity leaves `Activity.activities` promptly. `adoptableSystemActivities()` is written so a lagging eviction stays harmless, but I have not observed the lag.
- **The `.stale` refresh path is reasoned, not observed.** That standby sits `.stale` after 30s follows from `staleInterval` plus Apple's documented `staleDate` semantics; I have not watched `activityState` flip on a device.

## Device-test protocol (no 8-hour wait required)

Ending the activity by any other route produces the same non-`.active` state in seconds, so the ceiling can be simulated cheaply.

1. Launch the app, background it so the standby pill appears. Confirm it is there.
2. Swipe the Live Activity away from the Lock Screen (or tap Power in the expanded Dynamic Island, which runs `StopStandbyIntent`).
3. Foreground the app. **Expected:** no new pill at that moment — intended behaviour, not a failure. The exported log shows `liveActivityStandbySkipped reason=ensureAlive-clearedNotLive … activityState=ended` (or `dismissed`), plus `liveActivityEnded reason=ensureAlive-notLive-…`.
4. Background the app again. **Expected:** a fresh standby pill appears.
5. Repeat step 2, then — **without opening the app** — dictate from the keyboard. **Expected:** the Dynamic Island shows recording normally. This is the path that regressed silently before, and the one most worth your attention.
6. Regression check, normal use: dictate several times in a row; standby → recording → transcribing → ready → standby should behave as before, with no duplicate pills. **A pill vanishing ~30 seconds after it appears would mean the `.stale` decision is wrong** — that is the specific thing to watch for.
7. Export the log and confirm every `liveActivityStandbySkipped` line now carries an `activityState=` field.

## Scope

Out of scope and untouched: the idle-release arming in `UnifiedAudioEngine` (#256), which is what lets a standby activity survive long enough to reach the ceiling at all. Also explicitly not done, per the maintainer: terminating the app process to force a clean state. No `exit()`, `abort()` or equivalent was added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Live Activity synchronization with the Dynamic Island and Lock Screen.
  * Automatically clears ended or dismissed activities and refreshes stale activity content.
  * Strengthened recovery when starting standby or recording states.

* **Diagnostics**
  * Added activity-state details to standby-related logs for improved troubleshooting.

* **Tests**
  * Added coverage for active, stale, ended, dismissed, and unknown activity states.


Pivii:
**Merge-order note:** this PR and #258 both regenerate `.swiftlint-baseline.json` (this one for `LiveActivityManager` 491 → 558, #258 for `UnifiedAudioEngine` 537 → 550). The file is a single line of JSON with unstable key ordering, so the two will conflict. Whichever merges second should update its branch and then **regenerate** the baseline with `swiftlint lint --quiet --write-baseline` rather than hand-merging the line, and re-run `swiftlint lint --strict` to confirm. Both were verified independently as 45 entries before and after, one entry changed, no rule disabled.

Issue #271, opened 2026-07-30
Unify keyboard-area presentation behind a single mode value


Pivii:
> *This was generated by AI during triage.*

Extracted from the #241 design session (2026-07-30) as a standalone prerequisite. It needs no design input and de-risks two queued features, so it can ship on its own.

## Agent Brief

**Category:** enhancement
**Summary:** Replace the parallel booleans that decide what fills the keyboard area with a single mode value, owned in one place and read by both the SwiftUI and UIKit layers.

### Current behavior

The keyboard area can show one of three things: the key grid, the emoji picker, or the recording overlay. Which one is showing is currently derived from several independent flags held in **two different layers**:

- a SwiftUI `@State` boolean for the emoji picker, inside the root keyboard view;
- a separate UIKit boolean for the same picker, inside the keyboard view controller;
- the recording overlay, derived from `DictationStatus` plus an active-controller identity check.

The two emoji booleans are kept in step through `NotificationCenter`. Nothing in the type system prevents two of these states from being true at once, and the layout code that reacts to them is written as a chain of negated guards — restoring the collapsed toolbar height is conditional on the emoji flag being false. Any new full-area state must be added to that guard by hand.

That guard runs on **every dictation status change**. A state that is not added to it collapses to toolbar height while still being rendered.

This shape has already produced layout races in this area (#142, and the family around #260 / #261).

### Desired behavior

A single value describes what the keyboard area is presenting. Suggested shape:

```swift
enum KeyboardAreaMode {
    case keys
    case emoji
    case panel      // reserved for #241; no UI in this issue
    case recording
}
```

- It lives in the shared keyboard state object, as the single source of truth.
- The SwiftUI root view switches on it to decide which branch to render, instead of consulting its own boolean.
- The keyboard view controller switches on it to decide the hosting height, the hosting bottom anchor, and whether the key grid is hidden — instead of the current negated-guard chain.
- Mutual exclusion is a consequence of the type, not of coordination between flags.
- `NotificationCenter` round-trips that exist only to keep the two emoji booleans in step are removed.

Include the `.panel` case now, with no UI attached, so #241 adds a view rather than reopening this refactor.

### Key interfaces

- `KeyboardAreaMode` — new enum, owned by the shared keyboard state object, observable so SwiftUI re-renders on change.
- The root keyboard view — replaces its private emoji `@State` with a read of the shared mode.
[...]

Issue #267, opened 2026-07-30
Distinguish the LLM processing stage from transcription in the recording overlay


Pivii:
## Context

Captured during the #79 Smart Modes design session (2026-07-30).

The keyboard's recording overlay animates two states today (`RecordingOverlay.swift` + `KeyboardWaveformView`):

- **recording** — live waveform driven by mic energy
- **transcribing** — a sine sweep

But the LLM stage — polish today, Smart Modes tomorrow — renders as *the same sine sweep as transcription*. The user cannot tell whether the phone is transcribing (fast, a second or two) or running the language model (3–6 s today, and materially longer for a Smart Mode, where generation time scales with output length). A long wait under an animation that says "transcribing" reads as a hang, and invites the desperation tap on the cancel button.

Dictus Desktop already solved this: it has a third, visually distinct animation for the `processing` state.

## Proposal

Introduce a third animated state on iOS for the LLM stage, ported from the desktop idiom.

### The desktop animation, for reference

`~/dev/dictus-desktop/src/overlay/RecordingOverlay.tsx`. It is a **cylon / travelling-peak** sweep — a localized bump that scans back and forth across the bars, deliberately *not* a continuous wave, so it reads as a different family from the transcription sine:

```
BAR_COUNT             = 30
PROCESSING_PHASE_RATE = 0.7 Hz
peakPos  = (barCount - 1) * (1 + sin(2π · phase)) / 2
halfWidth = 10.0, baseline = 0.05, crest = 0.7
bar(i)   = dist >= halfWidth ? baseline
                             : baseline + (crest - baseline) · (0.5 + 0.5·cos(π · dist/halfWidth))
```

Two details from the desktop implementation that are worth carrying over, both of them lessons learned there:

- **No smoothing on this state.** The desktop assigns the targets directly rather than running them through the level smoother, because smoothing erases the grain that makes the sweep legible.
- **An outro.** On hide, the peak is driven back to centre by a cubic Hermite spline with matched starting velocity (300 ms), then collapses to zero (400 ms). Velocity matching is what removed the "abrupt stop" that plain easing produced.

The crest (0.7) matches the transcribing crest so the two states read as the same visual family — different motion, same vocabulary.

### The iOS-side blocker

`DictationStatus` has no state for this:

```swift
case idle, requested, recording, transcribing, ready, failed
```

[...]

Issue #261, opened 2026-07-29
App killed mid-recording leaves a stale recording status: phantom keyboard session for ~34 s, stop taps dropped, audio lost


Pivii:
## Summary

When iOS terminates DictusApp during a background recording, nothing reconciles the dictation state left behind in the App Group. `dictationStatus` stays at `recording` while `dictationSessionID` is gone, so the recreated keyboard extension faithfully restores a recording overlay for a session that no longer exists. The user keeps talking into nothing, stop taps are silently dropped, and the failure surfaces roughly 34 seconds later.

Observed once on device. The audio was lost, which is the part that matters: a long dictation the user believed was being captured.

## Evidence

Device log, build `1.8.0 (24)` at `6ce5cf2@fix/257-live-activity-state-check`, iPhone16,2 / iOS 26.5.2. Reporter had been using the phone for a while outdoors; the device was warm to the touch.

**Setup — a cold-start dictation.** The engine had been released 2 minutes earlier by the normal 10-minute idle timer, so this dictation had to boot the engine *and* load the model while already recording:

```
15:05:53  warmStateReleased idleSeconds=601
15:05:53  audioRouteChanged inputs=none
15:08:00  coldStartURLReceived isColdStart=false isEngineDead=true hasBeenActive=true
15:08:01  audioEngineStarted
15:08:01  statusChanged from=ready to=recording source=coordinator
15:08:02  appDidEnterBackground                      <- user swipes back to the host app
15:08:03  appWhisperKitLoaded model=parakeet-tdt-0.6b-v3   <- model loads DURING the recording
15:08:10  (last <APP> line: waveformSnapshot)
15:08:09  (last <KBD> line: waveformHeartbeat)
```

**Both processes stop emitting at ~15:08:10.** Ten seconds later the app is relaunched from scratch:

```
15:08:20  appLaunched version=1.8.0
15:08:20  deviceCapabilitySnapshot model=iPhone16,2 ramGB=8 availableMB=3369 thermal=nominal
15:08:20  modelLoadStateChanged from=ready to=loading reason=init-preload
```

The keyboard extension is also a fresh process — `viewDidLoad ... memMB=4`, and a brand-new `KeyboardState` instance:

```
15:08:33  KeyboardViewController 731F83D0  viewDidLoad  memMB=4
15:08:33  KeyboardState 30A850AD  init
[...]

Issue #260, opened 2026-07-29
Cold-start dictation: keyboard stays in typing layout because no controller owns the session (activeID=none)


Pivii:
## Summary

On roughly 1 cold start in 10, the keyboard stays in its typing layout while the app is already recording. The Dynamic Island shows REC, the audio engine is capturing, the transcription lands correctly — only the keyboard UI is wrong. It self-corrects after ~2.5 s, when iOS recreates the controller.

The cause is not layout geometry. It is controller ownership: at the moment the mic is tapped, `KeyboardState` has no registered active controller (`activeID=none`), so every status change is dropped by the `skippedInactive` guard and `hostingSet_recording` is never called.

This is distinct from #116, which is about keys rendering shrunken or misaligned. Here the geometry is correct — the keyboard simply never leaves the typing layout.

## Evidence

Device log, build `1.8.0 (24)` at `6ce5cf2@fix/257-live-activity-state-check`, iPhone16,2 / iOS 26.5.2. Captured incidentally during the #257 device validation.

**The failing cold start:**

```
14:14:36  KeyboardViewController 6C4DA89F  deinit  memMB=20
14:14:37  keyboardMicTapped
14:14:37  A49EED3D  dictStatusChange_enter            status=requested  activeID=none
14:14:37  A49EED3D  dictStatusChange_skippedInactive  status=requested  activeID=none
14:14:37  C295CB1A  deinit  memMB=19
14:14:37  coldStartDarwinFallback elapsedMs=524 status=requested
14:14:38  refreshFromDefaults  storedStatus=recording visible=false sessionID=AD6B3F8C status=requested
14:14:38  A49EED3D  dictStatusChange_enter            status=recording  activeID=none
14:14:38  A49EED3D  dictStatusChange_skippedInactive  status=recording  activeID=none
14:14:38  A49EED3D  skippedUnregister_activeSession   status=recording  activeID=none
14:14:38  <APP> statusChanged from=idle to=recording source=coordinator
```

No `hostingSet_recording` anywhere in that window. The hosting constraint stays at 52 pt (typing) while the app records.

**Recovery, ~2.5 s later**, when iOS creates a fresh controller that does register:

```
14:14:40  E9DE444D  viewDidLoad
14:14:40  E9DE444D  viewWillAppear_entry  status=recording storedStatus=recording coldStart=true hostingConst=52.0
14:14:40  refreshFromDefaults  storedStatus=recording visible=true sessionID=AD6B3F8C status=recording
[...]

PR #273, opened 2026-07-30
Unify keyboard-area presentation behind a single mode value

Pivii:
refs #271 — `Closes` does not auto-close on a merge into `develop` in this repo, so please close the issue by hand.

## What changed

The keyboard area could show one of three things, decided by three independent flags across two layers: a SwiftUI `@State` for the emoji picker, a separate `Bool` for the same picker in the view controller (kept in step through `NotificationCenter`), and the recording overlay derived from `DictationStatus`. `KeyboardAreaMode` replaces all of it with one value in `KeyboardState`, read by both layers.

- **`DictusCore/KeyboardAreaMode.swift`** — `keys / emoji / panel / recording`, plus the pure transition rule `resolving(status:current:)` and `DictationStatus.ownsKeyboardArea`. It lives in DictusCore because the keyboard extension target has no test bundle; `DefaultKeyboardLayer` and `LayoutType` set that precedent. `ownsKeyboardArea` is an exhaustive switch on purpose, so #267's `processing` case cannot silently fall through to "no overlay".
- **`KeyboardState`** — owns `areaMode`. It is recomputed in `dictationStatus`'s `didSet`, the one funnel none of its dozen write sites can bypass. Starting a dictation is therefore an explicit transition to `.recording`, and that transition is what dismisses the picker.
- **`KeyboardRootView`** — switches on the mode. Its private `@State`, the `onEmojiDismiss` closure and the `.dictusToggleEmoji` round-trip are gone.
- **`KeyboardViewController`** — `applyLayout(for:)` is a `switch` with every case setting the hosting height, the hosting bottom anchor and key-grid visibility explicitly, no fallthrough. It replaces the negated-guard chain where restoring the 52 pt toolbar was conditional on the emoji flag being false.

Two latent bugs fall out. `reloadKeyboardLayout` hardcoded the hosting view back to 52 pt, so switching language from the emoji picker's own toolbar squashed the picker, and a rebuild during a dictation put a fresh key grid on top of the overlay; it now re-applies the mode. And a mic tap from the picker that was refused (model loading, or the 1.5 s debounce) used to close the picker while the hosting view stayed expanded; the picker now just stays open.

`heightConstraint` — the keyboard's own declared height, the #166 no-go zone — is untouched by every branch.

## The bug this surfaced

Running the keyboard in the simulator caught a regression the first version introduced: opening the emoji picker hid the key grid and expanded the hosting view to 268 pt, but SwiftUI stayed on `.keys`, so the area was blank below the toolbar. `develop`, tested the same way, renders the picker.

The log has both halves:

```
KeyboardState        presentAreaMode      from=keys to=emoji
KeyboardViewController dictStatusChange_enter  status=idle mode=emoji hasAppeared=true activeID=B84B9758
KeyboardViewController hostingSet_emojiOpen    old=52.0 new=268.0 status=idle mode=emoji
```

The controller did its half correctly, with `activeID` matching its own `controllerID`, and no `showsOverlayChanged` followed from the view.

`@Published` notifies in `willSet`. The controller reacts to a mode change by forcing `layoutIfNeeded()` — deliberately, so the hosting constraint lands before SwiftUI draws (#99, #142) — and a forced UIKit layout makes `UIHostingController` run its SwiftUI update pass inside that `willSet`, where the body still reads the old mode. It renders the old mode and marks itself clean. A dictation survives this because `requested → recording → transcribing` each notify again; a one-shot picker toggle has no follow-up event, so it stays stale forever.

`areaMode` now notifies from `didSet`, via `objectWillChange` plus its own subject, so both observers read the value that is actually stored. **This fix is not runtime-verified — see below.**

## Deliberate asymmetry: the ownership guard

`activeControllerID == controllerID` gates `.recording` only, on both layers. An overlay is pushed from another process and can arrive while no controller owns the keyboard (#260), so it is worth withholding until one does. A picker is opened by a key the user just touched on the visible keyboard; gating it would blank the area for that whole window. This is the pre-#271 split.

## Log signatures

`dictStatusChange_enter`, `dictStatusChange_skippedInactive`, `hostingSet_recording`, `hostingSet_idle` and `showsOverlayChanged` are kept verbatim, with `mode=` added — #260 and #261 are open and quote these lines from device logs.

## Verification

- All four targets build (DictusApp, DictusKeyboard, DictusCore, DictusWidgets).
- 472 DictusCore tests pass, including 11 new `KeyboardAreaModeTests`.
- SwiftLint `--strict` clean. The baseline was regenerated in its own commit: same 45 entries, only KeyboardViewController's two length violations move, both shrinking (517→505, 103→100).
- Simulator: the mode transition, the ownership guard and the hosting geometry were read out of the device log on a real keyboard session.

**Not verified, and the first thing to check on device:** the emoji picker actually rendering after the `didSet` fix. The regression was found and root-caused by running the keyboard in the simulator, but the fix itself has not been seen working — re-running it needs the simulator GUI, which is off limits on the build machine (it steals the maintainer's focus), and driving the system globe key with synthetic events had already become unreliable. The two device-only acceptance criteria are untested for the same reason.

Concretely, on device, in order:

1. Open a text field on the Dictus keyboard and tap the emoji key. The picker must render — not a blank grey area under the toolbar. **This is the regression; if it is blank, the `didSet` fix is wrong.**
2. Dismiss the picker: key grid back, toolbar back to 52 pt (AC 4).
3. Open the picker and switch language from its toolbar. The picker must stay full-size — this was broken before this PR and should now work.
4. AC 7: force-quit the app, tap the keyboard mic, confirm the overlay fills the area and the toolbar returns to 52 pt afterwards.
5. AC 8: open the picker, start a dictation from it, let it complete, confirm the layout returns to normal.

## Summary by CodeRabbit

* **New Features**
  * Added coordinated, type-safe keyboard-area modes covering keys, emoji picker, panel placeholder, and recording overlay.
  * Keyboard presentation now transitions automatically based on dictation state, with consistent behavior when dictation starts/ends.
  * Emoji picker toggling is now unified through shared keyboard-area presentation.

* **Bug Fixes**
  * Prevented conflicting emoji and recording overlays, including safer restoration of the key grid after dictation.
  * Clamped emoji picker height to avoid negative sizing during initial layout.
  * Improved synchronization of UI updates to ensure the correct keyboard area is shown immediately and reliably.


Pivii:
Thanks — five remarks, two applied and three declined. Reasons below so the thread stays traceable.

## Applied

**Clamp `availableHeight` (`KeyboardRootView`, `.emoji`)** — `max(0, geo.size.height - toolbarHeight)`, in `63f3ee0`. Valid: the body can be evaluated on a frame where the hosting constraint has not landed, and handing a negative height to the picker is the same family as the blank-area bug this branch already had to fix.

**Main-thread guard in `applyAreaMode`** — applied in `2b946ec`, but as `assert(Thread.isMainThread, …)` rather than the proposed `dispatchPrecondition(condition: .onQueue(.main))`. `dispatchPrecondition` also traps in release builds, and crashing a shipped keyboard extension is worse than the silent UIKit corruption it would catch. `assert` is compiled out of release, so it is the same net during development with no production risk. Verified the Release configuration still builds.

## Declined

**Collapse `.emoji` / `.panel` / `.recording` into one branch.** Declined. The acceptance criterion in #271 is explicitly "a `switch` over the mode with every case handled explicitly, not a chain of negated guards", and the issue body spells out that "every case must set the hosting height, the bottom anchor, and key-grid visibility explicitly, with no fallthrough". Folding the three re-creates the shared-path shape the refactor exists to remove, and #241's panel is likely to want different geometry from the recording overlay — at which point the fold has to be undone. Noted as "low value" in the review itself, and the duplication is four lines per case.

**Route `reloadKeyboardLayout` through `applyAreaMode`.** Declined. The direct `applyLayout(for:)` call is deliberate and fixes a real bug: the old code hardcoded the hosting view back to 52 pt, which squashed the emoji picker when the language was switched from the picker's own toolbar, and put a fresh key grid on top of an active overlay. Adding the guards back would let a rebuild leave the area collapsed — a rebuild always produces a visible key grid pinned to the idle anchor, so the geometry has to be re-derived unconditionally or not at all. The current version is the one validated on device; the suggestion is not.

**Rename the `showsOverlayChanged` probe.** Deferred, not declined on merit — the name is stale now that the handler is mode-driven. #260 and #261 are open and quote that key verbatim from device logs, along with `dictStatusChange_enter`, `dictStatusChange_skippedInactive`, `hostingSet_recording` and `hostingSet_idle`. Renaming a load-bearing log key while those investigations are live would break the greps. `mode=` was added to the details instead. Worth revisiting once both issues close.

Issue #252, opened 2026-07-28
Parakeet downloads on the dictation hot path when models are absent — same asymmetry #249 removed for Whisper


Pivii:
> *This was generated by AI during triage.*

## Context

While fixing #249 (PR #251), the Parakeet load path was audited and found **not** to have the bug that made every Whisper model unusable offline. Parakeet resolves its FluidAudio cache locally when the models are present, which is exactly why it kept working in airplane mode while Whisper failed.

One asymmetry remains, deliberately left out of scope of that PR.

## The asymmetry

The Parakeet path goes through a single call whose name states what it does: `AsrModels.downloadAndLoad(version:)`. When the models are already cached it loads locally and never touches the network. When they are **absent**, it downloads them — **on the dictation hot path**, at the moment the user taps the mic.

That is precisely the behaviour #249 removed for Whisper. After PR #251, the two engines behave differently in the same situation:

- **Whisper** — model missing: fails fast with a clear localised message pointing at the model manager. Downloading is the model manager's job.
- **Parakeet** — models missing: silently starts a download while the user waits, with no progress indication in the keyboard and no way to cancel.

## Why this is low priority

This is **not a regression** and there is no known user report. Reaching it requires the Parakeet cache to be absent while Parakeet is the selected model — onboarding downloads a model before dictation is reachable, so the realistic routes are a partial or interrupted download, a restore that did not bring the cache along, or manual cleanup. Rare.

It is filed because the inconsistency is now *documented policy* rather than accident: after #251, "downloading never happens on the dictation hot path" is the rule for Whisper, and Parakeet is the remaining exception.

## Blocked until #251 merges

This touches the same engine-preparation code that PR #251 restructured, including moving those methods into an extension. **Do not start this before #251 is merged to develop**, or the two will conflict.

---

## Agent Brief

**Category:** bug
**Summary:** Align the Parakeet engine-preparation path with the Whisper one — never download at dictation time.

**Current behavior:**
Parakeet preparation calls a combined download-and-load routine. With models cached it resolves locally and works offline. With models absent it initiates a download during dictation, while the user waits, with no progress feedback in the keyboard and no cancellation.

**Desired behavior:**
Parakeet preparation loads strictly from the local cache. When the cache is absent or incomplete, it fails fast with the same clear, localised, actionable message the Whisper path now produces, directing the user to the model manager. Downloading remains exclusively the model manager's responsibility, where progress and cancellation already exist.

**Key interfaces:**
[...]

PR #276, opened 2026-07-30
fix(models): load Parakeet from the local cache on the dictation path (#252)

Pivii:
refs #252. (`Closes` does not auto-close on a merge into `develop` in this repo — please close manually.)

## The asymmetry

`ParakeetEngine.prepare` called `AsrModels.downloadAndLoad(version: .v3)`. With the FluidAudio cache present it resolves locally, which is why Parakeet kept working in airplane mode while Whisper failed (#249). With the cache **absent** it downloads from HuggingFace — on the dictation hot path, at the moment the user taps the mic, with no progress in the keyboard and no way to cancel.

After PR #251, "downloading never happens on the dictation path" is policy for Whisper, and Parakeet was the remaining exception.

## Investigation: does FluidAudio expose a load-only entry point?

This is the question the brief asked to settle. **No, it does not — despite appearances.**

`AsrModels.loadFromCache(configuration:version:progressHandler:)` reads like a load-only API. It is not. It calls `AsrModels.load(from: defaultCacheDirectory(for:))`, and `load` goes through `DownloadUtils.loadModels`, which:

1. `loadModelsOnce` checks whether every required `.mlmodelc` exists at the repo path, and calls `downloadRepo` — a network download — if any is missing;
2. `loadModels` wraps that in a catch-all: on **any** first failure it deletes the whole repo directory and runs `loadModelsOnce` again, i.e. wipe and re-download.

So no FluidAudio entry point is download-free, and the guard has to sit in front of it. That is what this PR adds. `AsrModels.load(from:)` is used rather than `downloadAndLoad` because it skips the `download(...)` step entirely; the guard in front of it is what makes the remaining path local.

### How "cached and complete" is determined, and why that check is trustworthy

`AsrModels.modelsExist(at:version:)` is public, but it only checks that the four `.mlmodelc` **directories** and the vocabulary file exist as paths. An interrupted download leaves a directory behind, so that check would pass on a partial cache, FluidAudio would then fail during compilation, and the catch-all above would wipe and re-download. So it is too weak, for exactly the reason #249 required a compiled bundle rather than merely a directory.

`ParakeetModelRepository` therefore requires, at `AsrModels.defaultCacheDirectory(for: .v3)`:

- `parakeet_vocab.json` present (the engine cannot decode a single token without it), and
- each of `Preprocessor.mlmodelc`, `Encoder.mlmodelc`, `Decoder.mlmodelc`, `JointDecision.mlmodelc` a **directory containing `coremldata.bin`**.

`coremldata.bin` is trustworthy because it is exactly what FluidAudio itself verifies immediately before calling `MLModel(contentsOf:)` (`DownloadUtils.loadModelsOnce`), and because `ModelRepoDownloader` moves each file into place atomically, so a file that is present is complete. The check is strictly stronger than both `AsrModels.modelsExist` and FluidAudio's own pre-load existence test. A partial cache reads as absent.

The file names are not restated anywhere: `ParakeetModelRepository` takes them as parameters and the call site passes FluidAudio's `ModelNames.ASR.requiredModels` / `.vocabularyFile`, so a rename in a dependency bump is a compile-time change rather than a silent one.

## What changed

**`DictusCore/Sources/DictusCore/ParakeetModelRepository.swift` (new)** — the completeness rule described above, pure Foundation. It lives in DictusCore because that is the only place with a test target; FluidAudio is a DictusApp-only dependency, so DictusCore cannot import it, hence the file names as parameters.

**`DictusApp/Audio/ParakeetEngine.swift`** — `installedModelCacheDirectory(version:)` binds FluidAudio's cache location and file names to that rule, in one place. `prepare` guards on it and loads with `AsrModels.load(from:version:)`; an absent or partial cache throws `SpeechModelError.modelNotInstalled`, the shared error #249 introduced, so both engines produce identical localised wording pointing at the model manager. No new user-facing string.

**`DictusApp/DictationCoordinator.swift`** — `ensureParakeetReady` fails fast on the same check *before* any init task is created, mirroring `ensureWhisperKitEngineReady`, and logs `localModelResolved` / `localModelMissing` under component `ParakeetLoad` so the local path is observable in an exported log.

**`DictusApp/Models/ModelManager.swift`, `ModelRepoDownloader.swift`** — comment fixes only. Two comments still described `prepare` as calling `downloadAndLoad`.

The model manager is otherwise untouched and stays the only place a Parakeet download starts: `downloadParakeetModel` fetches the repo through `ModelRepoDownloader` (byte-accurate progress, stall handling, resume) and only then calls `prepare` for the Core ML compilation step. Its `requiredPaths` tripwire already enforces the same file set this guard checks.

## Acceptance criteria

| --- | --- | --- |
| Cached models + airplane mode: cold-start dictation still works | **Not verified — device only** | The path is now guard + `AsrModels.load(from:)`. With a complete cache `loadModelsOnce` logs "Found … locally, no download needed" and never calls `downloadRepo`. Source-level reading, not a runtime observation. |
| Cache removed: dictation fails immediately with the same localised message, no download | Met in code, **device step listed** | `ParakeetEngine.installedModelCacheDirectory()` returns `nil`, the coordinator throws `SpeechModelError.modelNotInstalled` before creating the init task. Same enum case, same FR/EN string as Whisper. 10 unit tests cover absent / partial / missing-vocabulary / not-a-directory. |
| No download is ever initiated from the dictation path, online or offline | Met for our code; **one residual inside FluidAudio, see below** | `downloadAndLoad` no longer exists in the codebase (only in comments). `AsrManager.initialize(models:)` is pure assignment, no network. |
| Downloading Parakeet from the model manager still works, with progress | Unchanged code; **not run** | `ModelRepoDownloader` and `downloadParakeetModel` are untouched apart from a comment. `prepare` runs after the files land, so it finds a complete cache. Needs a real download to confirm. |
| Whisper behaviour is untouched | Met | No Whisper path appears in the diff. `swift test` in DictusCore: 482 tests, 0 failures. |
| Investigate FluidAudio's load-only entry point | Met | Section above. |

### The residual, stated plainly

With the cache verified complete, one download route remains **inside FluidAudio**: if `MLModel(contentsOf:)` fails on a bundle that passed the guard — a genuine Core ML/ANE failure — `DownloadUtils.loadModels`'s catch-all deletes the repo directory and re-downloads it. It is no longer reachable through a missing or partial cache, which is the case the issue describes, and it is unchanged from today's behaviour.

Closing it completely would mean building `AsrModels` ourselves from four `MLModel(contentsOf:)` calls plus vocabulary parsing, which duplicates FluidAudio's per-model compute-unit policy (preprocessor `.cpuOnly`, the rest `.cpuAndNeuralEngine` — the policy FluidAudio documents as what keeps background execution permitted on iOS). Silent drift in that policy on a `upToNextMinor` bump would be a worse failure than the one it removes, on a priority:low issue. I chose not to, and am flagging it rather than hiding it.

## Behaviour change worth reviewing

A Parakeet download that completes but leaves an incomplete cache now fails loudly at the compile step instead of being silently patched by a second download inside `downloadAndLoad`. That is intended, and the downloader's own `requiredPaths` verification already fails first on the same file set, so this should be unreachable in practice.

## How I verified

- `swift test` in DictusCore: **482 tests, 0 failures**, including 10 new `ParakeetModelRepositoryTests`.
- `xcodebuild build` for **DictusApp, DictusKeyboard and DictusCore**, `generic/platform=iOS Simulator`: BUILD SUCCEEDED for all three.
- `swiftlint lint --strict --no-cache`: **exit 0** with the regenerated baseline. Two refreshed entries, count unchanged at 45 (the `file_length` warning on DictationCoordinator.swift embeds its line count in the match key, 670 to 686; a baselined `force_unwrapping` in ModelManager.swift shifted one line from a comment edit). No rule disabled.
- `Localizable.xcstrings` is untouched — the error case and its FR/EN strings already exist from #249.

**I did not run this on a device and did not observe network traffic.** No simulator check can exercise this meaningfully: it needs a real Parakeet cache, a real mic, and airplane mode.

## Suggested device test (mirrors the #249 validation)

1. Online: select Parakeet (`parakeet-tdt-0.6b-v3`), confirm it transcribes.
2. Airplane mode on, force-quit the app, dictate from the keyboard. Expect success and `ParakeetLoad … localModelResolved folder=parakeet-tdt-0.6b-v3-coreml download=false` in the exported log. **This is the no-regression check — it works today and must keep working.**
3. Still offline, delete the Parakeet model in the model manager, then dictate. Expect the French message "Ce modèle n'est pas installé…" immediately, `localModelMissing` in the log, and no download attempt or spinner.
4. Back online, re-download Parakeet from the model manager. Expect the usual byte-accurate progress bar and a working transcription afterwards.
5. Switch to a Whisper model and dictate offline once, to confirm #249 did not regress.

## Out of scope

- The model manager's download UI, progress reporting and cancellation.
- Whisper's path (#249, PR #251).
- Any Parakeet download affordance in the keyboard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parakeet speech models now load only from a complete local cache.
  * Prevented unexpected, uncancelable downloads during speech model initialization.
  * Missing or incomplete model files now fail immediately with a clear error.
  * Model setup now verifies required vocabulary and compiled model files before loading.

* **Tests**
  * Added coverage for validating complete, incomplete, malformed, and missing local model caches.


Pivii:
Applied the CodeRabbit finding in `efb82e4`: the vocabulary file and `coremldata.bin` now have to be regular files, since `fileExists(atPath:)` also answers yes for a directory. Two tests added for the directory cases (484 tests, 0 failures).

The PR body's completeness section stands, with that refinement: a required path that is a directory now reads as absent, same as a missing one.

Issue #272, opened 2026-07-30
Decouple keyboard layout from dictionary language: four axes, currently two


Pivii:
> *This was generated by AI during triage.*

Surfaced during the #241 design session (2026-07-30). Filed because no existing issue covers it: #52 is the Settings picker UX, #151 is the QWERTZ layout itself, #157 is per-language long-press accents. None of them addresses the coupling described here.

## The four axes

Dictus has four independent language-shaped concepts. Users experience them as four separate choices:

| Axis | What it controls | Where it lives today |
|---|---|---|
| **Keyboard layout** | key positions — AZERTY, QWERTY, QWERTZ | `LayoutType` |
| **Dictionary language** | autocorrect, suggestions, long-press accents, spacebar and return labels | `SupportedLanguage` |
| **Transcription language** | what the speech model transcribes | decoupled in #226, Settings-level |
| **Translation target** | the output language of a translate Smart Mode | #79 |

The third was decoupled by #226. The fourth arrives with #79. **The first two are still welded together.**

## The coupling

Selecting a language writes both the language key and that language's default layout in one operation. `SupportedLanguage` exposes a `defaultLayout`, and the switcher applies it unconditionally.

There is no way to choose them independently. Choosing English forces QWERTY.

## Why it matters

The concrete case, from the maintainer: writing English on an AZERTY keyboard. A French typist writing an English message wants English autocorrect and suggestions, and wants to keep the key positions their fingers know. Today that combination is unreachable — picking English relayouts the keyboard under their hands.

This is not exotic. It is the normal condition of anyone bilingual who learned to type on one physical layout, which is most bilingual people.

The reverse case exists too: a German user on QWERTZ (#151) who occasionally writes English should keep QWERTZ.

## What makes this bigger than a settings toggle

- **`defaultLayout` becomes a default, not a rule.** It should seed the layout on first selection of a language, then stop overriding an explicit user choice. That means storing whether the current layout is inherited or chosen.
- **Long-press accents follow the dictionary language, not the layout** — which is the direction #157 is already going. The two issues should agree on which axis owns accents before either ships.
- **The onboarding language step** currently sets one thing that means two. It has to keep being one question for a first-time user while no longer being one value underneath.
- **Migration.** Existing users have a language and an implied layout. Nobody's keyboard may change shape on update.
[...]

Issue #241, opened 2026-07-28
Keyboard toolbar switcher rework: layout-only concept, designed with Smart Modes


Pivii:
## Placeholder resolved — decisions from the #79 design session (2026-07-30)

Captured during #226 triage as a placeholder: once the transcription language was decoupled, the toolbar's left slot should stop reading as a "language" switcher and be redesigned together with Smart Modes. That session happened. What follows is decided; the remaining open point is called out at the end.

## The constraint nobody had written down

Three competitors were surveyed (Wispr Flow, Superwhisper, Typeless — screenshots in `designs/competitors/`, gitignored). **None of them hosts an autocorrect suggestion bar.** Their top strip is a pure dictation cockpit: Wispr has hamburger + style pill + mic, Superwhisper has a mode stepper + record button, Typeless has no keys at all in its default layout.

Dictus has to fit a dictation cockpit *and* a three-slot suggestion bar into the same 52 pt. `ToolbarView.swift:37-54` already resolves that conflict by evicting the language switcher whenever suggestions appear. Any redesign inherits that arbitration; we cannot copy a competitor's cockpit.

## Decided layout

```
Idle:    [☰]  ← centre slot →  [🎤]
Typing:      [ sugg ][ sugg ][ sugg ]  [🎤]
```

- **Left slot: a hamburger button.** It costs the same 32 pt as today's language label, so it is free. Like the language label, it yields to the suggestion bar while typing.
- **Right slot: the mic pill, unchanged.** Its colour now also signals the armed Smart Mode (`#3D7EFF` normal, `#8B5CF6` armed) — see #79.
- **Centre slot: a priority-ordered state machine**, specified in #79 (error > undo > suggestions > armed mode name > discovery hint).
- **No mode pill in the toolbar.** 70–90 pt would push the suggestion bar below three legible slots.

## The language switcher leaves the toolbar

Since #226, transcription language is a Settings-level concept and the toolbar switcher only drives the keyboard language — autocorrect dictionaries and key labels — plus the layout it writes alongside (`LanguageSwitcherView.swift:47-49` writes both `SharedKeys.language` and `SharedKeys.keyboardLayout`).

That is a keyboard concern, not a dictation concern, and it does not deserve permanent residence in the most contested 32 pt of the UI. It moves into the hamburger panel.

## The hamburger panel

[...]

Issue #166, opened 2026-05-09
Keyboard overlap after switching keyboards with globe key


Pivii:
## Summary

A beta user reports that the Dictus iOS keyboard can sometimes render with a visible overlap / leaked host keyboard background when the user switches between keyboards using the globe key.

Pierre was able to reproduce the issue by repeatedly tapping the globe key to rotate through installed keyboards. The issue appears when opening/switching back to Dictus after the keyboard height changes between different keyboard implementations.

Important datapoint: Pierre also reproduced the same kind of issue with the Wispr Flow keyboard. This suggests the bug may be partly caused by iOS keyboard-extension lifecycle / inputView height limitations, but we should still investigate whether Dictus can make its implementation more robust.

## Reproduction

1. Install/enable Dictus keyboard alongside other keyboards.
2. Open any text field in a host app.
3. Tap the globe key several times to rotate through the available keyboards.
4. Switch back to Dictus.
5. Observe that, intermittently, the Dictus keyboard area can show overlap / leaked background from another keyboard or host UI.

## Observed behavior

- The bug happens when opening/switching to the keyboard, not only after tapping the mic or after dictation.
- It seems related to changing keyboard heights while rotating between keyboards.
- The Dictus keyboard can appear over another keyboard/background layer instead of fully owning/painting the keyboard area.
- Pierre can reproduce locally.
- Similar behavior was reproduced with Wispr Flow, so this may not be Dictus-specific.

## Expected behavior

When Dictus becomes the active keyboard, its input view should occupy and repaint the keyboard area cleanly, without visual leftovers or overlap from the previous keyboard/host UI.

## Investigation notes

Potential areas to inspect:

- `KeyboardViewController` / `inputView` sizing and lifecycle during keyboard switches.
- Whether Dictus sets an opaque enough background across the full keyboard bounds during transitions.
- Whether SwiftUI root view / hosting controller resizes correctly when iOS changes keyboard height.
- Whether the issue is triggered by animated height changes or stale snapshots from the previous keyboard.
- Whether a defensive full-size background layer or explicit invalidation/layout pass on `viewWillAppear`, `viewDidLayoutSubviews`, or trait/size changes reduces the issue.

Because Wispr Flow can show a similar issue, document what is actually controllable by the extension vs what may be an iOS limitation.

## Attachments / evidence

Pierre will add screenshots and screen recordings in comments.

Pivii:
[...]

Issue #236, opened 2026-07-27
Hide paywall entry points behind a visibility flag until Pro features and ASC setup exist


Pivii:
> *This was generated by AI during triage.*

## Problem

develop now ships the full paywall UI (#78/PR #218) and StoreKit gating (#55/PR #214), but NO premium feature exists yet (Smart Modes #79 not built) and the App Store Connect setup is not done (#215: Paid Apps agreement + product creation pending). Consequences of shipping a TestFlight build as-is:

- StoreKit cannot load any product (nothing exists in ASC), so the paywall renders with empty prices / a loading error — looks broken, even in sandbox.
- Advertising a 39,99 €/year subscription for features that do not exist yet pollutes tester feedback and hurts first impressions.

This currently BLOCKS TestFlight releases even though develop has accumulated significant shippable value (#227-#231). Decision (maintainer, 2026-07-27): decouple with a visibility flag so TestFlight builds can ship now.

## What to do

Introduce a single compile-time visibility flag, e.g. `PremiumFlags.paywallVisible` (a `static let`, default `false`), placed where all three entry points can reach it (DictusApp — or DictusCore if cleaner). Gate the paywall's entry points behind it:

1. `DictusApp/Views/HomeView.swift:44` — the `ProBannerView()` on the home screen.
2. `DictusApp/Views/SettingsView.swift:78` — the "Dictus Pro" section (crown icon, always-first section) in Settings.
3. `DictusApp/Views/SettingsView.swift:196` — the locked-feature rows (lock + PRO pill) that push `PaywallView`.

When the flag is `false`: no Pro banner, no Dictus Pro section, no locked-feature rows — the app looks like there is no subscription at all. When `true`: exactly today's behavior.

## What NOT to do

- Do NOT delete or comment out any paywall/subscription code — `PaywallView`, `ProBannerView`, `SubscriptionManager`, gating logic all stay compiled and intact.
- Do NOT touch `SubscriptionManager` logic or entitlement checks — `proStatus.isProActive` handling stays as is (a user can simply never reach the paywall while hidden).
- Do NOT add remote config / server-driven flags — a compile-time constant is enough (YAGNI).

## Acceptance criteria

- [ ] With the flag `false` (default): Home shows no Pro banner; Settings shows no Dictus Pro section and no locked-feature rows; no navigation path reaches `PaywallView`.
- [ ] Flipping the single flag to `true` restores today's behavior exactly (verify by toggling locally).
[...]

PR #277, opened 2026-07-30
feat(keyboard): hamburger panel with keyboard-language selection (refs #241)

Pivii:
Implements the hamburger panel decided in #241 (refs #241 — `Closes` does not auto-close on a merge into `develop` here, so this needs closing by hand).

## What changed

The toolbar's left slot no longer cycles the language. It holds a bare hamburger at the same 32 pt, and yields to the suggestion bar exactly where the language label did.

Tapping it puts the keyboard area into the `.panel` mode #271 reserved. The bar is **not** kept above the panel — it becomes the panel's header, so there is only ever one bar:

```
closed:  [☰]  ← centre slot →  [🎤 pill]
open:    [✕]                   [Dictus Pro] [⚙]
```

- `☰` becomes `✕` in place; the mic pill is replaced by the gear. There is no mic while the panel is open.
- The gear is last in the stack in both subscription states, so it never moves. The Pro entry inserts to its left for non-subscribers only.
- The bar keeps its 52 pt in both presentations, so opening the panel moves no geometry of its own.

The panel body lists the supported languages, checkmark on the active one, that language's layout as a display-only trailing label (#272 owns making it editable). Selecting one applies it and **leaves the panel open** so the checkmark visibly moves.

Two supporting changes:

- `SupportedLanguage.activate(_:)` in DictusCore now owns the two writes that were inline in the toolbar switcher (language + that language's `defaultLayout`). One place for the coupling #272 will dismantle, and testable — the keyboard target has no test bundle. `next()` went with the cycle it existed for.
- `KeyboardState.openDictusApp(intent:)` opens DictusApp through `extensionContext` (the existing `openDictusURL` path), for the gear and the Pro entry.

`LanguageSwitcherView.swift` is deleted; `KeyboardPanelView.swift` replaces it in the target.

## Notes for review

- **No animation on the mode change, by design.** A transition on the `switch` branch keeps the outgoing and incoming branches alive together, and they are the single child of a VStack — SwiftUI stacks them, putting the closed bar above the open one for the duration. That transient is the exact artefact the design was reshaped to avoid. The bar swaps instantly and the fade lives inside the panel body, where it stacks nothing and touches no constraint. Reduced motion removes it.
- Nothing in this PR touches `KeyboardAreaMode`, the presentation machinery from #271, or the keyboard's declared height constraint (#166).
- The key grid staying hidden after a language rebuild is inherited, not added: `reloadKeyboardLayout()` already ends with `applyLayout(for: areaMode)`.

## Verification

- Builds: `DictusApp`, `DictusKeyboard`, `DictusCore` schemes, iOS Simulator, all green.
- Tests: `swift test` in `DictusCore` — 477 passed, 0 failures, including 5 new for `activate(_:)`.
- SwiftLint `--strict`: clean, baseline unchanged.
- Not verified: everything visual and everything that needs a real keyboard extension. See the device checklist in the issue thread / the agent recap.

## Summary by CodeRabbit

* **New Features**
  * Added a keyboard panel for browsing and selecting supported languages.
  * Displays the active language and its default keyboard layout.
  * Added toolbar controls for opening/closing the panel, accessing Settings, and entering Pro.
  * Language selections are saved and applied automatically.
  * Added responsive panel sizing and reduced-motion support.
* **Bug Fixes**
  * Keyboard mode now reliably returns to the keys layout after reopening or leaving the language panel.
  * Selecting the currently active language no longer triggers unnecessary changes.
* **Style**
  * Updated toolbar and recording controls with consistent sizing, spacing, colors, and haptic feedback.


Pivii:
## Device validation — passed

Physical device, iOS 26.5.2, builds `bb071b0` then `227b49a`. Three rounds of feedback, each measured against a device log rather than an impression.

### Round 1 — touch targets were blocking

Every control in the panel bar was under Apple's 44 pt minimum: the toggles at 32, the gear at 36, the Pro pill at 28 tall. The language rows, at 44, were the only conforming targets — and the log showed exactly that asymmetry.

| | before (`ee165e3`) | after (`227b49a`) |
| --- | --- | --- |
| Open → close a panel | **10 to 22 s**, on all six cycles | **1 s** |
| Languages selected by mistake | 10 in 11 seconds | **0** |

Taps aimed at the close control were landing on the rows below it. Visual sizes are unchanged; only the touch targets grew.

### Round 2 — the panel outlived every exit

`areaMode` lives on the `KeyboardState` singleton, so it survived controller recreation: the globe, and returning from DictusApp, both put the user back in the menu. One rule now covers it — the panel closes on language selection, before either bar entry leaves for the app, and on any reappearance. The emoji picker keeps the old restore behaviour deliberately.

### Round 3 — visual alignment

The toggle took the mic's 56x36 pill footprint, and the overlay's top bar moved from 48 pt to 52 pt so both bars place a pill at y=10. Starting a recording was visibly jumping the pills up 4 pt. Horizontal alignment already worked: the overlay's 17 pt trailing inset offsets the mic's 66 pt ring around its 56 pt pill.

Bar icons also moved to a shared `dictusPillIconSecondary` token in DictusCore, so the toolbar and the recording overlay stop drifting apart.

### Pro pill removed from the build

The paywall has been hidden since #236, but this PR's Pro pill was not gated — it led to a paywall the user cannot reach. It could not have been gated as written: `PremiumFlags` was an `internal` constant in DictusApp, invisible to the extension. Moved to DictusCore unchanged and still `false`, so one flag now gates entry points in both targets. #279 tracks re-enabling everything.

## CodeRabbit review — all three findings declined, with reasons

**Reordering `presentAreaMode(.keys)` before `onLanguageChanged` — declined, it would introduce a regression.** `reloadKeyboardLayout` destroys and recreates the grid rather than re-theming it. In the current order the new grid is built while the mode is still `.panel`, so `applyLayout` leaves it hidden, and the switch to `.keys` reveals it once, already in the new language. The device log confirms it:

```
08:06:13 languageSelected language=en
08:06:13 reloadLayout_begin
08:06:13 hostingSet_panelOpen  mode=panel   <- new grid built hidden
08:06:13 presentAreaMode from=panel to=keys
08:06:13 hostingSet_idle                     <- revealed once, correct language
```

Reversed, `applyLayout(.keys)` would unhide the **old** grid before the rebuild replaced it — a visible flash of the previous layout, which is the artefact class of #92 and #142.

**Consolidating the two `ToolbarView` construction sites — declined for now.** A fair maintainability point, but it restructures a call site this PR only adds a parameter to, immediately before a merge, with no functional gain.

**`instanceID: ""` on the `languageSelected` probe — declined, noted.** The convention is real, but this event is already correlatable: it sits between `presentAreaMode` and `reloadLayout_begin` in the same second, both of which carry a controller ID. Threading an identifier into the panel view is plumbing that buys nothing the timestamp does not already give.

## Not validated

The Pro pill's rendering is now unreachable in this build, so its light-keyboard legibility — flagged during implementation as possibly reading like a key — remains unverified. #279 carries it as a step for whoever flips the flag.

Issue #255, opened 2026-07-29
Debug log spends its 200 KB budget on duplicate lines, retains only ~4 minutes


Pivii:
## Problem

The debug log is capped at 200 KB (`PersistentLog.maxFileSize = 200_000`) and rotates by keeping the last bytes. In real use that window holds about **4 minutes**.

Measured on a device export taken during the #249 device validation (`dictus-logs 18.txt`, 200 016 bytes, exactly the cap):

- Time span retained: `07:46:52Z` to `07:50:54Z`. **4 minutes, 5 dictations.**
- The maintainer had just run a full multi-scenario validation session (airplane mode, model switching, delete and reinstall). **None of it survived in the export.** The evidence for the acceptance criteria was gone before the file was shared.

The cap is not the real problem. The volume is. The log spends its budget on repetition rather than information, so it cannot be used for the one thing it exists for: validating a fix after a device session.

## Where the bytes go

| Pattern | Lines | Bytes | Share of file |
|---|---|---|---|
| `KeyboardRootView action=activeControllerChanged` | 145 | 27 358 | 13% |
| `statusChanged source=micButton` | 191 | 20 324 | 10% |
| `overlayShown` / `overlayHidden` | 228 | 17 611 | 8% |
| `waveformHeartbeat` | 97 | 10 739 | 5% |
| `dictStatusChange_enter` | 40 | 9 631 | 4% |
| `computeKbHeight` | 40 | 8 080 | 4% |

By category:

| Category | Lines | Bytes |
|---|---|---|
| `[keyboard]` | 982 | 154 811 (**77%**) |
| `[dictation]` | 234 | 25 144 |
| `[lifecycle]` | 86 | 8 300 |
| `[audio]` | 53 | 5 473 |
| `[model]` | 37 | 4 413 |
| `[transcription]` | 15 | 1 746 |

The signal is in the last three rows. They account for under 8% of the file.

## Root cause

Most of these lines are emitted from SwiftUI `onChange` handlers on views that exist in **many simultaneous instances**, with no check for which instance actually owns the keyboard.

`KeyboardRootView.swift:194` logs `activeControllerChanged` from every live root view. The evidence that non-owners are logging is in the payload itself:

```
component=KeyboardRootView instanceID=92762F25 action=activeControllerChanged details=newOwner=FC5F3E99 controllerID=E76598F3
[...]

Issue #128, opened 2026-04-21
KeyboardViewController instances leak across app-switches (19+ live, no deinit)


Pivii:
## Summary

`KeyboardViewController` instances are never deallocated when the keyboard detaches. Each app-switch creates a new instance while older ones remain alive and keep responding to `KeyboardState.shared.$dictationStatus` via their Combine subscription.

Observed on device 2026-04-21 session 12:08 (iPhone 17 Pro Max, iOS 26.3.1, Dictus 1.6.0 build 11) with `fix/116-keyboard-layout-glitch` diagnostic probes.

**This is the root cause of issue #116 (keyboard layout glitch).**

## Evidence

Log extract `dictus-logs 135.txt` (device session, ~17 minutes of app-switching):

- Only **4** occurrences of `action=viewDidLoad` (= 4 new controllers created).
- **19 distinct `instanceID` values** appear in `action=dictStatusChange_enter` logs — all receive every status change in parallel.
- **0** occurrences of `deinit` log.

Sample — every `statusChanged` event fires 14-19 `dictStatusChange_enter` probes in sequence (log lines 806-867):
```
instanceID=25C18895 action=dictStatusChange_enter status=transcribing
instanceID=77EFD1EB action=dictStatusChange_enter status=transcribing
instanceID=6DCC5C19 action=dictStatusChange_enter status=transcribing
... (14 more) ...
instanceID=A82BFFC7 action=dictStatusChange_enter status=transcribing
```

Each one then calls `computeKbHeight` and `hostingSet_recording`, mutating its own `hostingHeightConstraint`. The views are detached from the window but still allocated, subscribed, and responding.

## Hypothesised cause

`DictusKeyboard/KeyboardViewController.swift` L360-372 (`observeRecordingState`):

```swift
private func observeRecordingState() {
    dictationStatusCancellable = KeyboardState.shared.$dictationStatus
        .sink { [weak self] status in
            self?.handleDictationStatusChange(status)
        }
}
```

`[weak self]` is correct for the closure capture, but:

1. `KeyboardState.shared` is a **process-level singleton** (`DictusKeyboard/KeyboardState.swift` L16). The keyboard extension process persists across app-switches until iOS kills it for memory.
2. iOS does NOT release the previous `KeyboardViewController` when the keyboard detaches — it keeps the view hierarchy in memory for fast reattach.
3. Nothing cancels `dictationStatusCancellable` on `viewDidDisappear`.
[...]

PR #278, opened 2026-07-30
fix(log): say each thing once, at a 1MB cap (refs #255)

Pivii:
refs #255 — needs manual closing, `Closes` does not fire on a merge into `develop`.

## What this changes

The debug log retained ~4 minutes, reported each state transition ~9 times, and
rewrote the whole file on nearly every write once saturated. Five changes.

**A. One line per state change, at the source of truth.** iOS keeps ~9
`KeyboardViewController` instances alive at once — the accepted post-#128 norm —
and the ownership gate from #128 governs behaviour, not diagnostics. So one
transition was written ~9 times. That is a fidelity defect before it is a volume
one: a reader taking the file at face value sees 44 `idle → requested`
transitions where 4 happened, and reasonably infers a controller-lifecycle bug
that does not exist. Proving it was not one already cost a whole section of the
issue's scope note.

- `statusChanged` and `overlayShown`/`overlayHidden` now come from
  `KeyboardState.dictationStatus`'s `didSet`, the funnel no write site can
  bypass. The emission inside `refreshFromDefaults` is removed as redundant.
- `AnimatedMicButton` loses its log line only. The animation reset stays ungated
  so the app's own visible recording UI is unchanged.
- `KeyboardRootView`'s `showsOverlayChanged` and `activeControllerChanged` probes
  now fire only from the owning view.

Replaying the previous version of this branch against the device export the
issue measured turned up a gap worth calling out: `idle → requested` was
recorded **only** by the per-instance mic-button line, because `markRequested()`
writes `.requested` directly and never goes through `refreshFromDefaults`.
Simply deleting that line would have removed the transition from the log. Moving
the emission into the `didSet` is what closes it.

**B. The trim no longer rewrites the file on every write.** `coordinatedTrim`
kept exactly `maxFileSize` trailing bytes then dropped the partial first line,
leaving the file ~150 bytes *under* the cap, so the next single line re-exceeded
it. Steady state was a full read plus a full atomic rewrite per logged line,
inside a cross-process `.forReplacing` block contended between the app and the
extension. It now cuts back to a 70% low-water mark. Trailing-bytes and
line-boundary semantics are unchanged.

**C. Cap 200 000 → 1 000 000.** Derivation in the issue, confirmed by
measurement below.

**D. Consecutive identical lines collapse.** A run of N identical payloads writes
the first occurrence immediately plus one closing line carrying `repeated=N-1`
and the last timestamp. The first line is never buffered: this log exists to
survive a process being killed mid-run. Only consecutive payloads collapse, and
payloads differing in any parameter are different payloads.

**E. The Debug Logs screen stays responsive.** It read the whole file and parsed
every line into a `LogEntry`, allocating a UUID each, synchronously from
`.onAppear`. Read and parse now run off the main thread and only the last 3000
lines are rendered. Copy still yields the complete retained file, and a header
row states how many older lines the display dropped.

## Measured, before and after

Replayed line by line against `dictus-logs 18.txt` — the saturated 200 016-byte
export from the #249 device validation that the issue itself measured. Script
applies each filter this PR introduces. 4 mic taps over 4.0 minutes.

| | before | after |
|---|---|---|
| bytes | 199 887 | 138 691 |
| lines | 1 407 | 881 |
| bytes / dictation | 49 971 | 34 672 |
| bytes / minute | 49 559 | 34 386 |

Lines per state transition, whole session:

| transition | before | after |
|---|---|---|
| `statusChanged from=idle to=requested` | 44 | **4** |
| `statusChanged from=requested to=recording` | 48 | **4** |
| `overlayShown status=recording` | 46 | **4** |
| `overlayHidden status=idle` | 59 | **5** |

Exactly once per occurrence, one per mic tap.

Retained wall-clock span for this session shape: **4 min at the old 200 KB cap,
29 min at 1 MB** (20 min immediately after a trim, growing to 29 before the
next). Note this is a dense back-to-back validation session; 30 minutes of
ordinary use costs far less.

Nothing on the must-stay-readable list is lost — all still present after the
replay at unchanged counts: `coldStartDarwinFallback` (2), `coldStartURLReceived`
(2), `dictStatusChange_enter` (39, `activeID` intact), `computeKbHeight` (40),
`engineCollectResult` (5), `appDidEnterBackground` (6), `waveformHeartbeat` (97),
`transcriptionPerformance` (5), `localModelResolved` (2), `modelSelected` (3),
`modelCompilationCompleted` (1).

## Tests

483 DictusCore tests pass, including 11 new ones: low-water trim target,
line-boundary preservation after trim, the rewrite-frequency assertion (6000
appends to a saturated file cause ≤ 4 rewrites, not 6000), collapsing of 9
identical lines into 2, non-merging of non-consecutive duplicates, non-merging of
payloads differing in one parameter, run closed by the next different line,
`formatted() == timestamp + payload()`, and the export window header.

All three targets build. SwiftLint clean under `--strict`; the baseline is
regenerated in its own commit (two entries shift, no new violations, count stays
at 45).

## Not verified here

- **No device validation.** The measurements above are a replay of the real
  device export, not a fresh device session. The acceptance criterion asking for
  a live 30-minute session still needs running.
- **"No visible freeze" on the Debug Logs screen** needs the GUI. The parse
  itself was never the bottleneck (9 ms for 7000 lines, measured); the bounded
  `ForEach` and the off-main read are what address it.

## Out of scope, as instructed

No verbosity switch, no `SharedKeys` flag, no Settings row, no change to
`LogEvent.level`. The ~9 concurrent controller instances and the ~60 MB footprint
are untouched: this changes what they log, not how many there are.

## Summary by CodeRabbit

* **New Features**
  * Debug log loading now shows progress and distinguishes loading from unavailable states.
  * Large logs display the newest 3,000 lines while retaining complete content for copying.
  * Log exports include time range, line count, and size details.
  * Repeated consecutive log events are summarized to reduce noise.
  * Automatic scrolling follows the newest log entry.

* **Bug Fixes**
  * Improved log trimming, clearing, and export consistency.
  * Reduced duplicate status and overlay diagnostics for more accurate debugging.


Pivii:
## Device validation — passed

Physical device, build `1.8.0 (24)` at `0cd1e38`, iOS 26.5.2, Parakeet. A 19-minute real-usage session, 11 transcriptions (8 started from the keyboard, 3 from the app), one model switch, several cold starts. Export: 1643 lines / 264 923 bytes.

Measured on that export, not on a replay:

| Check | Result |
|---|---|
| `source=micButton` | **0** — the old grep target is gone |
| `statusChanged from=requested to=recording` | **8**, one per keyboard dictation |
| `overlayShown status=recording` | **8** |
| Collapsed runs | 12, all `repeated=1` |
| Window header | present and accurate |

Before this change each of those transitions was written roughly nine times, once per live `KeyboardViewController`.

**Retention, measured rather than derived.** Real usage produces **13.7 KB/min**, not the 34 KB/min the saturated replay in the PR body suggested — that export came from a dense back-to-back validation session. The file never reached the cap: 265 KB for 19 minutes. Projected against the 1 MB cap that is **62 to 73 minutes retained**, comfortably past the 30 the issue asks for.

The maintainer also confirmed by hand what the simulator could not: the Debug Logs screen opens without a freeze, the "older lines hidden" header is present, Copy yields the full file rather than the displayed tail, and the mic button animation is unchanged.

Commit `0cd1e38` raised the trim low-water mark from 70% to 85% before this session. Worth recording honestly: the real-usage rate means 70% would also have met the criterion (51 min). The change is margin for the dense case (~25 min instead of ~20), not a fix for a measured failure.

The dedup work (change D) contributes almost nothing on real data — 12 collapsed runs of two lines each. The reduction is essentially all change A, exactly as the PR body predicted. Worth knowing before anyone credits collapsing for the win.

## Review findings

Six actionable findings across three CodeRabbit passes. Four fixed in `e3398be` and `327ce87`:

- `reloadLogs()` had no generation token, so a slow reload could overwrite a newer Clear and put the cleared content straight back on screen.
- `clear()` reset the collapse state on `writeQueue` but wiped the file off it, leaving the two unordered against a concurrent `log()`.
- `.refreshable` dismissed its spinner immediately because the reload was fire-and-forget.
- Auto-scroll was keyed on `entries.count`, which is pinned at the 3000-line display limit once the file saturates — so it stopped following the newest line exactly when the log is busiest.

Making the reload a `Task.detached` also surfaced a latent problem GCD had been hiding: `parseEntries` is main-actor isolated like every member of a `View`, and the original `DispatchQueue.global` call already violated that invisibly. Warning today, error in Swift 6. Marked `nonisolated`.

Two findings deliberately not acted on:

- **Missing French strings on the Debug Logs screen.** The screen is English throughout, including its title and its pre-existing empty state, and #255 exists precisely because its reader is an agent. Localising two of its strings would make it inconsistent, not more correct.
- **`AutocorrectDebugLog.appendToLogFile` writing outside `writeQueue`**, which can interleave a line into a collapsed run. Real, but that entire type sits inside `#if DEBUG` and stays off unless explicitly toggled, so it cannot reach a shipped build. Routing it through `PersistentLog` would cross the privacy quarantine that file documents at length.

## Not validated

The 30-minute retention criterion was not observed directly — the session ran 19 minutes and the file never saturated, so no trim occurred during it. Retention is a projection from a measured rate, not an observation of a trimmed file.

The four post-validation commits changed the Debug Logs screen after the device session, so the pull-to-refresh and Clear paths as merged were verified by build and unit tests only.

Issue #142, opened 2026-05-02
Cold start overlay layout broken — host toolbar collapses into keyboard area, ghost text visible (~1/3 cold starts)


Pivii:
## Summary

Visual-only bug on cold start (app killed, keyboard triggers relaunch via mic tap). Approximately **1 in 3 cold starts** the keyboard area renders incorrectly: the **host app's input accessory toolbar collapses down into the keyboard region**, ghost / mirrored text from a stale frame is visible behind the toolbar, and the Dictus recording overlay sits at the bottom of the keyboard area rather than filling it.

Cold start otherwise functions correctly — recording starts, Dynamic Island shows REC, state transitions through recording → transcribing → ready cleanly. Only the layout is broken until the keyboard is dismissed and re-summoned.

<img width="645" height="1398" alt="Image" src="https://github.com/user-attachments/assets/63c15fef-6a69-469f-9402-4bf87b505adf" />

## Reproduction

- iOS: 26.x, iPhone 17 Pro (per latest dev environment)
- Build: current `develop` (post PR #140 merge, 2026-05-01)
- Steps:
  1. Force-quit Dictus from app switcher (or wait for iOS to evict it)
  2. Open a host app with rich text (Notes, Notion, etc.) and tap into a text field
  3. Switch to the Dictus keyboard
  4. Tap the mic button — keyboard launches DictusApp via URL scheme (cold start path)
  5. Stay on the keyboard view (do NOT swipe back to host)
  6. Observe layout: host text editing toolbar visible in middle of keyboard area, ghost mirror text behind it, overlay collapsed at bottom
- Frequency: ~1 cold start in 3 (intermittent)
- Recovery: dismiss keyboard (next field, globe-swap) and re-summon — layout recovers

## What's visible (see attached screenshot)

- Host app's text formatting toolbar (Aa, list, table, paperclip, pen, atom icons) rendered inside the keyboard region instead of above it
- Mirrored / ghost text fragment behind the toolbar (`J 03 J ali Q ene goog…`) — stale render not cleared
- Dictus recording overlay (gray panel with FR + mic button) compressed at the bottom, not filling the keyboard area
- At the very bottom, the keyboard's standard globe + small mic row is also visible — looks like multiple input accessory views are stacked

## Functional state

- Recording works: Dynamic Island shows REC, state machine transitions correctly (recording → transcribing → ready)
- Audio capture, transcription, and text insertion all succeed
- Bug is purely visual — no impact on dictation functionality

[...]

PR #280, opened 2026-08-01
fix(keyboard): stop a deallocated controller from disowning a live dictation (refs #260)

Pivii:
refs #260

> **Revised after review.** The first version of this PR keyed the fix on the dictation status at deallocation time. That misses the capture in the issue body entirely — see "What was wrong with the first attempt" below. The claims in the original PR body were wrong; this one replaces them.

## The defect

Keyboard-area ownership is last-writer-wins on appearance. Under controller churn the live keyboard's ownership is routinely overwritten by an instance iOS is about to throw away, and when that instance was deallocated, `deinit` released the area outright. The controller that is *actually* on screen is then left unable to present anything — and there is no path back, because its `viewWillAppear` already ran. That is `activeID=none`: every status change dropped by the ownership guards, the keyboard stuck in its typing layout while the app records.

## What changed

**1. `KeyboardOwnership` (new, DictusCore).** Three states: `.none`, `.owned(id)`, `.awaitingReclaim(previous:)`. The third reads as **ownerless and invisible** to every consumer, so relative to the outright release it replaces, nothing downstream observes any difference — a marker that is never claimed is inert, and `isKeyboardVisible` cannot stick `true`. What it adds is a licence: a controller that can prove it is on screen may claim the area.

**2. `deinit` marks rather than releases**, and does **not** consult the dictation status. Whether a session is in flight says nothing about whether the keyboard is still on screen.

**3. The claim is fail-closed on window attachment** (`viewIfLoaded?.window != nil || inputView?.window != nil`), not on lifecycle bookkeeping. Every controller that can reach a claim has appeared, been overwritten by a later appearance, and never been told it disappeared — which describes the live keyboard *and* the cached instances iOS never notifies (#128) equally. Only attachment separates them. A permissive test would let a stale controller **acquire** the area and lock the live one out (the claim is single-shot) — strictly worse than the bug. If the test is instead too strict, the claim never fires and behaviour is exactly what it is today; `claimRefusedOffScreen` and `hasWindow=` make that visible on device rather than silent.

**4. A deferred nudge** re-publishes the area mode when the area becomes reclaimable while a dictation owns it, so a controller does not wait for the next status transition. Async, because this is reached from the outgoing owner's `deinit`.

No branch touches the keyboard's declared height constraint or its timing (#166).

## What was wrong with the first attempt

`deallocating` took `sessionOwnsKeyboardArea` and only marked the area when it was true. In the issue-body capture the owner deinits at `14:14:36` and `keyboardMicTapped` is at `14:14:37` — the status is `.idle` at deallocation, so no marker was created and the on-screen controller was skipped exactly as on `develop`. Replaying that timeline against the old rule prints `state after deinit: none / can claim: false`. The unit tests passed because they took `sessionOwnsKeyboardArea: true` as a premise instead of deriving it from a status timeline; they are rewritten as event timelines that carry the status along, and the issue-body ordering is now a test in its own right.

## Verification

- `DictusApp` scheme builds clean on the iOS 26.4 simulator with both appexes embedded; `DictusKeyboard` scheme builds clean.
- DictusCore: 504 tests, 0 failures (16 for ownership, including both captured orderings and the off-screen claimant).
- `swiftlint --strict` clean; baseline regenerated as the config documents (46 entries before and after).

## Not verified — needs device

A keyboard extension cannot be exercised headlessly. Unverified: whether `window` is non-nil for the controller iOS is showing, which is what makes the claim fire at all.

- **Fixed if you see** `claimedOwnerlessArea … previousActiveID=none hasWindow=true` followed by `hostingSet_recording old=52.0 new=276.0`.
- **Inert if you see** `claimRefusedOffScreen … hasWindow=false` — the marker was there, the window test rejected the controller. Behaviour then equals `develop`; report that line and the test moves to a different signal.
- **Also expect** `registerControllerDeallocation … next=awaitingReclaim(…)` where `develop` printed `registerControllerDisappearance … ownsVisibility=true`.
- **Should not appear:** `dictStatusChange_skippedInactive … activeID=none`.

The four-second dead window in the 11:30 capture is **not** addressed and cannot be from inside the extension: by `appDidEnterBackground` every controller has had `viewDidDisappear` (subscription cancelled) and is deallocated a second later, so no code of ours is running with a view to draw. What changes there is that the successor takes over from `awaitingReclaim` instead of a state that claimed the keyboard had been dismissed.

https://claude.ai/code/session_013zHBRzTq5JLdBU5ptEmkGY


Pivii:
@coderabbitai review

Pivii:
## Device validation, 2026-08-01 — what was and was not proven

Build `426936c`, iPhone16,2 / iOS 26.5.2. 15 cold starts, one cancelled warm start, one warm start at the end as a control. 16 transcriptions completed, zero errors.

**Proven: no regression.**

| Signature | Occurrences |
| --- | --- |
| `skippedInactive` (the #260 failure signature) | 0 |
| `awaitingReclaim` created during a live recording | 4 |
| `claimedOwnerlessArea` / `claimRefusedOffScreen` | 0 |

All four marker episodes resolved cleanly through an ordinary controller appearance, within 0 to 2 seconds:

```
10:16:37  registerControllerDeallocation  419EC379 → awaitingReclaim(419EC379)  status=recording
10:16:37  reclaimNudge  mode=recording
10:16:39  registerControllerAppearance  2B8906CD  previousOwner=none  ownership=awaitingReclaim(419EC379)
10:16:39  hostingSet_recording  old=52.0 new=276.0
10:16:39  showsOverlayChanged   isShowing=true owner=2B8906CD
```

**Not proven: that this fixes anything.** In all four episodes `develop` would have reached the same end state, because recovery came from `registerControllerAppearance` — unconditional on `develop` too — not from the new claim path. No status change arrived inside any of the gaps, so there was nothing for the marker to rescue.

The claim never fired, so `hasWindow=` is still unanswered. That is the one fact the design depends on and it needs a real occurrence to settle. The probe stays in the build; the next occurrence in TestFlight will answer it.

## Why this is merged anyway

- Two independent reviews of the final design, no defect found in it.
- Non-regressive across 17 device dictations.
- It removes a behaviour that is wrong by code reading independently of reproduction: an owner that releases the keyboard area while a dictation is live, which is exactly what the `viewDidDisappear` guard was written for in #142 and which `deinit` bypassed.
- The targeted failure was observed once, ever, incidentally during the #257 validation. Blocking the branch on a repro that did not appear in 17 attempts has no end date.

## CI

**No CI run exists for this head.** The last build on this branch is `af3b21e`, which predates the redesign; the green check on the head is CodeRabbit's, not the build. Three attempts to trigger one — a push, a close/reopen, and an empty commit — produced no workflow run, and no run was created anywhere in the repository after 08:59 UTC despite Actions being enabled. This is a GitHub-side stall, not a branch problem.

Substituted verification, run locally on `056446e` by the reviewer:

- `swift test` in DictusCore: 504 tests, 0 failures (16 `KeyboardOwnershipTests`).
- `xcodebuild -scheme DictusKeyboard -destination 'generic/platform=iOS Simulator'`: BUILD SUCCEEDED.
- `swiftlint --strict`: 0 violations across 151 files.
- Baseline: 46 entries on both `origin/develop` and this head, zero added, zero dropped on (file, rule, matched text).

Merged with `--admin` for that reason. #260 stays open until `hasWindow=` is answered.

Issue #281, opened 2026-08-01
Cold-start dictation: the keyboard extension loses every live controller for ~10 s and gets none back


Pivii:
## Summary

After a cold-start dictation, the keyboard extension can lose **every** live `KeyboardViewController` and get none back for as long as 10 seconds, while the app keeps recording. The extension process survives — only the controllers are gone. Nothing in the process can draw a keyboard without a controller, so the user is left looking at a dead keyboard until they force an app round-trip by hand.

This is the mechanism behind the "worse variant" captured in the #260 comment thread, split out because it is **not** an ownership problem and PR #280 does not address it. #260 is now scoped to its issue-body capture: a controller that is alive but does not own the session (`activeID=none` → `skippedInactive`). Here there is no controller to own anything.

## Evidence

Device log, build `1.8.0 (24)` at `426936c@fix/260-keyboard-owner-gap`, iPhone16,2 / iOS 26.5.2, captured 2026-08-01 during the #260 device validation.

Cold-start dictation, mic tapped at 09:48:40. The keyboard comes back correctly at 09:48:43 — the last controller appears, takes ownership, expands and shows the recording overlay:

```
09:48:43  77A73C82  viewWillAppear_entry   status=recording storedStatus=recording coldStart=true
09:48:43  77A73C82  hostingSet_recording   old=52.0 new=276.0 status=recording mode=recording
09:48:43  489BFADF  showsOverlayChanged    isShowing=true mode=recording owner=77A73C82
09:48:43  77A73C82  viewDidAppear_settled  hostingConst=276.0 heightConst=276.0 status=recording memMB=22
```

**One second later every controller is torn down**, correctly refusing to release ownership on the way out, and then deallocated:

```
09:48:43  77A73C82  viewDidDisappear             memMB=21
09:48:43  77A73C82  skippedUnregister_activeSession  status=recording activeID=77A73C82
09:48:43  03C90410  viewDidDisappear             memMB=21
09:48:43  03C90410  skippedUnregister_activeSession  status=recording activeID=77A73C82
09:48:43  E6BBD869  deinit  memMB=21
09:48:44  03C90410  deinit  memMB=20
09:48:44  77A73C82  deinit  memMB=19
```

From **09:48:44 to 09:48:54 there is not a single live controller.** The full lifecycle trace over the window, every instance in the process:

| instance | viewDidLoad | viewDidAppear_settled | viewDidDisappear | deinit |
| --- | --- | --- | --- | --- |
[...]

Issue #118, opened 2026-04-14
Bug: transcription text silently not inserted into target app (intermittent)


Pivii:
## Summary

Intermittent bug where dictation completes backend-side but text does **not** appear in the target app's text field. User-facing symptoms: the "transcribing" sine wave appears to freeze after ~1 cycle, the overlay disappears, and no text lands in the field. Repro has happened a few times; reproducible without any app switch (same host app throughout).

## Observed session

Logs: `dictus-logs 124.txt`, session around `2026-04-14T13:37:43Z → 13:37:58Z` (Notes as host app, warm start).

Key entries:

```
1209  requestStop sessionID=200739F1
1210  engineCollectResult sampleCount=235347
1212  statusChanged recording→transcribing
1216  transcriptionStarted model=parakeet-tdt-0.6b-v3
1220..1232  statusChanged recording→transcribing source=micButton  (x7 duplicates)
1233  transcriptionCompleted duration=259ms words=45
1238  statusChanged transcribing→ready
1247  keyboardTextInserted
```

From the code path (`KeyboardState.handleTranscriptionReady` → `KeyboardState.swift:341`):
- transcription string was present in App Group
- `controller?.textDocumentProxy.insertText(transcription)` was called
- `PersistentLog.log(.keyboardTextInserted)` fired
- `HapticFeedback.textInserted()` + `onTranscriptionInserted?()` fired

Yet the user reports **no text in the Notes text field** and the transcribing animation "frozen" visually.

## Hypotheses

1. **Stale `textDocumentProxy`** — the proxy is technically non-nil (`keyboardTextInserted` fired) but may be disconnected from a live input session at that instant. `insertText` on a stale proxy is a silent no-op. The burst of 7 duplicate `statusChanged recording→transcribing source=micButton` entries (lines 1220–1232) right before insertion suggests state flutter that could coincide with an input session transition.

2. **UI-perceived freeze is a transition artifact** — transcription took 259ms, transcribing overlay only rendered ~24 frames before being dismissed. Not a real freeze, just a very short visible animation that feels abrupt.

## Current logging gap

`PersistentLog.log(.keyboardTextInserted)` confirms the call was made, but does not capture:
- whether `textDocumentProxy.hasText` changed before/after
- whether `documentContextBeforeInput` is non-nil (proxy connected)
- the `controller?.textInputMode` state
- whether the extension was in a hasFullAccess / input-session-valid state

[...]

PR #282, opened 2026-08-01
probe(keyboard): instrument and analyse the #281 controller teardown (refs #281)

Pivii:
Instrumentation and differential analysis for `refs #281`. **No behavioural fix.** See §0 first: the first pushed version of these probes crashed the extension and has been fixed. Every code change is observation-only: log strings and one counter. No control flow, lifecycle or layout is touched, and the height constraint is not gone near.

## 0. Correction — the first version of these probes crashed the extension

**The probes as first pushed (`d439467`) crashed `DictusKeyboard` on every launch.** Eight controllers reached `viewDidLoad` in 23 s, none got further, and iOS fell back to the Apple keyboard. Fixed in `355f1c3` by removal. Recorded here because the defect and the reason it survived verification both matter more than the patch.

**Cause.** `UIInputViewController.h:30` declares `@property (nonatomic, readonly, copy) NSUUID *documentIdentifier;` with **no nullability annotation**, in a header with no `NS_ASSUME_NONNULL_BEGIN` — note its siblings on lines 22-28 *are* explicitly `nullable`. Swift therefore imports it as a **non-optional `UUID`**, and a nil returned from ObjC traps in the value-type bridging thunk. It does return nil before the host input session is established, and `traitCollectionDidChange` — which the device logs show firing after `viewDidLoad_exit` and before `viewWillAppear_entry`, i.e. with no session — read it on every launch.

**The fix is removal, not defensiveness, because the read cannot be guarded.** Optional chaining on the property is a *compile error*, verified against the iOS 26.4 SDK:

```
error: cannot use optional chaining on non-optional value of type 'UUID'
```

Recovering the probe would take an ObjC shim re-declaring the property `nullable` behind a bridging header. That is a deliberate decision for the maintainer, not something to slip into a diagnostic PR.

**Hypothesis lost with it: whether the two replacement controllers edit the same text field.** That was the direct discriminator between the two readings of the teardown — "iOS rebuilt the input view for the same field" versus "the host app moved first responder mid-transition". Losing it means §4.1 and §4.3 below stay open on the *next* capture too, not just this one: a log showing the teardown will still not say whether the host app's first responder moved. `kbAppear` covers only the style-flip half. This is the single biggest hole in the instrumentation, and closing it needs the ObjC shim.

**A note on the log that shaped this diagnosis.** `PersistentLog.log` enqueues on `writeQueue.async` at **`.utility` QoS**, and every write is a coordinated whole-file read-modify-write. At process launch the main thread runs through several probe points in the time that queue flushes one line. **The last line in the log is therefore not the last line reached, and the log cannot localize a crash in this codebase.** The eight `viewDidLoad`-and-nothing-else entries do not mean the crash was immediately after `viewDidLoad`; they mean everything behind it was still queued. Worth remembering for any future crash investigation here.

## 1. The cause

**Best-supported, not confirmed.** One link in the chain is established across two independent occurrences; the final link — why iOS builds no replacement — is not, and the two logs cannot settle it.

### Method

Differential over six device logs on build `1.8.0 (24)`, same device. Two contain the failure (`38-bug281` at 09:48, `34` at 11:30 — the capture originally filed under #260), one is the 15-cold-start negative control (`39`), three are supporting captures (`35`, `36`, `37`).

### What separates a failing run from a healthy one

**The number of replacement controllers iOS builds during the return transition.**

Healthy cold start — all 15 in log `39`, and also the cold start that follows the failure in log `38` itself, seconds later in the same host app:

```
<APP> appDidBecomeActive
controller_N   viewDidDisappear
<APP> appWillResignActive
controller_N+1 viewDidLoad → viewDidAppear_settled  (recording, hosting 276)
controller_N   deinit
<APP> appDidEnterBackground          ← the replacement survives this
```

Exactly one replacement; peak of two controllers alive at once; it survives the transition completing.

Failing run (log `38`, and log `34` is structurally identical):

```
09:48:42  03C90410 viewDidLoad → settled (recording, 276)   ← replacement 1
09:48:42  03C90410 traitCollectionDidChange prevStyle=1 newStyle=2
09:48:43  <APP> appDidEnterBackground
09:48:43  77A73C82 viewDidLoad → settled (recording, 276)   ← replacement 2, while 1 is live and settled
09:48:43  77A73C82 viewDidDisappear + 03C90410 viewDidDisappear   ← both, together
09:48:44  E6BBD869 / 03C90410 / 77A73C82 deinit             ← zero live controllers
09:48:44 → 09:48:54   nothing
09:48:52  <APP> appDidBecomeActive  ← maintainer opens DictusApp by hand
09:48:54  163DDBFB viewDidLoad → recovers
```

Peak of **three** controllers alive. Measured across all six logs, the peak count separates every capture that shows #281 (peak 3) from every capture that does not (peak 2), with one qualifier below.

### The second marker: a light→dark trait flip

`traitCollectionDidChange prevStyle=1 newStyle=2` — the keyboard's `userInterfaceStyle` resolving to dark on a device in light mode:

| log | occurrences | #281 present |
| --- | --- | --- |
| `39` (15 clean cold starts) | 0 | no |
| `35`, `36` | 0 | no |
| `38` | 2 | **yes**, teardown 1 s after the second flip |
| `34` | 4, two of them at 11:30:38 and 11:30:42 | **yes**, teardown at 11:30:43 |

In both occurrences the flip lands on the first replacement controller, immediately before the second one is built. It is not a device appearance change: within log `38` the flip happens on two cold starts and not on the third, seconds apart in the same host app, so it marks a transient rendering context that appears during some return transitions and not others.

### Stated as a chain

1. A transient rendering context appears during the DictusApp→host return transition, marked by the dark trait flip. *Supported: 2/2 occurrences, 0/15 healthy.*
2. iOS builds a second input-view controller for it while the first replacement is already on screen and settled. *Supported: 2/2 occurrences, 0/15 healthy.*
3. When the transition completes (`appDidEnterBackground`), iOS discards the whole input-view hierarchy — both controllers at once. *Confirmed mechanically in both occurrences.*
4. **iOS then builds no replacement for 10 s. This step is not explained.**

Step 4 is the one that hurts, and the evidence argues *against* the obvious reading. Log `37` shows the same mass teardown, also preceded by a trait flip, at 08:48:04 — and iOS rebuilt immediately (`67751D0B viewDidLoad` in the same second). So the teardown alone does not produce #281. What differs must be whether anything still requires an input view afterwards, i.e. the host app's first-responder state — which the extension cannot observe, because during the gap it has no controller to observe from.

### Candidates the logs eliminate

- **Memory kill — eliminated.** `KeyboardState … action=init` appears once in log `38` (09:48:37) and never again in either log; the singleton is never rebuilt, so the process survives. Footprint 19-21 MB. The issue's reading is correct.
- **`inputBounds` progression — eliminated as a discriminator.** `430x932` at `viewWillAppear_entry` → `430x504` at `viewDidAppear_settled` is universal: 16/16 healthy controllers and 6/6 in the failing log. The issue flagged this as suspicious; it is normal.
- **Ownership — eliminated, as the issue states.** Every ownership probe in the window reads correctly.

### What the logs cannot do

The debug log has **one-second resolution** and interleaves two processes whose writes lag relative to each other (`<APP>` lines at 09:48:41 appear physically after `<KBD>` lines at 09:48:42). Cross-process ordering within a second — `appWillResignActive` against `viewDidDisappear` — is therefore **not recoverable from these captures**. The discriminators above are all intra-process, where write order is reliable, which is why the analysis stands without it.

Adding fractional seconds to the log timestamp is a one-line change in `LogEvent.isoFormatter`, but `PersistentLog.pruneOldEntries` parses timestamps with a plain `ISO8601DateFormatter`, which does **not** parse fractional-second strings and would silently keep every line — disabling pruning and the #255 budget. Left out deliberately; flagged here so the trap is on record if the maintainer wants it later.

## 2. The probes, and how to read the next capture

Four probes. One new log line per controller lifecycle; everything else is extra fields on lines that already exist and already carry per-instance ids and `memMB`, so #255's duplicate collapsing is unaffected. Worst case observed: 6 controllers in 35 s = 6 added lines. Log `39`'s 16 controllers would add 16 lines to 2405.

| probe | on | discriminates |
| --- | --- | --- |
| `live=N` | `viewDidLoad`, `viewDidDisappear`, `deinit` | the headline separator, peak 2 vs peak 3, without needing a script. `live=0` on a `deinit` marks the exact start of the window. |
| `fullAccess` / `kbAppear` | new `inputContext_settled` line at `viewDidAppear` | whether the controller that flips to dark is editing a field that requests `UIKeyboardAppearance.dark`. |
| `beingDismissed` / `movingFromParent` / `hasParent` / `hasWindow` / `hasInputWindow` | `viewDidDisappear` | "iOS is dismissing the keyboard" vs "iOS is rotating the instance". |
| `hasWindow` | `traitCollectionDidChange` | whether the dark flip lands on an attached controller or a detached one. |

Every read was typechecked against the SDK to confirm it imports as an Optional or a plain `Bool`, so none can trap on a missing host connection — see §0. `traitCollectionDidChange` deliberately reads no proxy property at all, because that callback runs before the input session exists. The one remaining proxy read (`keyboardAppearance`, explicitly `nullable` in the header) sits at `viewDidAppear` rather than `viewWillAppear` on purpose: the proxy read is a synchronous round trip to the host app, and `viewWillAppear` is the timing-sensitive path this failure sits on.

### Reading the next occurrence

**If `viewDidDisappear` shows `beingDismissed=false movingFromParent=false` with the window already gone:** UIKit dismissed nothing — iOS discarded the input-view hierarchy wholesale. The failure is a transition artefact and the missing rebuild is an iOS behaviour we cannot prevent from inside the extension.

**If `beingDismissed=true` or `movingFromParent=true`:** the keyboard was genuinely dismissed. This is the "user dismissed the keyboard" shape and the extension should be releasing state, not holding it.

**If `kbAppear` reads dark (`1`) on the controller that flips:** the flip is the input context — the keyboard is being laid out for a field requesting a dark keyboard. If `kbAppear` stays `.default` (`0`) through the flip, the dark appearance comes from somewhere other than the text field's traits.

**What no longer discriminates:** whether the host app moved first responder. That needed `documentIdentifier`, removed in §0. A capture can now show the teardown without saying whether the text field survived it.

**If `live=3` appears with no dark flip, or a dark flip appears with `live` never exceeding 2:** the two markers are independent and the correlation found here (2/2 vs 0/15) was coincidence on a small sample. Worth knowing.

## 3. Recommended fix — NOT IMPLEMENTED

**I do not recommend a behavioural fix on this evidence, and I recommend against one specifically in `KeyboardViewController` lifecycle or ownership.** #280 has already established that the ownership model handles the recovery correctly once a controller exists; the problem is that none does, and controller creation belongs to iOS. There is no in-extension lever that creates one.

The only in-process contributor I can see is worth naming and *not* acting on yet: `traitCollectionDidChange` calls `giellaKeyboard?.updateTheme(theme:)` on every colour change (`KeyboardViewController.swift:661-664`). If that triggers a layout pass or collection-view rebuild while iOS is mid-transition, our own code may be part of why iOS decides to rebuild the hierarchy. The new `hasWindow` field on that probe is there to measure whether the flip even lands on an attached controller. **This should not be changed until a capture shows the flip landing on a windowed controller during a failing run** — a theme-update change made on suspicion is exactly the #118 Phase 34 shape.

If the maintainer wants a mitigation before the mechanism is known, the defensible one is app-side and does not touch the keyboard: DictusApp already knows it is recording and already runs a Live Activity throughout the gap, so the recovery ("open Dictus and come back") could be made discoverable there. That is a UX mitigation, explicitly not a fix, and it carries none of the regression risk this file has a history of.

**Risk in any keyboard-side attempt:** the failing window is inside the cold-start handoff that produced #142, #128 and #166. Any change that alters when a controller registers, unregisters or lays out re-enters that window.

## 4. What I could not determine

1. **Why iOS creates no replacement for 10 s.** The extension has no controller during the gap, so it cannot observe anything. **What would settle it:** the host app's first-responder state during the window — unobservable from the extension. The nearest available proxy was `documentIdentifier` on the recovery controller, which is exactly the probe removed in §0. Restoring it via an ObjC shim is the only route to this answer.
2. **Whether the dark trait flip is causal or a co-symptom.** It is 2/2 in occurrences and 0/15 in healthy runs, and log `37` shows it preceding the same teardown shape — but log `37` recovered immediately, so the flip predicts the teardown, not the gap. **What would settle it:** `hasWindow` on the flip and `kbAppear` on the controller, both now instrumented. Note this is weaker than the original plan, which paired the flip with `documentIdentifier`.
3. **Whether the host app differed between captures.** Nothing in the log identifies the host app or its text field. The failing cold start and a healthy one seconds later in log `38` argue against a static host difference, but do not exclude the user having tapped a different field. This was to be settled by `documentIdentifier`; with that removed it stays open.
4. **Cross-process event ordering**, per the one-second-resolution limitation above.

## Verification

- `DictusKeyboard`, `DictusApp`, `DictusCore` all build (iOS 17 Pro simulator, `BUILD SUCCEEDED`).
- `swift test` in `DictusCore`: **516 tests, 0 failures**.
- SwiftLint clean, baseline regenerated, **entry count unchanged at 46** — the same two pre-existing `KeyboardViewController` length violations, re-recorded at shifted spans, which is the case `.swiftlint.yml` documents.
- Every ObjC-imported read in the probes typechecked against the SDK to confirm it imports as an Optional or a plain `Bool`.
- No simulator GUI was launched.

### What this verification cannot catch — and did not

The first version of these probes passed every check above and still crashed the extension on all eight launches. That is a property of the verification, not an accident:

- **Nothing here executes the extension.** `xcodebuild build` compiles and links; CI does the same. Neither loads `DictusKeyboard` into a host process. The 516 `DictusCore` tests do not touch the keyboard target. "Builds clean, tests pass" verifies compilation and nothing else for this target, and should not have been reported as verification.
- **The compiler cannot warn.** On an unspecified-nullability import Swift makes the property non-optional, so it *guarantees* non-nil and rejects the guard. No warning, no lint rule, no static signal of any kind exists for this class.
- **The simulator would not have caught it either**, GUI or not: enabling a third-party keyboard with Full Access is a Settings flow `simctl` cannot script — already documented on #260 when simulator automation was considered and rejected.

The class this misses is anything that only manifests once the extension is hosted: nullability-contract violations in ObjC-imported APIs, host-connection-dependent behaviour, input-session lifecycle. **The only gate is "does the keyboard present at all" on a device**, and for this target it belongs ahead of any feature testing — including for changes as apparently inert as adding log fields, which is precisely the reasoning that skipped it last time.

**Not verified — needs the maintainer.** The probes have never emitted successfully on a device. Before any dictation testing: open any app, tap a text field, switch to Dictus, confirm the keyboard presents. Then one cold-start dictation to confirm the new fields look sane.

## Summary by CodeRabbit

* **Diagnostics**
  * Added keyboard lifecycle diagnostics covering creation, appearance, dismissal, disappearance, and cleanup.
  * Added live controller counts and visibility into keyboard attachment state, input context, document identity, full-access status, and appearance details.
  * Diagnostics are observation-only and do not alter keyboard layout or lifecycle behavior.


Pivii:
## Device validation 2026-08-01, build `355f1c3`

**Smoke test passed** — the keyboard presents. That is the gate the first version of these probes failed, and it is now the first check on any DictusKeyboard change.

One warm start and one cold start, two transcriptions completed, no errors. Every probe emits and the values are coherent.

### The census tracks correctly

```
11:46:52  315905FE  viewDidLoad  live=1
11:46:54  315905FE  deinit       live=0
11:46:54  4262D3EA  viewDidLoad  live=1
11:47:07  B8591AAB  viewDidLoad  live=2
11:47:08  4262D3EA  deinit       live=1
11:47:14  B8591AAB  deinit       live=0
```

Peak 2 — the healthy signature. `live=0` on a deinit marks the empty window exactly as intended.

### The teardown flags answer one of the two original questions

Every `viewDidDisappear` in this capture reads identically:

```
beingDismissed=false movingFromParent=true hasParent=true hasWindow=false hasInputWindow=false
```

`beingDismissed=false` with `movingFromParent=true` is UIKit removing the controller from its parent, not dismissing the keyboard. On this evidence the routine churn is instance rotation. Whether the #281 mass teardown reads the same is the discriminator — if it does, the keyboard was never "dismissed" and the missing rebuild is the whole bug.

### Counter-example: the style flip is not sufficient

The PR identifies `userInterfaceStyle` flipping light→dark as a marker that separated 2/2 failures from 15 clean cold starts. **This capture contains the flip and no #281:**

```
11:46:52  315905FE  traitCollectionDidChange  prevStyle=1 newStyle=2 hadColorChange=true hasWindow=true
```

`315905FE` then went through an ordinary `viewDidDisappear` / `deinit` with a clean successor, peak live stayed at 2, and both dictations completed.

So the flip is a correlate, not a sufficient condition. It may still be necessary — no capture without it has failed — but the sample that made it look decisive was 17 runs, and it now has a negative case. Worth correcting in the analysis before anyone treats it as the mechanism.

`hasWindow=true` on that line also answers the sub-question the probe was added for: the flip lands on an **attached** controller, so `traitCollectionDidChange` → `updateTheme` runs on a live keyboard rather than a detached one.

### `kbAppear` discriminates as designed

`kbAppear=1` on the first controller and `kbAppear=2` on both later ones — different text fields requesting different keyboard appearances. The field-dependent signal works, which is what makes it usable as a proxy for "did the input context change" now that `documentIdentifier` is unavailable.

Merging so the probes ride along in TestFlight, where the next occurrence will be captured without anyone hunting for it.

Issue #146, opened 2026-05-04
Tighten SwiftLint progressively toward --strict


Pivii:
## Goal

Move SwiftLint from its current lenient setup to a `--strict` configuration that blocks any new style regression — without forcing a one-shot cleanup of the existing 123 warnings.

## Context

PR #145 introduced the SwiftLint job with a deliberately permissive `.swiftlint.yml`:

- No `--strict` flag
- Most rules at `warning` severity (including `force_unwrapping`, `force_cast`, `force_try`)
- Length thresholds sized so the current codebase passes
- `DictusKeyboard/Vendored/` excluded entirely
- 123 violations / 0 errors on `develop` at time of writing

This was a pragmatic starting point to ship the CI without blocking work. The cost: warnings are silent — they show up in CI logs but nothing prevents a contributor (or me, in 6 months) from adding more.

### Step 1 — Current state (already shipped via #145)
- Config in place, CI runs on every PR, no blocking on warnings
- Baseline: 123 warnings, 0 errors

### Step 2 — Freeze the past, block the future
- Generate a SwiftLint baseline file: `swiftlint baseline --write`
- Commit `.swiftlint-baseline.json`
- Update `.swiftlint.yml` to reference the baseline:
  ```yaml
  baseline: .swiftlint-baseline.json
  ```
- Update workflow to run with `--strict`
- Net effect: legacy violations are tolerated, **any new violation fails CI**
- This is the highest-ROI step — minimal cost, maximum protection

### Step 3 — Cleanup by rule (multiple small PRs)

Knock out the warning backlog rule by rule, in ROI order:

1. `trailing_comma` (~30 violations, mechanical 1-line fixes)
2. `implicit_optional_initialization` (~10, trivial)
3. `empty_enum_arguments` (~5, syntactic sugar)
4. `unneeded_parentheses_in_closure_argument`, `vertical_whitespace`, `for_where`, `redundant_discardable_let` (low count, mechanical)
5. `function_body_length` / `type_body_length` / `file_length` (refactor work — already flagged, more time-consuming)
6. `force_unwrapping` / `force_cast` / `force_try` (case by case — some are legitimate with a comment, others should `guard`)
7. `cyclomatic_complexity`, `large_tuple`, `function_parameter_count`, `nesting`, `inclusive_language` (long tail)

After each PR, regenerate baseline → it shrinks.

### Step 4 — Full strict
- Baseline empty → delete the file and the `baseline:` line
[...]

PR #284, opened 2026-08-01
chore(lint): resolve the force unwraps outside the keyboard extension (refs #146)

Pivii:
Removes the thirteen baselined `force_unwrapping` violations that live outside the keyboard extension, and shrinks the SwiftLint baseline from 46 to 33 entries. Scoped by the agent brief on the issue (comment of 2026-08-01), not by the original four-step issue body.

refs #146

## What changed

Every one of the thirteen sites fell into one of the brief's first two categories. **No site needed category (c)** — none of them required inventing a fallback, so nothing was left behind and nothing was converted into silent failure.

### (a) Provably non-nil — kept, with a `swiftlint:disable:next` directive and a stated reason

| Site | Why it cannot be nil |
| --- | --- |
| `UnifiedAudioEngine.targetFormat` | `AVAudioFormat`'s failable init returns nil only for a format/rate/channel combination Core Audio cannot represent. All four arguments are compile-time constants, and 16 kHz mono non-interleaved Float32 is canonical PCM. |
| `withPrewarmTimeout` in `ModelManager.swift` | `TaskGroup.next()` returns nil only when the group has no unfinished children. Two tasks are added directly above and neither has been awaited. |
| `PaywallView.termsURL` / `.privacyURL` | String-literal URLs that cannot be malformed. The existing WHY comment was already there; only the directive was added. |
| `SettingsView` GitHub row | Same — string-literal URL. |
| Four `dictus://` deep links in `DictusLiveActivity.swift` | Same — string-literal URLs. |

The audio engine keeps its original multi-line initializer so its doc comment stays attached to the declaration; the directive sits on the line carrying the `!`. Writing it any other way trips `orphaned_doc_comment`.

### (b) Converted — no behaviour change on any input

**`SoundFeedbackService.volume()`** — was `(val ?? 0) > 0 ? Float(val!) : 0.5` where `val: Double?`. The ternary already collapsed a nil suite to `0`, and the `!` branch is only reachable when `val` is non-nil. Folding the `?? 0` into the binding produces the same `Float` for every input: nil and `0.0` both take the `0.5` default, anything `> 0` converts identically.

**`UserDictionary.recordUsage()`** ×3 — each force unwrap was guarded by the line immediately above it (a `!= nil` check, then a `[key, default: 0] += 1` that guarantees presence). Binding the value instead of re-subscripting keeps both counters and both return paths identical.

## Verification

- `swiftlint lint --strict` — 0 violations in 153 files, exit 0.
- The brief's baseline query returns **0 rows**.
- **No file under `DictusKeyboard/` is touched.** The nine force unwraps there stay frozen in the baseline, per the brief.
- Baseline regenerated in its own commit. All 24 non-`force_unwrapping` entries survive with their embedded counts untouched; exactly one moved (`function_body_length` in `DictusLiveActivity.swift`, 246 → 250) — pure line drift from the directive comments, same 81-line span, because the rule excludes comments.
- All three targets build clean for the simulator (`DictusApp`, `DictusKeyboard`, `DictusCore`), no new compiler warnings in any edited file.
- DictusCore suite: **521 tests, 0 failures**, 1 skipped (was 516 before this PR).

### New tests

`UserDictionaryTests` covers the two `recordUsage` counter branches the sweep rewrote. `recordUsage` is called from the keyboard extension on every word boundary and owns the pending → learned promotion, and had no coverage at all.

These tests pin the behaviour, not the rewrite: **the same five tests pass unchanged against the pre-sweep implementation** (verified by reverting the source file and re-running them). That is the evidence for the "no behaviour change" claim on the `UserDictionary` sites.

The skipped test is the below-threshold case — `repetitionThreshold` is currently `1`, so there is no below-threshold call to observe. It starts running by itself if the threshold is ever raised.

## Not verified here

- **No device validation.** Nothing in this PR changes runtime behaviour on any path, but the Live Activity deep links and the audio engine format are simulator-verified only (build + lint). See the recap for the exact device steps.
- `SoundFeedbackService.volume()` is `private static` with no seam, so its equivalence is argued by inspection above rather than by a test.

## Out of scope, untouched

The 9 keyboard force unwraps, the 18 length/complexity entries, the 6 misc entries, `.swiftlint.yml` itself, and the `baseline:` reference — all left exactly as they were.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sound feedback volume handling when saved preferences are missing or invalid.
  - Improved reliability when recording word usage, including learned-word and pending-count updates.

- **Tests**
  - Added coverage for word-learning thresholds, repeated usage, case-insensitive matches, and invalid short or empty entries.

- **Documentation**
  - Added clarifying documentation for audio processing, preloading, and navigation links.
  - Clarified safe handling of fixed-format audio and required app URLs.


Pivii:

## Device validation: three checks, and one correction

The PR body says no device validation is required. That is not quite right, and the reason is worth stating because it is easy to miss.

**No file under `DictusKeyboard/` was modified, but `UserDictionary.recordUsage()` runs inside the keyboard extension.** It is called from `DictusKeyboardBridge` and from `KeyboardRootView`, and DictusCore is embedded in the extension. So rewritten code does execute in the extension, and the project rule that any change reaching the keyboard gets a device smoke test applies here.

The risk is low: the rewrite is pure Swift dictionary manipulation with no ObjC bridging and no host-connection dependency, and the new tests pass against both the old and the new implementation. But the smoke test costs thirty seconds and the class of defect it catches is invisible to the build.

**1. The keyboard presents.** Open a text field in any app, switch to the Dictus keyboard, confirm it appears. Non-negotiable; nothing in the build or test pipeline ever executes the extension.

**2. Word learning still works.** Type an unusual word followed by a space, then confirm it is known. `repetitionThreshold` is currently `1`, so one occurrence is enough. This exercises all three rewritten `recordUsage` branches on the real path rather than in a test harness.

**3. Start and stop sounds.** Run a dictation and confirm the start and stop sounds play at their usual volume. This is `SoundFeedbackService.volume()`, which is app-side only and does not involve the extension.

Nothing else needs checking. The other nine sites are a comment and a lint directive each, with no compiled change.

Pivii:

## CodeRabbit review triage

**1. `UserDictionaryTests` — reset before each test. Valid, applied.**

Correct catch. The singleton reads App Group UserDefaults, which outlive the test process, so `tearDown` alone leaves a window: a run interrupted before it completes leaves entries behind and the next run's first test observes a word it never recorded. `setUp` now resets too. Tests-only change, 521 tests still pass.

**2. `SoundFeedbackService.volume()` — distinguish a stored `0.0` from a missing key. Declining, and the reason is not "out of scope".**

The suggestion describes a state that cannot be reached. The only writer of `SharedKeys.soundVolume` is the `@AppStorage` binding behind the volume slider in `SoundSettingsView`, and that slider is declared `in: 0.05...1.0`. There is no path through the UI that stores `0.0`, so there is no stored zero to distinguish from an absent key.

The behaviour is also unchanged by this PR in every case, reachable or not. Before, `(val ?? 0) > 0 ? Float(val!) : 0.5`; after, `stored > 0 ? Float(stored) : 0.5` with `stored = defaults?.double(forKey:) ?? 0`. A nil suite, an absent key and a stored `0.0` all collapsed to the `0.5` default before this PR and still do. The original code says so in its own comment: "treat 0 as use default".

Adopting the suggestion would be a deliberate behaviour change — making a stored zero mean silence — in a PR whose contract is that nothing changes behaviour. If muting via the slider is wanted, that is a feature question about the slider's lower bound, not a correctness fix here.

PR #289, opened 2026-08-01
chore(lint): empty and delete the SwiftLint baseline, make --strict honest (refs #146)

Pivii:
Finishes issue #146. The SwiftLint baseline is deleted, `swiftlint lint --strict` passes on a clean checkout with no exemption file, and the force rules are at `error`.

refs #146

## Why

33 violations were frozen in `.swiftlint-baseline.json`. 18 of them embed a line count in their match key, so any PR that changed the length of a frozen span — even shrinking it — turned CI red until someone regenerated the file. That cost a dozen regeneration commits in five weeks, four on `fix/260-keyboard-owner-gap` alone.

## 1. Length and complexity thresholds raised (18 entries gone, no code touched)

Per the maintainer decision on the issue: raise the thresholds, refactor nothing. The five worst offenders are also the five files with the worst regression history in this repo, four of them in or feeding the keyboard extension that nothing in CI ever executes. Restructuring them for a style property with no runtime consequence is the highest-risk change available here.

Maxima are measured from the lint run, not hand-copied:

| Rule | Measured max | Old warning | New warning | Headroom |
| --- | --- | --- | --- | --- |
| `type_body_length` | 558 (`LiveActivityManager`) | 400 | **650** | +16% |
| `function_body_length` | 127 (`TextPredictionEngine.spellCheck`) | 80 | **150** | +18% |
| `file_length` | 686 (`DictationCoordinator`) | 600 | **800** | +17% |
| `cyclomatic_complexity` | 18 (`AccentExpander`) | 15 | **20** | +2 branches |

Error levels moved up with them (900 / 250 / 1200 / 30). `cyclomatic_complexity` gets proportionally less headroom on purpose: it counts branches, not lines, so a single added `if` is a far bigger relative step than a handful of added lines. The reasoning and the measured maxima are recorded in a comment in `.swiftlint.yml`.

These rules now do the only job they can honestly do — catch a **new** type or function ballooning. Verified below.

## 2. The 6 miscellaneous violations

| Rule | Site | Fix |
| --- | --- | --- |
| `for_where` | `ModelManager.loadState` | outer `if` folded into a `where` clause |
| `function_parameter_count` | `AutocorrectDebugLog.bigramRerank` | 6 params to 4 by pairing each correction with its score |
| `nesting` | `DictusLiveActivityAttributes.ContentState.Phase` | `Phase` hoisted one level |
| `opening_brace` | `DictusKeyboardBridge` | brace onto the declaration line |
| `large_tuple` ×2 | `TextPredictionEngine` | `SplitEvaluation` and `BoundaryCandidate` structs |

On the `nesting` fix: unqualified `Phase` still resolves from inside `ContentState` through the enclosing scope, and the `String` raw values are untouched, so **the Codable payload ActivityKit persists is byte-identical**. Only one explicit reference existed (`LiveActivityManager.mapContentPhase`).

## 3. The 9 keyboard force unwraps

Bias inverted from PR #284, as the brief requires: documentation is the default, conversion the exception.

**Kept with a `swiftlint:disable:next` directive and a written reason (3):**

| Site | Why it cannot be nil |
| --- | --- |
| `KeyboardState:789` | `URL(string: "dictus://dictate?source=keyboard")!` — compile-time literal, well-formed absolute URL |
| `MicButtonDisabled:27` | `URL(string: "app-settings:")!` — same |
| `DictusKeyboardBridge:799` | `&&` short-circuits: the unwrap runs only once the `lastChar != nil` conjunct to its left has succeeded |

**Converted, with the equivalence argument per site (6):**

| Site | Change | Why identical for every non-nil input |
| --- | --- | --- |
| `AOSPTrieEngine:137`, `:141`, `TextPredictionEngine:418` | `prefix != nil ? (prefix! + x) : x` → `(prefix ?? "") + x` | non-nil: same concatenation. nil: `"" + x == x`, exactly what the `: x` arm returned. |
| `TextPredictionEngine:679`, `:688`, `:699` | `best == nil \|\| s > best!.score` → `improves(s, over: best?.score)` | the helper returns `true` on nil and the same `>` otherwise — the two branches of the original, unchanged. |

One trap worth naming: the tempting one-liner `s > (best?.score ?? 0)` is **not** equivalent. A first candidate scoring zero must still beat "no candidate at all", and `?? 0` silently drops it. The helper preserves that, and says so in a comment.

Nothing was left as category (c) — every site is either justified in place or provably converted.

## 4. Baseline deleted

- `.swiftlint-baseline.json` removed, `baseline:` line removed
- `force_unwrapping`, `force_cast`, `force_try` promoted from `warning` to `error` — only honest now that no unjustified instance remains. `--strict` already failed CI on a warning; `error` additionally fails a plain local `swiftlint` run, which is where a contributor should find these.
- Every reference updated: `.swiftlint.yml` header now documents the directive-and-reason convention, `.github/workflows/ci.yml` comment, `.coderabbit.yaml` path filter, `CONTRIBUTING.md`, and `.claude/skills/ship-issue/SKILL.md` (which told agents to keep the baseline stable).

## Verification

| Check | Result |
| --- | --- |
| `swiftlint lint --strict`, no baseline file present | **0 violations in 153 files**, exit 0 |
| `.swiftlint-baseline.json` exists | no |
| Deliberate new force unwrap / cast / try | 4 `error:` lines, **exit 2 on plain `swiftlint lint`** (not just `--strict`) |
| Deliberate 700-line type | flagged `Struct body should span 650 lines or less` |
| A 640-line type | passes — the threshold is a real boundary, not a disabled rule |
| DictusApp / DictusKeyboard / DictusCore | all `BUILD SUCCEEDED` (iOS Simulator, Debug) |
| DictusCore test suite | **521 tests, 0 failures**, 1 skipped |
| Every remaining force unwrap in linted code | 12, each with a directive and a stated reason |

The negative controls were run in a scratch file that was deleted; the working tree is clean.

## Correction to the brief

The brief says the TextPrediction code "is algorithmic code with test coverage". **It is not.** `DictusCoreTests` mentions `AOSPTrieEngine` and `TextPredictionEngine` only in comments describing logic that was migrated *out* of them into DictusCore. Neither engine is reachable from any test target. Every conversion above is therefore justified by local case analysis alone — the reasoning is in the table, not in a green test.

## What still needs a device

`DictusKeyboard` is touched, so the first check is that the keyboard still presents at all. Beyond that, three behaviours could plausibly be affected — and only these three:

1. **Autocorrect on apostrophe words** (`AOSPTrieEngine`, `TextPredictionEngine`): type `qu'ils` misspelled, e.g. `qu'ilss`, and confirm the correction keeps the `qu'` prefix rather than dropping it or doubling it.
2. **Compound-word splitting** (`trySplitWithSignal`): type `pasmal` and `tuboeux` in French; both should still split to `pas mal` and `tu peux`. This is the path the `improves()` helper and the two structs sit on.
3. **Autocapitalization after punctuation** (`DictusKeyboardBridge`): type `Bonjour. ` and confirm the next letter is capitalized; type `Bonjour ` and confirm it is not.

Not affected, but free to glance at: the mic button's disabled popover (`MicButtonDisabled`) and cold-start dictation launch (`KeyboardState` fallback URL) — both are force unwraps kept **byte-identical**, only a comment was added above them.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text prediction handling for apostrophes and split-word evaluations.
  - Preserved model state restoration more reliably for downloaded models.
  - Clarified live activity phase handling without changing its behavior.

- **Refactor**
  - Improved autocorrect diagnostics and prediction evaluation structure.

- **Documentation**
  - Updated contribution guidance with stricter Swift code-quality validation requirements.
  - Clarified how justified lint exceptions must be documented.


Pivii:
## CodeRabbit review — one applied, three declined with reasons

**Applied — pin the local SwiftLint version.** Real problem: CI pins its image, `brew install swiftlint` does not, and a rule added in a newer release fails a PR whose local run was green. `CONTRIBUTING.md` now tells contributors to check `swiftlint version` against the workflow and says CI is the authority. Commit `c764166`.

**Declined — `ContentState.Phase` compatibility typealias.** This one is not just unnecessary, it is self-defeating. I tested it:

```
DictusLiveActivityAttributes.swift:40:16: warning: Nesting Violation:
Types should be nested at most 1 level deep (nesting)
```

SwiftLint's `nesting` rule counts a nested `typealias`, so the shim re-triggers the exact violation this PR removes. Beyond that, `DictusCore` is a local SPM package inside this repo, not a published library — there are no external consumers whose source could break, and the single in-repo reference is updated in the same commit.

**Declined — `bigramRerank` compatibility overload.** Same reasoning on consumers: `AutocorrectDebugLog` is a debug logger in a local package with exactly one caller in the repo, inside `#if DEBUG`. A deprecated six-parameter overload forwarding to the four-parameter one would also keep the `function_parameter_count` violation alive, which is the violation the change exists to clear.

**Declined — `persist-credentials: false` on the lint job's checkout.** Fair hygiene point, and I am not disputing it. But the checkout step is untouched by this PR and the finding applies equally to the `build` job, so it belongs in its own change rather than riding along in a lint-configuration PR. Worth a separate issue if the maintainer wants it.

CI is green: SwiftLint pass, Build pass.

Issue #226, opened 2026-07-25
Add Chinese transcription support and clarify translation vs transcription


Pivii:
## Summary

User request from public group: add Chinese support to Dictus.

The user asked:

```text
I want chinese
I was wondering, why cant you use Chinese?
Actually, there is software called whisper, right? Wispr can translate to any language, but why can you set only these four languages?
```

## Current situation

Dictus currently exposes four supported languages in the app/keyboard:

- French (`fr`)
- English (`en`)
- Spanish (`es`)
- German (`de`)

Confirmed in `DictusCore/Sources/DictusCore/SupportedLanguage.swift`.

### Whisper / WhisperKit

OpenAI Whisper multilingual models support multilingual ASR and include Chinese (`zh`) in the language tokenizer list. Whisper `small` and `medium` both have multilingual variants.

Important distinction:

- **Transcription:** Chinese speech → Chinese text. This should be possible with multilingual Whisper models, subject to CoreML/WhisperKit model packaging and testing.
- **Translation:** Whisper's built-in translation task is mainly non-English speech → English text, not arbitrary translation into any target language. If a competitor translates “to any language”, they are likely using an additional translation/LLM layer, not just raw Whisper ASR.

### Parakeet TDT 0.6B v3

NVIDIA `parakeet-tdt-0.6b-v3` is documented as supporting 25 European languages:

`bg, hr, cs, da, nl, en, et, fi, fr, de, el, hu, it, lv, lt, mt, pl, pt, ro, sk, sl, es, sv, ru, uk`

Chinese is **not** in the documented Parakeet v3 support list.

## Product question

Do we want Chinese support as:

1. **Chinese transcription** — user speaks Mandarin/Cantonese/etc., Dictus outputs Chinese text.
2. **Chinese UI localization** — app interface translated into Chinese.
3. **Translation** — user speaks one language and Dictus outputs Chinese or another language.

The current user request most likely asks for transcription, but the wording also mentions translation because they compare to Wispr/Whisper.

## Proposed MVP scope

Add Chinese transcription support first, likely Mandarin/Simplified Chinese (`zh-Hans`) if WhisperKit supports it cleanly.

Tasks:

- [ ] Confirm available WhisperKit/CoreML models support Chinese reliably on-device.
- [ ] Add Chinese to the language model as a supported transcription language.
- [ ] Decide language code / display naming:
  - `zh`
  - `zh-Hans` / Simplified Chinese
[...]

PR #290, opened 2026-08-01
feat(keyboard): undo the last dictation insertion from the toolbar (refs #266)

### LINKED ISSUE #266: Undo the last dictation insertion from the keyboard toolbar
## Context

Captured during the #79 Smart Modes design session (2026-07-30).

Today, once a dictation result is inserted into the host app, the only way to remove it is to hold the backspace key. For a short transcription that is merely annoying; for a two-minute dictation it is genuinely painful, and it is the natural recovery action in three distinct situations:

- the transcription is wrong or the user changed their mind;
- the dictation landed in the wrong text field;
- **a Smart Mode did not deliver what was asked** (e.g. the user armed "translate to English" and French text came out) — see #79.

The third case is what surfaced the idea, but the feature stands on its own and is useful for every free user.

## Proposal

After the keyboard inserts a dictation result, offer a one-tap **Undo insertion** affordance that removes exactly the text that was just inserted.

### Safety contract (this is the important part)

The keyboard must never delete blindly. Before offering the affordance, and again immediately before acting on it, verify that the inserted text is still the tail of the field:

- read `textDocumentProxy.documentContextBeforeInput`;
- check it ends with the inserted string (including the trailing separator the pipeline appends — `". "` or `" "`, see `DictationCoordinator.swift:543-550`);
- if the check fails — the user typed, moved the caret, or the host app rewrote the text — **do not show the affordance at all**.

Known limitation to design around: `documentContextBeforeInput` returns a bounded amount of context, so for a long insertion the full string may not be visible. Verifying the longest suffix the proxy exposes is an acceptable weaker guarantee; the decision on how much is enough belongs to implementation.

### Deletion

`deleteBackward()` repeated for the character count of the insertion. Two caveats to handle: some host apps are slow to process a long burst of `deleteBackward()` calls, and the burst must not race with a new dictation starting.

### Lifetime

The affordance is transient. It disappears on:

- the first keystroke after the insertion;
- the start of another dictation;
- a short timeout (a few seconds);
- any failure of the safety check above.

## Placement

Proposal from the #79 session: the affordance lives in the **centre slot** of `ToolbarView`, not in the left slot.

The left slot is reserved for the hamburger button (see #241). Putting a destructive, transient action at the exact position of a persistent navigation control invites the muscle-memory error of reaching for the menu and deleting text instead.

The centre slot is free by definition right after an insertion (the user has not typed yet), and it is already transient. The resulting priority order for that slot:

| Priority | Content | When |
|---|---|---|
| 1 | Error message (red) | dictation or Smart Mode failure |
| 2 | Undo insertion | just after an insertion, while verifiable |
| 3 | Autocorrect suggestions | while typing |
| 4 | Armed Smart Mode name | idle, a mode is armed |
| 5 | "Long-press for Smart Modes" hint | idle, no mode armed |

Undo outranks suggestions because its window is short and it expires on the first keystroke.

**Open point:** the placement above is a recommendation from the design session, not a locked decision. The alternative considered and set aside was replacing the hamburger button with the undo button while the affordance is live.

## Acceptance criteria

- [ ] After a dictation insertion, an undo affordance appears in the keyboard toolbar
- [ ] Tapping it removes exactly the inserted text, including the trailing separator
- [ ] The affordance never appears when the inserted text is no longer the verifiable tail of the field
- [ ] The affordance disappears on the first keystroke, on a new dictation, and after a short timeout
- [ ] Undoing does not interfere with a dictation started in the meantime
- [ ] Works for long insertions without visibly freezing the host app

## Out of scope

- Redo / re-insert after an undo
- Undoing an insertion made in a previous keyboard session
- Any history or clipboard behaviour (see the History Pro feature)

## Related

- #79 — Smart Modes (the failure-recovery path that surfaced this)
- #241 — Keyboard toolbar rework (owns the left slot and the hamburger)

Pivii:
Implements the transient undo control for the last dictation insertion. refs #266

`Closes` is deliberately not used: it does not auto-close on a merge into `develop` in this repo. Close #266 by hand after merge.

## What changed

**`DictusCore/DictationUndo.swift` (new).** The safety check, as a pure function next to `AutocorrectReplacement` and for the same reason: it is the part that must not be wrong, and the keyboard extension has no test bundle. It answers one question — is the string we inserted still the tail of the field? A bounded proxy window is accepted on a weaker guarantee (the whole visible window must be a suffix of the insertion, and at least 24 graphemes of it), and the result records which of the two guarantees was obtained so a partial match is never read as a full one. 21 unit tests.

**`KeyboardState`.** `handleTranscriptionReady` had two copies of the insert-and-reset block: the direct path and the 100 ms retry that absorbs the App Group propagation race. They are now one `insertTranscription(_:)` called from both. The undo record has to be written by both, and the retry is the path that loses the race, so a copy that forgot it would fail intermittently and only on slow devices.

The record holds the exact string handed to `insertText`. There is nothing to reconstruct: `DictationCoordinator` appends the separator before writing to the App Group, so the separator (or its deliberate absence in Whisper auto-detect, #226) is already inside the recorded string.

It is destroyed — not hidden — on a new dictation, a keystroke, a caret move, the keyboard being dismissed, an 8 s timeout, and any failed check.

**The check runs twice.** Once one main-queue turn after the insertion, and that is what decides whether the control appears at all: a host that reports no usable context (a secure field, say) never sees it. Again immediately before deleting, because the first check proves nothing about the moment of the tap — a host that rewrites text without reporting it leaves a stale offer on screen, and that is caught here. A refusal deletes nothing and fires the refused haptic.

**Deletion** is chunked at 40 `deleteBackward()` per main-queue turn so a long dictation does not block the host, and a mic tap is refused while a burst is in flight so a burst and a new dictation can never interleave. Each chunk re-checks the dictation status as a second line of defence.

**Toolbar.** `dictationBar` gains a fourth branch, in the order settled during triage: error status message, then undo, then suggestions, then the hamburger. The control reuses the accent tint and uturn arrow of the autocorrect undo chip; the two do the same thing to different text and appear in the same strip of the same bar seconds apart. No geometry moves — the bar keeps its 52 pt and swaps content, so nothing here goes near the declared height constraint.

Strings ship in English and French.

## Verified here

- All three targets build (`DictusApp` scheme, iOS Simulator).
- `swift test` in DictusCore: 542 tests, 0 failures (21 of them new).
- `swiftlint lint --strict`: 0 violations.

## Not verified here

Everything that touches `textDocumentProxy`. Nothing in the build or test pipeline exercises a host app's text, and context reporting differs between hosts. The device checklist is in the issue thread; the short version is Notes, Safari, Messages and a chat app, one dictation of a few hundred characters, one insertion that arrives late enough to take the 100 ms retry path, and a length-limited field.

## Summary by CodeRabbit

* **New Features**
  * Added a temporary Undo option for the latest dictation.
  * Undo verifies the inserted text before removing it and supports Unicode text safely.
  * Added accessible Undo controls with haptic feedback.
  * Undo availability updates as text, cursor position, and keyboard state change.

* **Bug Fixes**
  * Prevented unintended undo actions after edits, cursor movement, new dictation, or keyboard dismissal.
  * Refreshed suggestions and capitalization after undoing dictation.

### DIFF
--- a/DictusCore/Sources/DictusCore/DictationUndo.swift
+++ b/DictusCore/Sources/DictusCore/DictationUndo.swift
@@ -1,0 +1,111 @@
+// DictusCore/Sources/DictusCore/DictationUndo.swift
+// Safety check for undoing the last dictation insertion (issue #266).
+//
+// WHAT THIS PROTECTS AGAINST:
+// The undo control deletes text the user cannot see being counted. If the
+// keyboard deletes `insertedText.count` characters without first proving that
+// those characters are still the ones sitting at the caret, it eats whatever
+// the user typed, pasted or moved the caret in front of instead. Deleting the
+// wrong text from someone's message is worse than offering no undo at all, so
+// the check below is the feature rather than a guard on it.
+//
+// THE INVARIANT:
+// Before the control is shown, and again immediately before it acts, the caller
+// re-reads the LIVE `documentContextBeforeInput` and this function confirms the
+// inserted string is still the tail of the field. Any doubt fails closed.
+//
+// THE BOUNDED WINDOW:
+// `documentContextBeforeInput` returns a bounded amount of text, and some hosts
+// return only as far back as the last line break. A two-minute dictation is
+// therefore routinely longer than anything the proxy will show. Refusing to undo
+// in that case would disable the feature exactly where holding backspace hurts
+// most, so a truncated window is accepted on a weaker guarantee: the whole
+// visible window must be a suffix of what was inserted. The result says which
+// of the two guarantees was obtained, so a partial verification is never
+// silently mistaken for a full one.
+
+import Foundation
+
+public enum DictationUndo {
+
+    /// Result of validating a pending dictation undo against the live context.
+    public enum CheckResult: Equatable {
+        /// Safe to undo: issue exactly `deleteCount` `deleteBackward()` calls
+        /// (one per grapheme cluster).
+        ///
+        /// - `verifiedCount`: how many trailing graphemes were actually matched
+        ///   against the document. Equal to `deleteCount` on a full match, lower
+        ///   when the proxy's window was too small to show the whole insertion.
+        /// - `windowTruncated`: true when the document context ran out before the
+        ///   insertion did, i.e. `verifiedCount < deleteCount`.
+        case ok(deleteCount: Int, verifiedCount: Int, windowTruncated: Bool)
+
+        /// Unsafe: show nothing, delete nothing. `reason` is a machine-readable
+        /// slug for device logs ("empty-insertion", "no-context",
+        /// "suffix-mismatch", "window-too-short", "truncated-mismatch").
+        case failed(reason: String)
+    }
+
+    /// How much of a truncated window must match before a partial verification is
+    /// accepted.
+    ///
+    /// WHY a floor at all: a short window is cheap to match by accident. A field
+    /// ending in ". " matches the tail of half the transcriptions this app
+    /// produces. Two dozen graphemes of running text is not something a different
+    /// field coincidentally ends with.
+    ///
+    /// This only ever applies to the truncated branch. A full match of a
+    /// three-character insertion is still a full match.
+    public static let minimumTruncatedMatch = 24
+
+    /// Validates that `insertedText` is still the tail of the field.
+    ///
+    /// - Parameters:
+    ///   - context: the LIVE `documentContextBeforeInput`, re-read at the moment
+    ///     of the check — never a value captured when the text was inserted.
+    ///   - insertedText: the exact string handed to `insertText`, separator
+    ///     included. `DictationCoordinator` appends the separator before writing
+    ///     to the App Group, so this is the whole of what reached the field.
+    public static func verify(
+        context: String?,
+        insertedText: String,
+        minimumTruncatedMatch: Int = minimumTruncatedMatch
+    ) -> CheckResult {
+        guard !insertedText.isEmpty else {
+            return .failed(reason: "empty-insertion")
+        }
+        // No context means no proof. Secure fields report nothing at all, and a
+        // host that reports nothing is exactly the host whose text must not be
+        // deleted on a guess.
+        guard let context = context, !context.isEmpty else {
+            return .failed(reason: "no-context")
+        }
+
+        let insertedCount = insertedText.count
+
+        // The window shows at least as much as we inserted: full verification.
+        // String comparison uses canonical equivalence, so a host reporting
+        // decomposed accents still matches a precomposed insertion — and both
+        // normalisations have the same grapheme count, which is what
+        // deleteBackward() consumes.
+        if context.count >= insertedCount {
+            guard context.hasSuffix(insertedText) else {
+                return .failed(reason: "suffix-mismatch")
+            }
+            return .ok(deleteCount: insertedCount, verifiedCount: insertedCount, windowTruncated: false)
+        }
+
+        // The window ran out first. Everything it does show has to be the tail of
+        // what we inserted; anything else means the field is no longer ours.
+        guard context.count >= minimumTruncatedMatch else {
+            return .failed(reason: "window-too-short")
+        }
+        guard insertedText.hasSuffix(context) else {
+            return .failed(reason: "truncated-mismatch")
+        }
+        // The delete count is still the full insertion. Deleting only the verified
+        // part would leave the head of the transcription stranded in the field,
+        // which is not an undo.
+        return .ok(deleteCount: insertedCount, verifiedCount: context.count, windowTruncated: true)
+    }
+}
--- a/DictusCore/Sources/DictusCore/DictationUndo.swift
+++ b/DictusCore/Sources/DictusCore/DictationUndo.swift
@@ -23,6 +23,20 @@
 // visible window must be a suffix of what was inserted. The result says which
 // of the two guarantees was obtained, so a partial verification is never
 // silently mistaken for a full one.
+//
+// WHAT THE TRUNCATED BRANCH DOES NOT PROVE, AND WHO FINISHES THE PROOF:
+// It says the tail of the field is the tail of the insertion. It says nothing
+// about the characters further back, which are exactly the ones a
+// `deleteCount` of the full insertion would go on to delete. On its own that
+// permits over-deletion: a field ending in text that merely resembles the
+// insertion's tail would authorise deleting a length that reaches past it.
+// The count returned here is therefore an intention, not a licence. The caller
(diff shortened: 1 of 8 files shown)


Pivii:
## Response to the review (e914b47)

**Over-deletion through the truncated branch — acted on, by a different route.**

The stated mechanism does not reproduce: after five backspaces the proxy window is `insertedText[35..<95]`, which is not a *suffix* of `insertedText`, so `hasSuffix` fails and the undo is refused. The window has to end where the insertion ends for the truncated branch to accept it, and non-periodic text does not do that after a deletion.

The underlying point stands and is worse than the example: the branch proved 24 graphemes and the caller then deleted 600 in fixed chunks. That is real, and it is reachable — a host that silently drops part of a long insertion (a length-limited field) leaves the recorded count larger than what is in the document.

Fixed at the caller rather than by returning a bound. The burst now re-proves the part still to go between chunks, and no chunk may be longer than the last check matched (`proven`). The proxy's window slides back over the text as the deletion proceeds, so every character removed has been seen at the tail of the field at some point. Where a host reports its context too late to prove the remainder, the undo stops and logs `dictationUndoAborted reason=... remaining=N` rather than deleting on: leftover text can be removed by hand, text eaten out of the middle of a message cannot. What the truncated branch does and does not prove, and who finishes the proof, is now written in the file header. Two tests model the loop, one of them on exactly the truncated-insertion case.

**Burst continuing into a different field — acted on.** The proxy is captured once at the start, so a controller attaching mid-burst cannot redirect the rest of the deletion. Pinning `activeControllerID` as well was considered and rejected: iOS churns roughly nine controllers per dictation (#281), and an ownership change during a 300 ms burst would abort a legitimate undo half-done. The per-chunk re-proof covers the field-change case on its own — a different field does not match the remainder.

**Non-editing keys — acted on.** Correct, and it contradicted a comment I had written two files away. Shift, the symbol layers, the globe and the emoji key no longer end the offer; the emoji key is told apart the same way `didTriggerKey` tells it apart, since it is an `.input` key carrying the smiley glyph.

**Hardcoded undo strings — not actionable.** `Text("Undo")` and `Text("Undo dictation insertion")` take `LocalizedStringKey`, and both keys are in `DictusKeyboard/Localizable.xcstrings` with French values in this PR's diff. The catalog's source language is English, so the key is the English string, which is how every other string in that file works (`Keyboard menu`, `Close menu`).

Pivii:
## Device validation checklist

Nothing in the build or test pipeline touches `textDocumentProxy`, so none of the behaviour below has been observed running. Context reporting differs between hosts, which is why the host list matters more than the number of repetitions.

The keyboard writes a probe line for every step (`PersistentLog`, component `KeyboardState`): `dictationUndoArmed` with the counts it verified, `dictationUndoInvalidated reason=...`, `dictationUndoStarted`, `dictationUndoCompleted`, `dictationUndoAborted reason=... remaining=N`. If something behaves oddly, that log says which branch ran.

**Per host — Notes, Safari (a text area and a single-line input), Messages, and one chat app (WhatsApp or Slack):**

1. Dictate a short sentence. The undo control appears where the hamburger was, tinted, with a uturn arrow. Tap it: the field returns to exactly what it held before, separator included. Nothing else is removed.
2. Dictate again and wait. After about 8 seconds the control disappears on its own, and the text stays.
3. Dictate, then type one character. The control goes immediately.
4. Dictate, then tap somewhere earlier in the text to move the caret. The control goes, and nothing is deleted.
5. Dictate, then tap shift, then 123, then the emoji key. The control should still be there after each — these do not touch the field. Insert an emoji from the picker: now it goes.
6. Dictate into a field that already has text. Undo. Only the dictation goes.

**Long insertion (the case the feature exists for):** dictate for a minute or more, a few hundred characters, then undo. Watch for two things: the host must not visibly freeze while it deletes, and the field must land exactly on what preceded the dictation. Leftover characters mean a chunk could not prove the remainder — the log line `dictationUndoAborted reason=... remaining=N` says so, and the number is how much was left.

**Retry path.** The insertion sometimes arrives through a 100 ms retry rather than directly. It is the same code path now, so it cannot behave differently, but it is worth catching once: the cold-start shape (kill DictusApp, then dictate from the keyboard) is the one most likely to take it. Undo must work there too.

**Two dictations in a row.** Dictate, do not undo, dictate again. The control must refer to the second insertion only: undoing removes the second and leaves the first.

**Racing a new dictation.** Undo a long dictation and tap the mic while it is still deleting. The tap should be refused with the refusal haptic and no recording should start; a tap straight after the burst finishes must start one normally (it must not be swallowed by the 1.5 s debounce).

**A length-limited field**, if you can find one (some sign-up forms cap input). This is the case where a host can silently drop part of the insertion. Expected: either the control never appears, or the undo stops early and leaves characters behind. What must NOT happen is text that predates the dictation disappearing.

**Password field.** The control must not appear at all — the proxy reports no context there, and the check fails closed.

**Autocorrect undo unchanged.** Type a word the keyboard corrects on space, then tap the undo chip in the suggestion bar. It must revert as before. This shares the bar with the new control but not the code.

Pivii:

## CodeRabbit review triage: 4 findings, all declined, each against the code

**1. "The truncated branch can delete more text than the insertion still contributes." (Major)**

The worked example does not hold. If a 100-grapheme insertion has 5 graphemes backspaced away, the field tail is the first 95 graphemes of the insertion. A 60-grapheme window over that tail ends at grapheme 95, so it is a *slice* of `insertedText`, not a suffix of it, and `insertedText.hasSuffix(context)` returns false → `.failed(reason: "truncated-mismatch")`. Nothing is deleted.

The deeper concern — that `deleteCount` is the full insertion while only the window was verified — is answered in the burst rather than in `verify`. Each chunk re-proves the remainder against the window that has slid back over previously unreachable text, and chunk size is capped at `proven`. No character is deleted that has not been seen to belong to the insertion. The code says this at length, and it is the correct place for the guarantee: `verify` cannot see past the proxy window by definition.

**2. "A deletion burst can continue into a different field." (Major)**

The premise is that `controller` is re-read each turn. It is not. `deleteInsertedText` takes `proxy: UITextDocumentProxy` as a parameter, captured once when the burst starts, precisely so a controller swap cannot redirect the rest of the deletion. The code documents this decision explicitly.

**3. "Do not clear dictation undo for non-editing keys."**

The invalidation is guarded by `editsDocument(key)`, and that function excludes the emoji key by its glyph and returns `false` in its `default` branch, which covers shift, the symbol layer and the keyboard switch. The exclusion asked for is already the implementation.

**4. "Localize the new undo strings."**

Both are catalog-backed. `Text("Undo")` and `.accessibilityLabel(Text("Undo dictation insertion"))` resolve through the string catalog, and `DictusKeyboard/Localizable.xcstrings` carries French for both — "Annuler" and "Annuler l'insertion de la dictée" — added in this PR.

Nothing changed in response to this review.

Issue #264, opened 2026-07-30
[UX] Avoid home screen flash before cold-start swipe-back overlay


Pivii:
## Context

When Dictus is force-quit and the user taps the keyboard microphone, the app correctly enters the cold-start flow and shows the swipe-back guidance. However, the normal Dictus home screen flashes briefly before the swipe-back overlay appears.

The user reproduced the same flash in both the current development build and TestFlight Lite, so this is independent of issue #262.

## Reproduction

1. Force-quit Dictus.
2. Open a text field in another app with the Dictus keyboard.
3. Tap the Dictus microphone.
4. Observe the transition while Dictus launches.

## Observed

The regular Dictus home screen is visible for less than one second before `SwipeBackOverlayView` replaces it. The flash is subtle but makes the cold-start transition feel unfinished.

Logs from 2026-07-30 show the cold-start sequence:

- `coldStartDarwinFallback`
- `coldStartURLReceived isColdStart=true`
- `coldStartFlagSet active=true`
- then the existing cold-start dictation flow continues.

## Expected

The first visible Dictus frame should already be the cold-start swipe-back overlay, with no intermediate home-screen flash.

## Initial technical hypothesis

`MainTabView` initially renders its normal `TabView` because `isColdStartMode` is set only inside `.onOpenURL`. The app may therefore render one frame of the default branch before SwiftUI applies the cold-start state.

The fix should preserve:

- the existing warm-start path;
- the cold-start recording behavior;
- the swipe-back overlay animation;
- the prepare-only model flow from issue #262.

A possible direction is to make the initial cold-start presentation state available before the first normal `TabView` render, or introduce a launch gate that renders the correct overlay while the keyboard URL is being handled.

## Priority

Important UX polish, but not a blocker.

Pivii:
> *This was generated by AI during triage.*

## Triage: hypothesis confirmed, and the obvious fix does not work

The issue body's hypothesis is correct, and reading the code sharpens it into something an agent can act on. It also rules out the first fix most people would reach for, which is worth recording before anyone spends an afternoon on it.

[...]

Issue #262, opened 2026-07-29
feat(keyboard): route cold-start dictation to model preparation instead of refusing the tap


Pivii:
## Context

The keyboard currently refuses a microphone tap whenever the shared `ModelLoadState` is `loading`. That guard is useful while the app is actively swapping models (#144), but it also blocks the cold-start path before Dictus can open.

This is especially visible after an app update, an iOS update, or another situation where the CoreML compilation cache is cold. The model may legitimately take tens of seconds or longer to prepare. The keyboard has no useful way to resolve that wait by itself, while Dictus already has the model-preparation UI and the cold-start navigation needed to bring the user into the app (#167).

The previous approach in #250/#254 attempted to make the keyboard infer freshness from a fixed 30-second cutoff. Device measurements showed that this is the wrong abstraction: legitimate loads can exceed the cutoff, while force-quit can leave a persisted `loading` value behind.

## Product decision

When the user taps the keyboard microphone while the model is not ready, the tap becomes a **prepare-only** action:

1. Dictus opens and shows the model-preparation screen.
2. No recording starts automatically, even after a long preparation.
3. Once the model is ready, the screen briefly confirms readiness and tells the user to return to their application and tap the microphone again.
4. The user explicitly starts recording with a fresh tap after preparation.

This is preferable to auto-starting a microphone session after a 30–120 second wait: the user's original intent may no longer be current, and starting audio without a new explicit action would be surprising and privacy-sensitive.

The existing fast path remains unchanged: when the model is already ready, tapping the keyboard microphone starts dictation without opening Dictus.

## Goal

Route cold-start model preparation through the existing app UI instead of leaving the user with a refusal message in the keyboard. The preparation UI should explain that this is an exceptional operation and should not happen before every dictation.

## Shared preparation component

`ModelLoadingOverlay` is the shared full-screen component used by onboarding and model management. It already distinguishes these phases:

- `downloading`: byte progress bar and downloaded/total counters;
- `compiling`: spinner and cycling optimization messages;
- `loading`: spinner and cycling load messages;
- `ready`: brief success confirmation before dismissal.

Keep one shared visual component, but add a preparation context so its explanatory copy and completion message can adapt without duplicating the screen:

- onboarding / initial model download;
- model selection or replacement;
- keyboard cold-start preparation.

[...]

PR #292, opened 2026-08-01
fix(app): decide cold start before the first frame (refs #264)

Pivii:
refs #264

## What changed

`MainTabView` decided cold start inside `.onOpenURL`, and SwiftUI evaluates `body` at least once before any URL handler runs. On a keyboard-launched cold start the home screen was therefore **always** drawn for a frame before the swipe-back overlay replaced it. The flash was deterministic, only its visible duration varied.

The launch URL is the only source of truth available early enough: iOS hands it to the scene at connection time, before SwiftUI builds any view.

1. **`DictusCore/KeyboardDictationURL`** — the rule "scheme `dictus`, host `dictate`, `source=keyboard`, `intent=prepare` selects the prepare path" now lives in one place, with unit tests. `MainTabView` and `DictusApp` both call it instead of hand-rolling query lookups.
2. **`DictusApp/ColdStartLaunch.swift`** — a scene delegate, installed from the existing `AppDelegate` via `application(_:configurationForConnecting:options:)`, records the launch intent in `scene(_:willConnectTo:options:)`. It creates no window; SwiftUI keeps owning the view hierarchy.
3. **`MainTabView.init`** seeds `isColdStartMode` / `preparingModelID` from that intent, so the first frame is already correct. The intent is forgotten when the scene goes to the background.

**On double-handling** (the hazard named in the brief): the early path deliberately does **not** consume the URL. `.onOpenURL` still fires for it and remains the only place that starts the dictation, writes the App Group cold-start flag, and flips the "has handled a URL in this process" flags — so an engine-dead restart is still classified as one. Re-asserting the same state from the handler is a no-op. Simulator logs confirm `coldStartURLReceived` / `coldStartFlagSet` / `dictationStarted` fire exactly once per launch.

## Verification

All on the iOS 26.5 simulator (iPhone 17 Pro), headless, using a throwaway opener app so the URL arrives app-to-app exactly as the keyboard sends it. Three targets build, `swiftlint lint --strict` reports 0 violations, `swift test` in DictusCore passes 530 tests (9 new).

| Criterion | Status |
| --- | --- |
| No home frame before the swipe-back overlay | **Needs device.** Verified on the simulator that the overlay is the first Dictus content drawn; the flash itself does not reproduce there (see below) |
| Normal launch shows the tab navigation, no splash or delay | Verified — screenshot shows the tab bar immediately |
| Engine-dead restart still reaches the overlay | Code path unchanged; **needs device** |
| `intent=prepare` shows the model loading overlay, never swipe-back | Verified — prepare cold start renders `ModelLoadingOverlay` |
| Dictation still starts on cold start | Partly — `dictationStarted fromURL=true` logged; no model installed in the sim, so no transcription |
| URL handled exactly once | Verified in the log |
| Background still clears cold-start mode, not the flag while recording | Logic untouched; **needs device** |
| The matching rule exists in one place | Verified — only `KeyboardDictationURL` matches; the keyboard still *builds* its URL strings inline |

**Honest limitation:** the one-frame flash does not reproduce in the simulator. I recorded the cold start at 60 fps with the fix enabled and again with it disabled: both go launch screen → cross-fade → swipe-back overlay, with no home-screen frame in either. So the simulator confirms the flow is correct but cannot demonstrate the flash is gone. That needs a device.

## Device validation

1. Force-quit Dictus. Open a text field in another app, switch to the Dictus keyboard, tap the microphone. Watch the first frame Dictus draws — a slow-motion screen recording makes a sub-second flash reviewable. Expected: swipe-back overlay, no home screen.
2. Same again, then speak and swipe back: the transcription must land in the host app, once.
3. With Dictus already running, kill the audio engine (Power button in the Dynamic Island), then tap the keyboard microphone: the swipe-back overlay must still appear.
4. Launch Dictus normally from the home screen: tab navigation, no splash, no added latency.
5. A prepare-only handoff (`intent=prepare`, #262): model loading overlay, never swipe-back.
6. Background the app mid-recording: the keyboard must keep its watchdog grace period (the App Group flag must not be cleared while recording).

https://claude.ai/code/session_01EwNiou6WfSt8vJzVUxpzQW

## Summary by CodeRabbit

* **New Features**
  * Improved keyboard dictation launch handling, including launches from a closed app.
  * Added support for preparing dictation before recording begins.
  * Preserved dictation state and overlays when opening or returning to the app.
* **Bug Fixes**
  * Improved recognition of keyboard dictation links regardless of parameter order or letter casing.
  * Prevented unsupported or unrelated links from triggering dictation.
* **Tests**
  * Added coverage for recording, preparation, validation, and fallback behaviors.

### DIFF
--- a/Dictus.xcodeproj/project.pbxproj
+++ b/Dictus.xcodeproj/project.pbxproj
@@ -20,6 +20,7 @@
 		AA000003 /* KeyboardViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA100003 /* KeyboardViewController.swift */; };
 		AA000004 /* KeyboardRootView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA100004 /* KeyboardRootView.swift */; };
 		AA000005 /* DictationCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA100009 /* DictationCoordinator.swift */; };
+		AA000264 /* ColdStartLaunch.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA100264 /* ColdStartLaunch.swift */; };
 		AA000006 /* DictationView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA10000A /* DictationView.swift */; };
 		AA000007 /* KeyboardState.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA10000B /* KeyboardState.swift */; };
 		AA0000F7 /* KeyboardLifecycleProbe.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA1000F6 /* KeyboardLifecycleProbe.swift */; };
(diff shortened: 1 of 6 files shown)


Pivii:
## CodeRabbit review triage: 1 nitpick, addressed by documentation

**"`ColdStartLaunch.intent` is nonisolated global mutable state."**

Real as a future concern, not as a bug today. The three call sites are `scene(_:willConnectTo:options:)`, `MainTabView.init` and the `scenePhase` handler, all main-thread by construction, and every target builds with `SWIFT_VERSION = 5.0` with no strict-concurrency setting.

The suggested `@MainActor` annotation is not the right fix here: `MainTabView.init` is nonisolated, so isolating the enum would make it unreachable from exactly the call site the fix exists for. Enforcing this properly belongs to a Swift 6 migration of the whole app target.

Documented the invariant on the type instead (`0d8a6c4`), which is the alternative the review itself offers for a Swift 5 target. No behaviour change.

Issue #233, opened 2026-07-26
Live Activity fails to bootstrap from idle after warm-state release / relaunch


Pivii:
## Summary

Fresh 1.8.0 (21) logs show the Dynamic Island / Live Activity no longer appears because `LiveActivityManager` fails to bootstrap a standby activity from `.idle`. Once this happens, the manager remains in `.idle`, so later recording/transcribing/ready transitions are rejected.

This is a different failure signature from the earlier #175 logs: in #175 the app-side lifecycle reached `standby` but the Dynamic Island was not visible. In these logs, the app-side lifecycle does **not** successfully create/recover a Live Activity after `.idle`.

## Environment

From exported debug logs:

- iOS: 26.5.2
- App: 1.8.0 (21)
- Revision: `56a01b2@fix/apostrophe-accent-expansion`
- Device: iPhone / `iPhone16,2`, 8 GB RAM
- Model: `parakeet-tdt-0.6b-v3`
- User report: Dynamic Island has not worked since the previous day / likely after latest update
- Force-killing the app and retrying does not fix it

## Logs analyzed

User-provided logs:

- `dictus-logs-DI-broken.txt`
- `dictus-logs-DI-apres-kill.txt`

### Earlier in the same log, Live Activity works

Before the regression point, normal transitions are present:

```text
2026-07-25T13:19:25Z liveActivityTransition from=idle to=recovery-standby
2026-07-25T13:19:25Z liveActivityStarted id=384AB99D-CF10-46F9-BBF7-ACBE30249D8A
2026-07-25T13:19:25Z liveActivityTransition from=standby to=recording
2026-07-25T13:19:44Z liveActivityTransition from=recording to=transcribing
2026-07-25T13:19:45Z liveActivityTransition from=transcribing to=ready
2026-07-25T13:19:46Z liveActivityTransition from=ready to=standby
...
2026-07-25T13:31:14Z liveActivityEnded reason=warmStateReleased
```

So ActivityKit / Live Activity support was working earlier in the session.

### After app launch / retry, standby recovery logs but no activity is started

At 14:10 after app launch:

```text
2026-07-25T14:10:20Z appLaunched version=1.8.0
2026-07-25T14:10:20Z liveActivityTransition from=idle to=recovery-standby
```

Expected next signal would be one of:

```text
liveActivityStarted id=...
liveActivityFailed context=startStandby error=...
```

But neither appears.

Then recording starts and the bootstrap path fails:

```text
[...]

PR #302, opened 2026-08-03
Stop a recording whose app is gone, and tell the user (refs #261)

Pivii:
refs #261

## Root cause, in my own words

`dictationStatus` is written forward and never audited. When iOS terminates DictusApp mid-dictation, the App Group keeps saying `recording`, and the next keyboard extension process faithfully restores a recording overlay for a session no process is running. The user keeps talking, their stop taps reach a dead listener, and 34 seconds later the session ends in silence with the audio lost.

The maintainer's comment sharpens what to do about it, and I took that as the contract: the keyboard must **stop the recording itself and tell the user**, without waiting for the app to relaunch and without waiting out the 15 s cold-start watchdog grace.

## Three things in the issue that the code contradicts

I checked every claim rather than trusting the list. Three did not hold, and two of them changed the design.

1. **There is no `dictationSessionID` in the App Group.** `SharedKeys` has no such key. The session id is `KeyboardState.activeSessionID`, in-memory in the extension process. So `sessionID=none` is what **every** rebuilt keyboard reports, whether or not the app is healthy — it is a symptom, never a cause, and it cannot be the discriminator. Criterion 2's "clear `dictationSessionID`" is satisfied by `forceResetToIdle()`, which already nils it.

2. **`requestStop` does not drop anything.** It writes `stopRequested=true` and posts the Darwin notification unconditionally. The two taps produced nothing because nobody was listening, not because the keyboard swallowed them.

3. **Criterion 3, read literally, would cause a regression.** "A `requestStop` arriving with `sessionID=none` … resets the keyboard locally" also describes a keyboard rebuilt mid-dictation while the app records perfectly happily — iOS rebuilds this extension routinely (~9 controllers per dictation, #281). Resetting there would drop the overlay during a live transcription. **The log line is unconditional; the local reset is gated on evidence.** That is a deliberate divergence and the one thing in here worth your veto.

## What the evidence actually is

`SharedKeys.recordingHeartbeat`. `UnifiedAudioEngine` writes it from the audio thread every 1 s while recording and every 3 s while the engine is warm but idle — which covers transcription (#106 Phase C, sized at 3 s precisely so it stays under the watchdog's 5 s threshold). It is the one value written by the process that would have died. An active status whose heartbeat stopped is decidable on the spot, with no grace to wait out. In the #261 capture the app died at 15:08:10 and the rebuilt keyboard read the state at 15:08:33 — a 23 s stale heartbeat, decidable at the first read.

## What changed

**`DictusCore/DictationSessionLiveness.swift`** (new) — `evaluate(status:heartbeat:now:)` → `.notActive` / `.live` / `.unproven` / `.orphaned`, threshold 8 s. Pure, no UserDefaults, 13 tests. It fails **closed** in three places, and those are the clauses that would break cold-start dictation if they ever regressed:
- `.requested` is never orphaned — the keyboard writes it moments before launching the app, so a leftover heartbeat would condemn every cold start;
- a missing heartbeat is `.unproven`, not `.orphaned` — an absent value is not evidence, and the existing watchdog still covers it;
- a clock that moved backwards reads as `.live`.

Why 8 s and not the watchdog's 5 s: the slowest legitimate write is the 3 s idle heartbeat, and a cross-process `UserDefaults` read can lag. 8 s clears both, and is still roughly half the 15 s grace this exists to bypass.

**`DictusKeyboard/KeyboardState.swift`** — `reconcileOrphanedDictation(source:)`, evaluated from `refreshFromDefaults` (so a rebuilt extension catches it on its **first read**) and from the watchdog tick (so a keyboard that outlives the app catches it too). On `.orphaned` it logs the event, posts a stop, resets, clears the recording keys, and sets a localised toolbar message.

**It posts a stop before resetting**, and that is the deliberate part. The verdict rests on a cross-process read. If that read were ever wrong, resetting alone would leave a live engine capturing with nobody watching — that is #60. A stop costs nothing when nobody is listening and salvages the recording when someone is, since `handleTranscriptionReady` inserts irrespective of the keyboard's local status. The cost is that a false positive shows "interrupted" and then delivers the text anyway. I would rather be confusing than silently orphan an engine.

**The message.** `String(localized: "Recording interrupted. Please try again.")` / « Enregistrement interrompu. Réessayez. » It goes through `statusMessage`, i.e. the same red toolbar line and the same 3 s lifetime as the existing `.failed` path — same register, reached from a different direction. It says *interrupted*, never *crash*.

**`DictusApp/DictationCoordinator.swift`** — `reconcileOrphanedDictationState()` at init. A process that has just started owns no session by construction, so a stored `recording` or `transcribing` is decidable without a heartbeat. **`.requested` is excluded**: the keyboard writes it moments before opening `dictus://dictate`, i.e. moments before launching this very process, and the `didBecomeActive` retry proceeds only while it is still there. Reconciling it would break every cold-start dictation — that is the trap in this issue and it is verified below, not merely reasoned about.

It writes `.failed` plus a localised `lastError` rather than `.idle`, so the keyboard's existing failure path tells the user. It does **not** go through `updateStatus`: this process is not failing anything, it is auditing what a dead one left behind, and its own `status` is correctly `.idle`.

**`DictusCore/LiveActivityStateMachine.swift`** — `.idle -> .failed` is now valid (issue change 3). `endWithFailure()` still returns early when it holds no activity, so this creates no pill out of nothing.

Not done, on purpose: **no case added to `DictationStatus`** (#267 owns that, and nothing here needed one), **no keyboard height constraint touched** (#166), **no change to `KeyboardOwnership`** (#260).

## Verified, with the commands and what they printed

- `xcodebuild build -scheme DictusApp -destination 'platform=iOS Simulator,id=231991C2…'` — **BUILD SUCCEEDED** (signed, no CI signing flags — an unsigned build makes `AppGroup.containerURL` nil, which is the surface being changed).
- `swiftlint lint --strict` — **0 violations, 0 serious in 159 files.** One `large_tuple` violation I introduced was fixed properly (a named struct), not suppressed.
- `cd DictusCore && swift test` — **603 tests, 0 failures**, 17 of them new.

### Simulator run, seeding the App Group and reading the exported log

`.recording` left behind by a dead process:

```
[2026-08-03T13:59:38Z] INFO    [lifecycle] <APP> appLaunched version=1.8.0
[2026-08-03T13:59:38Z] WARNING [dictation] <APP> dictationStateReconciled source=app-launch staleStatus=recording heartbeatAgeMs=-1
```

and the resulting shared state, read back from the App Group plist:

```
{'dictus.dictationStatus': 'failed', 'dictus.lastError': 'Enregistrement interrompu. Réessayez.'}
```

`recordingHeartbeat` is gone. The French string is the one that reached the shared error key, so the localisation is real and not just present in the catalogue.

`.transcribing` behaves the same:

```
[2026-08-03T14:00:50Z] WARNING [dictation] <APP> dictationStateReconciled source=app-launch staleStatus=transcribing heartbeatAgeMs=-1
```

**The negative control, which is the one that matters most.** Seeded `.requested` **plus a 600 s-old heartbeat** — the exact shape that would break every cold-start dictation if the guard were wrong:

```
[2026-08-03T14:00:22Z] INFO    [lifecycle] <APP> appLaunched version=1.8.0
        (no dictationStateReconciled line)
{'dictus.dictationStatus': 'requested', 'dictus.recordingHeartbeat': 1785765017.499578}
```

Status and heartbeat untouched. The cold-start path is not disturbed.

## What I could NOT verify

Plainly, because it matters more than the green checks above.

- **The reported failure itself is unverifiable by me.** It needs force-quitting DictusApp from the app switcher mid-dictation on a device. A simulator cannot reproduce a jetsam kill faithfully, and it cannot even enable the Dictus keyboard (`docs/agents/simulator.md` §7 — four documented attempts, all failed).
- **Nothing keyboard-side has been executed at all.** `KeyboardState` compiles and is covered by review; the predicate it consults is covered by tests. The message has never been rendered, the overlay has never been observed leaving, and the timing of "within ~1 second" is reasoned from the code path, not measured.
- **The heartbeat's cross-process propagation is assumed, not measured.** The existing watchdog already depends on exactly this and was device-validated in #106 Phase C, which is why I relied on it — but I have not watched a heartbeat cross the App Group myself.
- **Criterion 5 (a normal dictation across an app switch still completes) is entirely untested.** It is the regression that would hurt most.

## Known narrow false positive

Tapping Power in the Dynamic Island **during transcription** runs `deactivateSession()`, which stops the engine and therefore the heartbeat, while the transcription Task keeps running. About 8 s later the keyboard would declare an orphan and show "interrupted" — and then the text would arrive anyway, because insertion does not depend on the keyboard's status. Narrow (it needs a deliberate user action, mid-transcription), self-correcting, and the fail-safe direction. Recorded rather than hidden.

## Two things you should know that are not mine

- **`xcodebuild test -scheme DictusCore-Package -destination 'platform=iOS Simulator,…'` is broken on `develop`.** The scheme builds `polish-harness`, which is macOS-only by intent (`'SystemLanguageModel' is only available in iOS 26.0 or newer`) but is not excluded from iOS builds. I confirmed it by stashing my changes and re-running on a clean tree — same four failures. `swift test` is what actually runs the suite here, contrary to what the ship instructions say. Worth its own issue.
- The `.idle -> .failed` change interacts mildly with #233's no-activity suppression: a failure now moves the phase off `.idle`, so the next cycle logs one `bootstrap-fallback` line instead of one `bootstrapUnavailable` line. Same volume, different name. No flood.

## Device test protocol

**Do the smoke test first. A green build does not prove the extension launches.**

0. **Keyboard smoke.** Open any app, bring up the Dictus keyboard. It must present, type, and dictate normally. If it does not, stop here — nothing below is meaningful.

1. **The reported failure (~2 min).** Start a keyboard dictation, swipe back to the host app, then force-quit DictusApp from the app switcher while it is still recording.
   **Expect:** the recording overlay leaves within about a second, and the toolbar shows « Enregistrement interrompu. Réessayez. » in red for 3 seconds. No stop tap needed.
   **In the log:** `dictationStateReconciled source=keyboard-watchdog staleStatus=recording heartbeatAgeMs=~8000`, then `statusChanged from=recording to=idle source=keyboardState`. No `watchdogReset`.

2. **The rebuilt-keyboard variant.** Same kill, but dismiss the keyboard first and bring it back afterwards (or wait for iOS to rebuild the extension).
   **Expect:** the keyboard never shows a recording layout at all. `dictationStateReconciled source=keyboard-refresh` with a large `heartbeatAgeMs`.

3. **The stop tap (criterion 3).** Same kill, then tap stop before the reconciliation lands.
   **Expect:** something happens on the first tap. `requestStopWithoutSession storedStatus=recording liveness=orphaned` in the log, followed by the reconciliation.

4. **App relaunch after the kill.** Reopen Dictus by hand.
   **Expect:** `dictationStateReconciled source=app-launch` if the keyboard did not get there first, and nothing at all if it did.

5. **Regression, and the one I most want run (criterion 5).** Dictate normally from the keyboard, switch apps mid-dictation, come back, stop, and let it insert. Then do it two or three more times.
   **Expect:** completely unchanged behaviour. **Zero** `dictationStateReconciled` lines in the whole session. Any occurrence here is a false positive and this PR should not merge.

6. **Cold start.** Force-quit Dictus, then dictate from the keyboard so the app launches from scratch.
   **Expect:** the dictation works as today. No reconciliation line at launch.

https://claude.ai/code/session_016AL9tgVzFmZ1nY6S5Ko3bN

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when dictation is interrupted or unexpectedly terminated.
  * Automatically detects stale dictation sessions and resets them to an idle state with a temporary interruption message.
  * Prevented orphaned sessions from interfering with new dictation attempts.
  * Improved handling when stopping dictation without an active session.

* **Reliability**
  * Added heartbeat-based detection for live, stale, and unverified dictation sessions.
  * Preserved eligible dictation requests for retry after app relaunch.
  * Live Activity status can now transition from idle to failed when recovery is required.
  * Improved status-message visibility and lifecycle tracking.

### DIFF
--- a/DictusApp/DictationCoordinator.swift
+++ b/DictusApp/DictationCoordinator.swift
@@ -731,11 +731,19 @@ class DictationCoordinator: ObservableObject {
     /// `didBecomeActive`, which proceeds only while the stored status is still
     /// `requested`.
     ///
-    /// WHY this writes `.failed` rather than `.idle`: the keyboard's existing failure
-    /// path reads `lastError` and shows it, so the user is told the recording was
-    /// interrupted instead of watching the overlay vanish. WHY it does NOT go through
-    /// `updateStatus`: this process is not failing anything. It is auditing state a
-    /// dead one left behind, and `status` is correctly `.idle` here.
+    /// WHY it writes `.idle` and no `lastError`: this audit is hygiene, not speech.
+    /// The first version wrote `.failed` plus a localised error so the keyboard's
+    /// failure path would show it, which had two problems. By the time the app
+    /// relaunches the moment has passed — the keyboard reconciles in-process, within
+    /// seconds, and that is the surface that tells the user. And `failed` has no
+    /// expiry in the App Group: `KeyboardState.refreshFromDefaults` adopts a stored
+    /// `failed` and shows its `lastError` whenever the extension is next loaded,
+    /// however old it is. That is pre-existing behaviour, but writing `failed` here
+    /// on every reconciliation would have made a stale error message a routine sight.
+    ///
+    /// WHY it does NOT go through `updateStatus`: this process is not failing
+    /// anything. It is auditing state a dead one left behind, and `status` is
+    /// correctly `.idle` here.
     private func reconcileOrphanedDictationState() {
         guard let raw = defaults.string(forKey: SharedKeys.dictationStatus),
               let stored = DictationStatus(rawValue: raw),
@@ -746,13 +754,11 @@ class DictationCoordinator: ObservableObject {
             staleStatus: raw,
             heartbeatAgeMs: -1
         ))
-        defaults.set(
-            String(localized: "Recording interrupted. Please try again."),
-            forKey: SharedKeys.lastError
-        )
-        defaults.set(DictationStatus.failed.rawValue, forKey: SharedKeys.dictationStatus)
+        defaults.set(DictationStatus.idle.rawValue, forKey: SharedKeys.dictationStatus)
         // Clears the heartbeat, the waveform keys and the pending stop/cancel flags
-        // the dead process never got to clean up, and synchronizes.
+        // the dead process never got to clean up, and synchronizes. Dropping the
+        // heartbeat matters most: one that outlives its session is what the keyboard
+        // would otherwise judge the *next* session by (#261).
         cleanupRecordingKeys()
         DarwinNotificationCenter.post(DarwinNotificationName.statusChanged)
     }
--- a/DictusApp/DictationCoordinator.swift
+++ b/DictusApp/DictationCoordinator.swift
@@ -117,6 +117,9 @@ class DictationCoordinator: ObservableObject {
             defaults.synchronize()
         }
 
+        // Audit the shared dictation state before anything reads it (issue #261).
+        reconcileOrphanedDictationState()
+
         // Phase 37 instrumentation: snapshot device capabilities once per app launch
         // so the log stream has a stable anchor for per-device gating analysis.
         let snapshot = DeviceCapabilities.current()
@@ -711,6 +714,47 @@ class DictationCoordinator: ObservableObject {
         defaults.synchronize()
 
         DarwinNotificationCenter.post(DarwinNotificationName.waveformUpdate)
+    }
+
+    /// Clear a dictation the previous process never finished (issue #261).
+    ///
+    /// The invariant this restores: `dictationStatus` describes an active dictation,
+    /// but no live coordinator owns a session for it. That is trivially decidable
+    /// here — this method runs from `init`, so the process has just started and owns
+    /// no session by construction. Nothing else audited the key, which is why an app
+    /// terminated mid-recording left `recording` behind for the keyboard to restore
+    /// an overlay from.
+    ///
+    /// WHY `.requested` is excluded: the keyboard writes it moments before opening
+    /// `dictus://dictate`, i.e. moments before launching this very process. Clearing
+    /// it here would break every cold-start dictation — including the retry in
+    /// `didBecomeActive`, which proceeds only while the stored status is still
+    /// `requested`.
+    ///
+    /// WHY this writes `.failed` rather than `.idle`: the keyboard's existing failure
+    /// path reads `lastError` and shows it, so the user is told the recording was
+    /// interrupted instead of watching the overlay vanish. WHY it does NOT go through
+    /// `updateStatus`: this process is not failing anything. It is auditing state a
+    /// dead one left behind, and `status` is correctly `.idle` here.
+    private func reconcileOrphanedDictationState() {
+        guard let raw = defaults.string(forKey: SharedKeys.dictationStatus),
+              let stored = DictationStatus(rawValue: raw),
+              stored == .recording || stored == .transcribing else { return }
+
+        PersistentLog.log(.dictationStateReconciled(
+            source: "app-launch",
+            staleStatus: raw,
+            heartbeatAgeMs: -1
+        ))
+        defaults.set(
+            String(localized: "Recording interrupted. Please try again."),
+            forKey: SharedKeys.lastError
+        )
+        defaults.set(DictationStatus.failed.rawValue, forKey: SharedKeys.dictationStatus)
+        // Clears the heartbeat, the waveform keys and the pending stop/cancel flags
+        // the dead process never got to clean up, and synchronizes.
+        cleanupRecordingKeys()
+        DarwinNotificationCenter.post(DarwinNotificationName.statusChanged)
     }
 
     /// Clean up recording-related App Group keys after recording completes or is cancelled.
--- a/DictusApp/Localizable.xcstrings
+++ b/DictusApp/Localizable.xcstrings
@@ -1709,17 +1709,6 @@
         }
       }
     },
-    "Recording interrupted. Please try again." : {
-      "comment" : "Shown when a dictation is cleared because the app that was recording it is gone (issue #261).",
-      "localizations" : {
-        "fr" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "Enregistrement interrompu. Réessayez."
-          }
-        }
-      }
-    },
     "Reformulate your text with on-device AI" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/DictusApp/Localizable.xcstrings
+++ b/DictusApp/Localizable.xcstrings
@@ -1709,6 +1709,17 @@
         }
       }
     },
+    "Recording interrupted. Please try again." : {
+      "comment" : "Shown when a dictation is cleared because the app that was recording it is gone (issue #261).",
+      "localizations" : {
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Enregistrement interrompu. Réessayez."
+          }
+        }
+      }
+    },
     "Reformulate your text with on-device AI" : {
       "extractionState" : "stale",
       "localizations" : {
--- a/DictusCore/Sources/DictusCore/DictationSessionLiveness.swift
+++ b/DictusCore/Sources/DictusCore/DictationSessionLiveness.swift
@@ -1,0 +1,92 @@
+// DictusCore/Sources/DictusCore/DictationSessionLiveness.swift
+// Decides whether an active dictation status in the App Group still has a live
+// process behind it. Testable in isolation -- no UserDefaults, no AVFoundation.
+
+import Foundation
+
+/// What the shared dictation state says about the process that wrote it.
+public enum DictationSessionLiveness: String, Equatable, Sendable {
+    /// The stored status describes no dictation in flight. Nothing to reconcile.
+    case notActive
+    /// An active status backed by a heartbeat fresh enough to prove the writer is alive.
+    case live
+    /// An active status with no heartbeat to judge by. Not evidence of death.
+    case unproven
+    /// An active status whose heartbeat stopped. The writing process is gone.
+    case orphaned
+}
+
+/// Reads the shared dictation state and says whether a process still owns it.
+///
+/// WHY this exists (issue #261): when iOS terminates DictusApp mid-dictation,
+/// `SharedKeys.dictationStatus` keeps saying `recording` and nothing audits it.
+/// A keyboard extension rebuilt afterwards faithfully restores a recording overlay
+/// for a session no process is running, the user keeps talking into nothing, and
+/// their stop taps reach a dead listener.
+///
+/// WHY the heartbeat is the evidence and the session id is not: there is no session
+/// id in the App Group at all. `KeyboardState.activeSessionID` is in-memory in the
+/// extension process, so `sessionID=none` is what *every* rebuilt keyboard reports,
+/// whether or not the app is healthy. `SharedKeys.recordingHeartbeat` is the only
+/// value written by the process that would have died: `UnifiedAudioEngine` writes it
+/// from the audio thread every 1 s while recording and every 3 s while the engine is
+/// warm but idle -- which covers `.transcribing` too (#106 Phase C). A stored active
+/// status whose heartbeat stopped is therefore decidable without waiting for anything.
+///
+/// WHY an enum with statics (not a struct): there is no state to carry. The caller
+/// owns the facts; this type only combines them. Same pattern as `IdleReleasePolicy`.
+public enum DictationSessionLivenessPolicy {
+
+    /// How stale the heartbeat must be before the writer is declared gone.
+    ///
+    /// WHY 8 s and not the watchdog's 5 s: the slowest legitimate cadence is the
+    /// 3 s idle heartbeat, and a cross-process `UserDefaults` read can lag behind
+    /// the write. 8 s clears both with margin, and is still roughly half the 15 s
+    /// cold-start grace this check exists to bypass.
+    public static let staleHeartbeatThreshold: TimeInterval = 8
+
+    /// Whether `status` describes a dictation that some process should be driving.
+    ///
+    /// WHY an exhaustive switch and not a `Set`: adding a case to `DictationStatus`
+    /// (`processing`, #267) must stop the compiler here rather than silently fall
+    /// through to "nothing is in flight".
+    public static func isActive(_ status: DictationStatus) -> Bool {
+        switch status {
+        case .requested, .recording, .transcribing:
+            return true
+        case .idle, .ready, .failed:
+            return false
+        }
+    }
+
+    /// What the shared state says about the process behind it.
+    ///
+    /// - Parameters:
+    ///   - status: the stored `SharedKeys.dictationStatus`, or nil when unset or
+    ///     unrecognised.
+    ///   - heartbeat: the stored `SharedKeys.recordingHeartbeat` as seconds since
+    ///     1970, or nil when the key is absent.
+    ///   - now: seconds since 1970.
+    ///
+    /// Fails **closed** in two places, deliberately. `.requested` is active but is
+    /// never orphaned here: the keyboard writes it moments before launching the app
+    /// for a cold start, so a heartbeat left over from an earlier session would
+    /// condemn every cold-start dictation. And a missing heartbeat yields
+    /// `.unproven` rather than `.orphaned` -- an absent value is not evidence, and
+    /// the keyboard's own 5 s/15 s watchdog already covers that case. A clock that
+    /// moved backwards reads as `.live` for the same reason: killing a live
+    /// recording is the worse error.
+    public static func evaluate(
+        status: DictationStatus?,
+        heartbeat: TimeInterval?,
+        now: TimeInterval,
+        threshold: TimeInterval = staleHeartbeatThreshold
+    ) -> DictationSessionLiveness {
+        guard let status, isActive(status) else { return .notActive }
+        guard status != .requested else { return .unproven }
+        guard let heartbeat, heartbeat > 0 else { return .unproven }
+        let age = now - heartbeat
+        guard age > threshold else { return .live }
+        return .orphaned
+    }
+}
--- a/DictusCore/Sources/DictusCore/DictationSessionLiveness.swift
+++ b/DictusCore/Sources/DictusCore/DictationSessionLiveness.swift
@@ -10,7 +10,7 @@ public enum DictationSessionLiveness: String, Equatable, Sendable {
     case notActive
     /// An active status backed by a heartbeat fresh enough to prove the writer is alive.
     case live
-    /// An active status with no heartbeat to judge by. Not evidence of death.
+    /// An active status with no heartbeat that can speak about it. Not evidence of death.
     case unproven
     /// An active status whose heartbeat stopped. The writing process is gone.
     case orphaned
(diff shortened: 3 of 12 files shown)


Pivii:
## Round 2 — the device run found this wrong in both directions, and here is why

Build `1.8.0 (25)` at `0bce079` failed device validation. Both defects came from the same design hole, and the fix is a rule rather than two patched call sites.

### First, a correction to the diagnosis

The failure was read as *"nothing polls, so nothing can beat the watchdog"*. That is not what the code did. `reconcileOrphanedDictation(source: "keyboard-watchdog")` was already **the first statement of the watchdog tick**, evaluated once per second throughout the failure. It never returned true for one arithmetic reason:

| clock | trips at |
|---|---|
| the watchdog's own staleness test | waveform age > **5 s** (no cold-start grace was active — `coldStartFlagSet active=false` at 14:36:31) |
| my reconciliation | heartbeat age > **8 s** |

Both ages start at the same instant, because the app dies and stops writing waveform *and* heartbeat together. So the watchdog resets to `.idle` first, its `activeStates` guard then stops the timer, and 8 s is never reached. **A threshold above the watchdog's window is unreachable by construction.** App dies 14:36:51, `watchdogReset` 14:36:58; the reconciliation would have needed ~14:36:59.

This matters because it changes the fix. There is already a 1 Hz poll in exactly the right place — **no new timer**, which is also the right answer for a ~50 MB extension where a second timer would buy nothing the watchdog does not already provide.

### The rule the design was missing

> **A heartbeat can only speak about a session it postdates.**

One older than the moment this process started its current session belongs to an earlier session and says nothing about this one, however stale it looks. That fact is in-process (`activeSessionID` + a new `sessionStartedAt`) and needs no cross-process read.

Both defects fall out of it, which is why it is stated as a rule.

### Defect 1 — never fired

Fixed in two parts.

**The threshold now tracks the writer's cadence** instead of being one flat number: `.recording` → **4 s** (the audio thread writes at 1 Hz and that path `synchronize()`s every 200 ms), `.transcribing` → **8 s** (the warm-idle path writes every 3 s, #106 Phase C). 4 s is under the watchdog's 5 s, so the reconciliation now wins the race it structurally could not win before — on the device timeline it fires at ~14:36:55 instead of the watchdog's 14:36:58. `testTheRecordingThresholdIsUnderTheKeyboardWatchdogWindow` pins that ordering, because it is the whole fix and a comment would not survive the next edit.

**The watchdog's own reset routes through the reconciliation.** Its 5 s / 15 s waveform test decides *when* — those numbers are device-proven and I am not moving them in this issue — and the reconciliation decides what happens: clear both sides and say something, instead of resetting locally and silently. So a watchdog win and a predicate win are the same event to the user. `watchdogReset` is still logged, so that line keeps meaning what it means in every log written before this PR.

**Revised promise, honestly.** Criterion 1's "within ~1 second" holds only when the keyboard is *rebuilt* after the app died — the verdict is reached on the first read. When the keyboard stays alive the floor is the staleness window, **~4-5 s**, not ~1 s. What this PR adds in that case is that the reset now speaks and leaves the App Group clean, where before it was silent and dirty.

### Defect 2 — fired during a healthy cold start

Three things went wrong; the first alone is sufficient.

1. **The verdict ignored in-process evidence.** At 14:37:03 the keyboard held `sessionID=F67BC4F5`, started one second earlier by its own mic tap, and judged it against a heartbeat from 14:36:51. The postdating rule refuses that: `heartbeat <= sessionStartedAt` → `.unproven`. This is the fix, and it needs no timing assumption.
2. **`forceResetToIdle` left the App Group dirty.** It wrote `idle` but left the heartbeat, so a corpse survived to be read against the next session. It now clears `recordingHeartbeat`, `waveformEnergy`, `recordingElapsedSeconds` and `coldStartActive` — whatever gives up on a dictation leaves the App Group describing no dictation at all. `stopRequested` / `cancelRequested` stay untouched: those are commands in flight, not state, and clearing them would race the post below.
3. **The `.requested` guard was watching the wrong variable**, exactly as you said — the *stored* status, not the keyboard's own. Kept as documented defence in depth even though rule 1 now subsumes it.

Why my simulator control missed it: it seeded a clean `.requested` and asserted survival. The failure needs a stale `recording` **underneath** a live local `.requested` — local and stored state disagreeing, which one process's shared state cannot express. That is now `testLocalAndStoredDisagreementMatrix`, eight cells covering every combination of "do we own a session and when did it start" against "what does the App Group say", plus `testHeartbeatFromBeforeOurSessionProvesNothing` carrying the device's real numbers (11.609 s heartbeat, 1 s session).

### Question 4 — the stop does not survive the evidence, and is gone

I argued a stop "costs nothing when nobody is listening". The device shows the flaw: **a false positive is by definition the case where somebody is listening**, so the cost lands exactly where the insurance was supposed to pay out. `engineCollectResult sampleCount=0` → `statusChanged recording → failed` is that cost, and it turned a recoverable UI desync into a destroyed dictation.

**Replaced with a cancel**, which discards and returns the app to `.idle` — no `failed`, no zero-sample cascade, and the clean state the next mic tap needs. A no-op if the app is dead, exactly as the stop was. Doing nothing at all was the other candidate; I rejected it because it leaves the app stuck in `.recording`, where `startDictation`'s duplicate guard refuses every subsequent tap and the user has no way back.

**And only for a stale `recording`, not a stale `transcribing`.** `cancelDictation` also cancels the in-flight transcription task, and reaching the reconciliation during `transcribing` proves only that the audio *engine* stopped writing — not that the transcriber died. Those are independent (tapping Power in the Dynamic Island does exactly that). Posting a cancel there would destroy text that would otherwise still arrive, since `handleTranscriptionReady` inserts irrespective of the keyboard's local status. The app's own 30 s transcription watchdog covers the case where the transcriber really is stuck.

**I will not claim a false positive is now impossible.** What I claim is weaker on purpose: after the postdating rule, one requires an app that is alive, holds `status=recording`, and has not delivered an audio buffer for 4 s *during a session this keyboard itself started and has been watching*. That is not a healthy recording — it is the zombie-engine state `verifyAudioFlow` already treats as a failure, and cancelling it is the correct outcome rather than a regrettable one.

### The stale `failed` on keyboard load — not mine, and I stop feeding it

`refreshFromDefaults` adopts a terminal `failed` from the App Group and shows its `lastError` with no freshness rule, however old it is (14:36:33, `storedStatus=failed`, no reconciliation line). Pre-existing.

**Its own issue, not this PR.** Fixing it means giving `lastError` an expiry, which is a separate contract. What belongs here is not making it worse, so **the launch audit now writes `idle` and no `lastError` at all**. Verified below with a sentinel. The audit's job is hygiene, not speech: the keyboard tells the user, in-process.

### Verified

- `swift test` — **612 tests, 0 failures** (30 new since round 1).
- `swiftlint lint --strict` — **0 violations in 159 files.** The reconciliation moved into a same-file `private extension` to stay inside `type_body_length`, the same device `DictationCoordinator` already uses; nothing was widened to make room.
- `xcodebuild build -scheme DictusApp` — **BUILD SUCCEEDED** (signed).
- Simulator, seeded App Group, log read back:
  - `recording` + 40 s heartbeat → `dictationStateReconciled source=app-launch staleStatus=recording heartbeatAgeMs=-1`, resulting state `{'dictus.dictationStatus': 'idle'}`, heartbeat gone, **and a `lastError` sentinel left untouched** — proving the audit no longer writes an error string.
  - `.requested` + 600 s heartbeat → no reconciliation, both keys intact.

### What a simulator still cannot settle

**Neither defect is fixed until you say so on device.** A simulator cannot enable the Dictus keyboard, so every keyboard-side claim above is unit tests plus review. Both defects were invisible to the round-1 simulator run and I am not treating this one as stronger evidence than it is.

### Out of scope, flagged not fixed

`liveActivityFailed context=rejectedTransition error=recording->failed` at 14:37:03 in your log. `.recording` allows only `[.transcribing, .standby]`, so a dictation that fails *while recording* (permission, zombie engine, zero samples) cannot show the failure on the Dynamic Island. Same family as the `.idle -> .failed` gap the issue named, but the issue did not name it and there is an existing test pinning the current behaviour. Reported, not touched — your call whether it wants its own issue.

### Device test list, revised

**Smoke test first: a green build does not prove the extension launches.**

0. Bring up the Dictus keyboard. It must present, type and dictate normally.

1. **Defect 1 — force-quit mid-recording, keyboard on screen.** Start a keyboard dictation, swipe back to the host, force-quit DictusApp from the app switcher, keep the keyboard visible.
   **Expect:** the overlay leaves after ~4-5 s (not ~7 s, not silently) and the toolbar shows « Enregistrement interrompu. Réessayez. » in red for 3 s.
   **In the log:** `dictationStateReconciled source=keyboard-watchdog staleStatus=recording heartbeatAgeMs=~4000`. If you instead see `watchdogReset` followed by `dictationStateReconciled source=keyboard-watchdog-stale`, that is the fallback path working — also a pass, just slower.
   **Fail:** a bare `watchdogReset` with no reconciliation line, or no message.

2. **Defect 2 — the regression, and the one that must not fail.** Immediately after step 1, tap the mic again so the app cold-starts.
   **Expect:** a completely normal dictation that records, transcribes and inserts.
   **Fail:** any `dictationStateReconciled` line in that window, or `sampleCount=0`, or `statusChanged recording → failed`.

3. **Rebuilt keyboard.** Force-quit mid-recording, dismiss the keyboard, bring it back.
   **Expect:** it never shows a recording layout. `dictationStateReconciled source=keyboard-refresh` with a large `heartbeatAgeMs`.

4. **Stop tap.** Force-quit mid-recording and tap stop before the reconciliation lands.
   **Expect:** something happens on the first tap; `requestStopWithoutSession` in the log.

5. **App relaunch.** Reopen Dictus by hand after a kill.
   **Expect:** `dictationStateReconciled source=app-launch` if the keyboard did not get there first, and **no** error banner in the keyboard afterwards.

6. **Regression — three or four normal dictations**, including one across an app switch.
   **Expect:** unchanged behaviour and **zero** `dictationStateReconciled` lines. Any occurrence here is a false positive and this must not merge.

7. **Long transcription.** Dictate 20-30 s so transcription runs a while.
   **Expect:** no reconciliation during transcription, text inserted normally.

https://claude.ai/code/session_016AL9tgVzFmZ1nY6S5Ko3bN

Pivii:
## Round 3 — instrumentation for the message, no behaviour change

Round 2 passed on device: 5 `dictationStateReconciled`, 0 `watchdogReset`, including one at `heartbeatAgeMs=4829` where the old 5 s watchdog would have won. What is still unsettled is whether the user is *told*, and nothing in the log could answer that. This adds the trail. **No behaviour changed** — not the threshold, not the text, not the 3 s lifetime, not when the message is set. Every assigned value is byte-identical to before.

### Why three events and not one

The message is assigned on a singleton, but iOS keeps several controllers alive, so **"assigned" and "on screen in front of the user" are different facts** and only the first was observable. One "I set the message" line would reproduce the ambiguity we already have.

| event | answers |
|---|---|
| `dictationMessageSet reason= owner= visible=` | it was assigned, and what asked for it |
| `dictationMessageDisplayed rootView= controller= owner= visible=` | a live view actually put it on screen — **one line per view that rendered it** |
| `dictationMessageCleared reason= displayedCount=` | it ended, and how many views had rendered it over its life |

**`displayedCount=0` means nobody could have read it**, and that line is emitted at `WARNING` so it is findable by level as well as by reading the number.

`dictationMessageDisplayed` is the line that tests the hypothesis directly: compare `rootView`/`controller` against `owner`. A view that rendered the message while it did not own the keyboard area drew into a tree the user was not looking at.

### How it is wired

`statusMessage` is assigned from seven places across three lifecycles, so it gets a `didSet` funnel — the same argument this file already makes twice for `dictationStatus` and `ownership`: a property observer is the only path no writer can bypass. Writers name their reason through `assignStatusMessage(_:reason:)`; one that skipped the funnel would log `reason=unspecified` rather than silently inheriting the previous reason.

Reasons you will see: `reconciled-keyboard-refresh`, `reconciled-keyboard-watchdog`, `reconciled-keyboard-watchdog-stale`, `reconciled-keyboard-stop`, `reconciled-timeout`, `appError`, `appError-timeout`, `forceResetToIdle`, `requestCancel`, `textInserted`, `replaced`.

The rendering side is reported from the `Text` itself, via `onAppear`/`onDisappear` — the only place that can attest to it. `ToolbarView` gained two defaulted identity parameters; `KeyboardRootView` supplies its own `instanceID` and its `controllerID` at both call sites. The disappear side is a `diagnosticProbe` (`statusMessageHidden`) rather than an event, since it also fires when the toolbar itself goes away.

**Cost:** one line when a message is assigned, one per view that renders it, one when it ends. The message changes a handful of times per session, so this is proportional to the message and not to the render loop. Nothing fires per frame or per render tick.

### Verified

- `swift test` — **615 tests, 0 failures** (3 new, pinning the exact line format the device script greps for, so the names and field order are a contract).
- `swiftlint lint --strict` — **0 violations in 159 files.**
- `xcodebuild build -scheme DictusApp` — **BUILD SUCCEEDED** (signed, no CI signing flags).

### What this instrumentation cannot prove

- **None of it was executed.** Every event added here fires only inside the keyboard extension, and a simulator cannot enable the Dictus keyboard (`docs/agents/simulator.md` §7). There is no simulator evidence for any of these lines; what is verified is that they compile and render as tested.
- **`dictationMessageDisplayed` proves a view rendered the message, not that a human saw it.** In `.keys` mode every live root view renders the toolbar, whether or not it is the one on screen. That is precisely why `owner` and `visible` travel on the line — the comparison is the inference, and the log makes it possible rather than making it for you.
- **It cannot see a message clipped, covered, or drawn off-screen.** If `displayedCount` is healthy and the maintainer still sees nothing, the cause is below this layer and the next step is a screen recording, not a log.
- **"Contradictory content" is not directly instrumented.** Two views rendering *different* text cannot happen — they read one `@Published` value — so the trail will show a `replaced` clearing if the message really changed, and nothing if it did not. That distinction is the closest this gets to the report.

### Device test script

Two things make a run readable, and both matter more than usual here: **note the wall-clock time of each mic tap**, and run the steps in order without clearing the log in between.

Greps to run afterwards, in this order:

```
grep -E "dictationMessageSet|dictationMessageDisplayed|dictationMessageCleared" log.txt
grep "dictationMessageCleared" log.txt | grep "displayedCount=0"
grep "dictationStateReconciled" log.txt
```

---

**0 — Keyboard smoke.** Bring up the Dictus keyboard in any app. Type a few characters, dictate once normally, let the text insert.
*Screen:* everything normal, no red message at any point.
*Log, expected:* a `dictationMessageCleared reason=textInserted displayedCount=0` per dictation is **normal and not a failure** — it is the funnel reporting that there was no message to clear. **No** `dictationMessageSet`. If a `Set` appears here, something raised a message during a healthy dictation and that is worth stopping for.

**1 — Reconciliation while the overlay is live.** Start a keyboard dictation, swipe back to the host app, force-quit DictusApp from the app switcher, and **keep the keyboard on screen** without touching it.
*Screen:* after ~4-5 s the recording overlay leaves and « Enregistrement interrompu. Réessayez. » appears in red on the toolbar for 3 s.
*Log, expected:*
```
dictationStateReconciled source=keyboard-watchdog staleStatus=recording heartbeatAgeMs=~4000
dictationMessageSet reason=reconciled-keyboard-watchdog owner=<id> visible=true
dictationMessageDisplayed rootView=<id> controller=<id> owner=<same id> visible=true
dictationMessageCleared reason=reconciled-timeout displayedCount=1   (or more)
```
*This is the shape that should work.* The failure to look for is `displayedCount=0`, or a `Displayed` line whose `controller` differs from `owner`.

**2 — Reconciliation at appear time.** Force-quit DictusApp mid-recording, then **dismiss the keyboard** (tap outside the field, or switch to another app) and bring it back.
*Screen:* the keyboard comes back in its normal typing layout. **Does the red message appear at all?** This is the case the maintainer suspects, so the answer here is the deliverable.
*Log, expected:*
```
keyboardDidAppear
registerControllerAppearance controllerID=<A> previousOwner=none wasVisible=false
dictationStateReconciled source=keyboard-refresh staleStatus=recording heartbeatAgeMs=<large>
dictationMessageSet reason=reconciled-keyboard-refresh owner=<?> visible=<?>
```
Then the discriminating question: **is there a `dictationMessageDisplayed` after it, and does its `controller` match the `owner` on the `Set` line?**
- `displayedCount=0` on the cleared line → the message was set into a tree that never rendered it. That is the bug, and the `owner=` and `visible=` fields on the `Set` line say why.
- `displayedCount≥1` with `controller` = `owner` → it did render, and the reason the maintainer did not see it is below this layer (a screen recording is then the next step, not another log).
- `displayedCount≥1` with `controller` ≠ `owner` → it rendered in the wrong tree. That is the hypothesis confirmed.

**3 — Keyboard reload immediately after a kill.** Force-quit DictusApp mid-recording, then switch keyboards with the globe key and come back to Dictus (this rebuilds the extension rather than just re-presenting it).
*Screen:* note whether the message appears, and for how long.
*Log:* same shape as step 2, but expect a fresh `KeyboardState … action=init` before the reconciliation. If a `Set` lands with `visible=false` and no `Displayed` follows, the message was raised before any view existed to show it — a timing answer rather than an ownership one.

**4 — Two kills in a row.** Repeat step 1, then immediately tap the mic again for a cold start, let it complete, then force-quit mid-recording a second time.
*Screen:* the second message should behave like the first.
*Log:* two complete `Set` → `Displayed` → `Cleared` groups. Watch for `reason=replaced` on a cleared line, which would mean a second message arrived while the first was still up — the closest thing in the trail to the "contradictory content" report.

**5 — App error, for contrast.** Trigger an ordinary dictation failure (airplane mode with a Whisper model selected is the reliable one), so the message comes from the app rather than from the reconciliation.
*Screen:* the red toolbar message, as before this PR.
*Log:* `dictationMessageSet reason=appError` → `Displayed` → `Cleared reason=appError-timeout`. This is the control: if the app-error message renders reliably and the reconciliation message does not, the difference is the moment they are raised, not the message plumbing.

---

Send the log back whole. Steps 1 and 2 side by side are what settle it: same message, same code path, two different moments in the controller lifecycle.

https://claude.ai/code/session_016AL9tgVzFmZ1nY6S5Ko3bN

Issue #311, opened 2026-08-04
Swiping back before DictusApp becomes active strands the dictation on "Démarrage…" until the user returns by hand


Pivii:
## Summary

A cold-start dictation is parked in `pendingColdStartDictation` when the URL launch lands while DictusApp is still `.inactive`, and the only code that ever consumes that flag runs in `didBecomeActive`. If the user swipes back to the host app before the app reaches `.active`, `didBecomeActive` never fires. The dictation never starts, the keyboard overlay sits on "Démarrage…" indefinitely, the Dynamic Island stays in standby, and nothing times out or tells the user. The only recovery is to open DictusApp by hand, which fires `didBecomeActive` and starts the recording several seconds late.

Reproduced deliberately: swipe back to the host app as fast as possible after tapping the mic.

## Evidence

Device log, build `1.8.0 (25)` at `adcf8c7@develop`, iPhone / iOS 26.5.2, window 12:36:57–12:39:31Z, Parakeet. Twelve consecutive dictations, every one of them deliberately fast. **Four stranded.**

| session | `dictationDeferred` | app backgrounds before ever becoming active | `requested → recording` | stranded for |
| --- | --- | --- | --- | --- |
| `039AFF5F` | 12:37:12 | 12:37:14 | 12:37:20 | 8 s |
| `E057BAA9` | 12:37:39 | 12:37:40 | 12:37:46 | 7 s |
| `C67B4FB4` | 12:38:06 | 12:38:06 | 12:38:11 | 5 s |
| `4AC2CD53` | 12:38:47 | 12:38:49 | 12:39:09 | **22 s** |

In the eight healthy runs `appDidBecomeActive` arrives within a second of the deferral, before the swipe-back lands, and the dictation starts normally. The only difference in the four failures is that the swipe-back won the race.

The 22 s case in full:

```
12:38:47  <APP> dictationDeferred reason=cold start deferred to didBecomeActive, appState=1
12:38:48  92FF6573  viewWillAppear_entry  status=requested
12:38:48  F22CF6F7  showsOverlayChanged   isShowing=true   <- overlay up, showing "Démarrage…"
12:38:49  <APP> appDidEnterBackground                      <- app never reached .active
12:38:49  <APP> liveActivityStarted
        ... 20 seconds of nothing ...
12:39:09  <APP> urlComponents host=open                    <- user opens DictusApp by hand
12:39:09  <APP> coldStartRetry keyboardStatus=requested
[...]

Issue #73, opened 2026-04-03
Cold start dictation fails with AUIOClient_StartIO error (2003329396)


Pivii:
## Description

When the keyboard triggers dictation during a cold start (app was killed by iOS), the AVAudioEngine fails to start with `AUIOClient_StartIO error 2003329396`. The dictation fails immediately even though Parakeet finishes loading ~28 seconds later.

## Reproduction

1. Force-kill Dictus from the app switcher
2. Open Notes, switch to Dictus keyboard
3. Tap the mic button → keyboard sends Darwin notification, then URL fallback
4. App launches cold, receives URL while `.inactive`
5. Deferral logic correctly waits for `didBecomeActive`
6. `coldStartRetry` fires → `startDictation(fromURL: true)` → `audioEngine.startRecording()`
7. **CRASH:** `AUIOClient_StartIO failed (2003329396)` → `dictationFailed`

The user must manually retry (second tap works because engine is now warm).

## Xcode Logs (captured 2026-04-03)

```
appLaunched version=1.3
Initializing ParakeetEngine for model: parakeet-tdt-0.6b-v3
engineStateSnapshot engineRunning=false isRecording=false hasWhisperKit=false
coldStartURLReceived isColdStart=true isEngineDead=false hasBeenActive=false
Cold/engine-dead start from keyboard (first launch) — showing swipe-back overlay
coldStartFlagSet active=true context=first launch
dictationStarted fromURL=true appState=1 engineRunning=false
dictationDeferred reason=cold start deferred to didBecomeActive, appState=1

[didBecomeActive fires]
engineStateSnapshot engineRunning=false isRecording=false hasWhisperKit=false
coldStartRetry keyboardStatus=requested
dictationStarted fromURL=true appState=0 engineRunning=false
appDidBecomeActive
AURemoteIO.cpp:1710  AUIOClient_StartIO failed (2003329396)
AVAEInternal.h:104   [AVAudioEngineGraph.mm:1549:Start: (err = PerformCommand(*ioNode, kAUStartIO, NULL, 0)): error 2003329396
dictationFailed error=Cold start engine load: L'opération n'a pas pu s'achever. (com.apple.coreaudio.avfaudio erreur 2003329396.)

[~28 seconds later]
Compiled model Encoder.mlmodelc in 24840.48 ms
ParakeetEngine ready for model: parakeet-tdt-0.6b-v3
```

### The Race Condition

[...]

PR #312, opened 2026-08-04
fix(app): never leave a cold-start dictation parked with no way out (refs #311)

Pivii:
Fixes the stranded cold start in #311: a URL-launched dictation parked while the app is not yet `.active` had exactly one consumer, the `didBecomeActive` observer. A user who swipes back to the host app before the app settles never produces that event, so the request stayed parked forever. The keyboard held "Démarrage…", the Dynamic Island sat in standby, nothing timed out, and the only recovery was opening DictusApp by hand, which started the recording up to 22 s late.

`refs #311`

## Read this first: part 1 of the brief does not hold up

The brief asks to retry the parked start from `willEnterForeground` "or the equivalent scenePhase transition", on the premise that it fires after the parking and before `.active`.

The issue's own evidence disproves that. The deferral logs `appState=1`, i.e. `UIApplication.applicationState == .inactive` when the URL was handled, so the background-to-foreground transition had **already completed**. `willEnterForeground` — and SwiftUI's `scenePhase == .inactive`, which this app logs as `appWillResignActive` — is posted during that transition, therefore strictly *before* `pendingColdStartDictation` is set. An observer for it can never see the flag set in any of the four captured sessions. After the parking, the only lifecycle event that arrives is `.background`.

A timer was considered and rejected. Fire early and it pre-empts the healthy `didBecomeActive` path (which lands within a second in 8 of the 12 runs) by attempting a start from `.inactive`, the exact state #73 proved fails, turning healthy runs into failures. Fire late and the background transition (0-2 s in the four captures) wins anyway.

So part 1's intent is implemented at the only point where it can still fire: **one last-chance attempt at the background transition, immediately before reporting the failure.** It runs only on a path that strands 100% of the time today, so it cannot regress a working run; a failed `engine.start()` is already a handled state (`startEngine()` removes the stale tap left by exactly that failure); and `coldStartStranded ... action=retry` followed by either `audioEngineStarted` or `dictationFailed` will tell us from the next device log whether a non-`.active` start works, instead of us guessing.

Second, smaller correction: the brief says "the keyboard already has a message channel from #261 (`dictationMessage*`)". Those are log events. `statusMessage` is an in-process `@Published` on the keyboard's `KeyboardState` and the app cannot write it. The app's actual channel is `SharedKeys.lastError` plus status `.failed`, which `KeyboardState.refreshFromDefaults` picks up, shows for 3 s and clears — and `.failed` does not own the keyboard area, so the overlay drops on the same write. `handleError` already does all of it, and that is what part 2 goes through. The outcome the brief asked for is unchanged.

## What changed

**`DictusCore/Sources/DictusCore/ColdStartResolutionPolicy.swift`** (new) — the rule that decides what a parked start gets when its last chance arrives: `retry`, `report`, `dropped`, or `none`. It lives here for the same reason `IdleReleasePolicy` does: `DictationCoordinator` is `@MainActor`, owns a live `AVAudioEngine` and an `AVAudioSession`, reads `UIApplication.applicationState`, and can only be exercised on a device. It also owns `canStartNewDictation(from:)`, which is the entry guard `startDictation` used to spell out inline, so the two cannot drift.

**`LogEvent.coldStartStranded(keyboardStatus:action:)`** — the line that makes this self-diagnosing. Before it, a stranded dictation left nothing behind but a `dictationDeferred` followed by silence. `.lifecycle`, `.warning`.

**`DictationCoordinator.resolvePendingColdStartOnBackground() -> Bool`** — the guaranteed exit. Clears the flag, logs the resolution, then either attempts the start (`allowInactiveStart: true`, which is the only thing that suppresses the defer) or reports the failure through `handleError`. Not `@discardableResult`: a caller that ignores the answer would clear `coldStartActive` and start a standby activity against a recording that is beginning.

**`DictusApp.swift`** — the `.background` scene phase calls it, and ORs the result into `isRecordingActive`. Called from that handler rather than from a `didEnterBackground` observer of its own so the recovery and the `appDidEnterBackground` line the regression grep pivots on come from one event and cannot drift apart.

**The cold-start catch** now substitutes the localized message when the failure came from a last-chance start. The #73 `AUIOClient_StartIO` failure reaches `localizedDescription` as a bare CoreAudio error number; the raw text stays in the log line above it, and the banner gets the one action that works.

**New string**, EN + FR: "Dictation could not start. Tap the microphone again." / "La dictée n'a pas pu démarrer. Touchez à nouveau le micro."

App-side only. Nothing in `DictusKeyboard/` moved, so no keyboard device smoke gate.

## The two constraints on the last-chance start

**A success must be a complete start, not a half one.** Satisfied by construction, and the reason is now pinned in the code. `resolvePendingColdStartOnBackground` calls `startDictation` rather than starting the engine itself, and `allowInactiveStart` suppresses the defer and nothing else. Lines 483-488 of `DictationCoordinator.swift` are untouched on this path, so a successful last-chance start still does all of:

- `audioEngine.startRecording()` — engine up, capture on
- `updateStatus(.recording)` — writes `SharedKeys.dictationStatus` and posts `statusChanged`, which is what flips the keyboard off "Démarrage…" onto the recording overlay by itself
- `LiveActivityManager.transitionToRecording()` — and it auto-bootstraps a standby activity when none exists, so skipping `startStandbyActivity()` at the call site cannot leave the Dynamic Island empty
- `PersistentLog.log(.audioEngineStarted)`
- `verifyAudioFlow()` — the zombie-engine check, which force-restarts and then errors if no samples arrive within 2 s
- waveform energy streaming to the App Group directly from the audio thread

Recording on into the background is the warm path's normal behaviour under `UIBackgroundModes: audio`, so a success here is genuinely usable rather than a special case. The reuse is load-bearing: a future refactor that inlined an engine start here would silently reintroduce a half start, which is why it is now written down.

**The failure message.** No em dashes, short, actionable, and it never surfaces a CoreAudio string — the cold-start catch substitutes it for `error.localizedDescription` whenever the failure came from a last-chance start, and the raw text stays in the `dictationFailed` log line above.

- FR (what the user actually sees): "La dictée n'a pas pu démarrer. Touchez à nouveau le micro."
- EN: "Dictation could not start. Tap the microphone again."

On "French first": the user-visible result is French first, but the storage convention is the other way round and I kept it. Both catalogues declare `sourceLanguage: en`, and all 290 keys across them are English keys carrying an `fr` translation — zero exceptions, verified. A French key would be read as the source string against `sourceLanguage: en` and would show French on an English device. So the string follows the existing 290, and a French device shows French.

**Pre-existing, not touched, worth knowing:** two other `handleError` call sites reachable from any cold start pass unlocalized literals — `"Microphone permission denied"` (English on a French device) and `"Micro indisponible. Relancez l'application."` (French on an English one). Neither is introduced here and neither is in #311's scope, so they stayed out of the diff.

## Background execution assertion (review finding)

The review was right, and the mechanism is worse than "probably fine". `resolvePendingColdStartOnBackground` runs from the `.background` transition and `startDictation` does its real work inside a `Task`. Nothing held the process up in between, and `UIBackgroundModes: audio` structurally cannot: the runtime it grants begins once the engine is capturing, which is the far side of the very gap that needed covering. The only thing protecting that window was iOS's default post-background grace, which is a behaviour, not a contract.

So the resolution now holds a `UIApplication` background assertion — the first in this codebase, `grep beginBackgroundTask` was empty. It is released on both outcomes at the earliest moment each is safe:

| exit | released by | reason string |
| --- | --- | --- |
| engine started, audio mode takes over | in the cold-start task, right after `audioEngineStarted` | `engine-started` |
| failure reported to the keyboard | in the cold-start task's catch, after `handleError` | `failure-reported` |
| permission denial, abandoned session | the task's `defer` | `task-exit` |
| `startDictation` failed synchronously, so no task exists (no model downloaded) | detected at the call site by re-reading the App Group status | `resolved-synchronously` |
| `.dropped` / `.report`, both synchronous | at the end of the branch | `dropped` / `reported` |
| assertion expired, request still unresolved | the expiration handler, **after** reporting the failure | `expired-reported` |
| assertion expired, already resolved | the expiration handler | `expired-resolved` |

It is deliberately released before the model load rather than at the end of the task: that load can run for tens of seconds, the engine is already capturing by then, and sitting on an assertion it no longer needs is how an app earns an expiration it did not have to.

**The `.report` branch, checked as asked.** Every App Group write on it is synchronous — `handleError` does `set` / `synchronize` and `updateStatus` does `set` / `synchronize` / `DarwinNotificationCenter.post`, none of which hop off the current turn. The user-facing outcome was already safe without the assertion. What was *not* safe is the evidence: `PersistentLog.log` appends asynchronously on a `.utility` queue, and those lines are what the ten-iteration device test greps for. Hence `PersistentLog.flush()` (new, `writeQueue.sync {}` on a serial queue, so a full barrier) draining before the assertion is released. Without it the assertion would have protected everything except the proof that it worked.

**File split.** These additions pushed `DictationCoordinator.swift` past both `file_length` (830 > 800) and `type_body_length` (668 > 650). That is my code doing it, not a pre-existing knife-edge, so it got fixed rather than suppressed: the concern moved to `DictusApp/ColdStartResolution.swift`, next to `ColdStartLaunch.swift`. Three narrow members lose `private` because Swift scopes it per file — `pendingColdStartDictation`, `coldStartAssertion`, `reportStrandedColdStart()`. `handleError` is deliberately not among them, which is why `reportStrandedColdStart()` stayed behind with it. The new file is wired into the four `project.pbxproj` places by hand; a mistake there fails the build loudly, and the build passes.

## Verification

| check | result |
| --- | --- |
| `DictusApp`, `DictusKeyboard`, `DictusCore` build (iPhone 17 sim, `-derivedDataPath build/DerivedData`) | `** BUILD SUCCEEDED **` ×3 |
| `cd DictusCore && swift test` | 671 tests, 1 skipped, 0 failures |
| `swiftlint lint --strict` | exit 0, no output |
| no spurious resolution, and no assertion taken, on an ordinary background | app run on the simulator after the assertion landed: `appWillResignActive` → `appDidEnterBackground` → `coldStartFlagSet active=false context=background-cleanup`, with **no `coldStartStranded` and no `assertionEnded` line**. The `.none` guard returns before `beginColdStartAssertion`, verified in the running app rather than only in a test. |
| decision logic, every `DictationStatus` and every branch | `ColdStartResolutionPolicyTests`, 9 tests, including a sweep asserting that while the keyboard is still waiting, *no* coordinator status resolves to an outcome that says nothing to the user |
| log line name, fields, level, subsystem | `LogEventTests`, 2 tests; roster entry in `LogCoverageTests` |

**Not verified, and it needs a device.** The race itself. `simctl` cannot produce a swipe-back, and on this simulator `simctl openurl` never delivered `dictus://dictate` to the app at all (no `delegateSourceApp`, no `urlComponents` line) — pre-existing and orthogonal to this change, but it means the URL path was not exercised here either. The manual steps are in the recap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where voice dictations could become stuck when the app transitions to the background during initialization.
  * Stranded requests are now retried automatically when possible or reported with a clear, actionable message instead of being silently abandoned.
  * Improved background recovery to preserve recording state and prevent conflicting activity changes.

* **Tests**
  * Added comprehensive coverage for voice dictation recovery scenarios and lifecycle-related logging.


Pivii:
## Device validation: both branches exercised, and #73's open question answered

Device log, build `1.8.0 (25)` at `6d9c97e@fix/311-deferred-cold-start-stranded`, iPhone16,2 / iOS 26.5.2, window 13:40:33–13:43:18Z, Parakeet. Sixteen dictations, ten of them with the swipe-back performed as fast as possible.

**Fourteen took the healthy path** — `dictationDeferred` followed by `appDidBecomeActive` and `coldStartRetry` within a second, exactly as before this change. No regression on the path that already worked.

**Two stranded, and each got a different exit. Both are the ones this PR added.**

### The retry succeeded, from the background (13:42:08)

```
13:42:08  appDidEnterBackground
13:42:08  coldStartStranded  keyboardStatus=requested action=retry
13:42:08  dictationStarted   fromURL=true appState=2 engineRunning=false
13:42:08  audioSessionConfigured category=playAndRecord
13:42:08  audioEngineStarted
13:42:08  statusChanged from=idle to=recording source=coordinator
13:42:08  liveActivityStarted / liveActivityTransition from=standby to=recording
13:42:08  assertionEnded reason=engine-started
13:42:08  <KBD> hostingSet_recording old=52.0 new=276.0 / showsOverlayChanged isShowing=true
```

`appState=2` is `.background`. The engine started anyway, the Live Activity bootstrapped itself, and the keyboard flipped off "Démarrage…" onto the recording overlay on its own. This is the complete-start guarantee holding in the field, on a dictation that would have hung indefinitely before this change.

### The retry failed, and the user was told (13:42:16)

```
13:42:16  appDidEnterBackground
13:42:16  coldStartStranded  keyboardStatus=requested action=retry
13:42:16  dictationStarted   fromURL=true appState=2 engineRunning=false
13:42:16  dictationFailed    error=Cold start engine load: ... (com.apple.coreaudio.avfaudio erreur 2003329396.)
13:42:16  statusChanged from=idle to=failed source=coordinator
13:42:16  assertionEnded reason=failure-reported
13:42:16  <KBD> overlayHidden status=failed
13:42:16  <KBD> dictationMessageSet reason=appError owner=001073F8 visible=true
13:42:16  <KBD> dictationMessageDisplayed rootView=F3D8E80D controller=001073F8 owner=001073F8
13:42:20  <KBD> dictationMessageCleared reason=appError-timeout displayedCount=1
```

Overlay dropped, banner shown by the owning controller, cleared after its 3 s. The raw CoreAudio number stayed in the log line and never reached the screen. `displayedCount=1`, so exactly one tree rendered it.

### What this settles about #73

`AUIOClient_StartIO` (2003329396) from a non-`.active` state is **not deterministic**. Two attempts eight seconds apart on the same device, same session: one started the engine cleanly, one failed. So the `.active` defer must stay as the primary path — it is the reliable one — but attempting anyway when `.active` is no longer coming is worth roughly a coin flip, and the losing half now ends in a message instead of silence.

That is the empirical answer the log line was added to obtain, and it could not have been reasoned to from the code.

### Coverage

| | |
| --- | --- |
| dictations | 16 |
| stranded and resolved | 2 (1 recording started, 1 failure reported) |
| stuck on "Démarrage…" | **0** |
| transcriptions inserted | 15 of 15 that reached recording |
| `reclaimNudge` / dead windows (#281) | 0 — this issue did not occur in the capture |

Validated. Ready to merge as far as this change is concerned.

Issue #350, opened 2026-08-13
The lifetime plan decided in #54 does not exist in the app: the paywall ships two plans, not three


Pivii:
> *This was generated by AI during triage.*

## Summary

Pricing was locked on 2026-08-04 (#54, product table in #215) as **three plans**: monthly 4,99 €, yearly 39,99 € with a 7-day trial, and a **lifetime** non-consumable at 49,99 € during a founder window, 79,99 € after. The app knows about two. Nothing open tracked the gap, so #215 can be completed in App Store Connect and the third product would still be unreachable from the app.

Found during the triage pass of 2026-08-13.

## What is missing in code

`DictusApp/Subscription/SubscriptionManager.swift:31-37`

```swift
static let monthlyProductID = "solutions.pivi.dictus.pro.monthly"
static let yearlyProductID  = "solutions.pivi.dictus.pro.yearly"

private let productIDs: Set<String> = [
    SubscriptionManager.monthlyProductID,
    SubscriptionManager.yearlyProductID
]
```

`solutions.pivi.dictus.pro.lifetime` is absent, so `loadProducts()` never requests it and `Product.products(for:)` never returns it. The plan selector renders whatever came back — two rows.

`DictusApp/Subscription/StoreKitConfig.storekit` has two entries as well, both subscriptions. There is no non-consumable, so the lifetime cannot be exercised in local StoreKit testing at all.

## The CTA will mislabel it

`DictusApp/Views/PaywallView.swift:346-357` — `ctaLabel` has two branches: yearly gets `/year`, **everything else falls through to `/month`**.

```swift
if product.id == SubscriptionManager.yearlyProductID { ... /year ... }
return Text("Subscribe for \(product.displayPrice)/month")
```

Add the lifetime to `productIDs` alone and the CTA reads *"Subscribe for 49,99 €/month"* for a one-off purchase. The fall-through is the bug, not the missing branch: the label should be derived from whether the product has a `subscription` and its period, not from an ID equality chain.

Same shape in the plan rows and in any `/year` or `/month` suffix on the selector.

## What already works

`updateProStatus()` (`SubscriptionManager.swift:162`) iterates `Transaction.currentEntitlements` without filtering on product type, so a verified, non-revoked **non-consumable already grants Pro**. The entitlement side needs no change. Restore also goes through the same path.

This is worth stating explicitly because it is the part that would normally be hard, and it is done.

## Copy required before the first sale

[...]

Issue #279, opened 2026-08-01
Re-enable Dictus Pro: how to unhide the paywall and every entry point


Pivii:
## Why this issue exists

The paywall and every route to it are hidden behind a single compile-time flag, `PremiumFlags.paywallVisible`, currently `false`. #236 introduced that flag and is closed, so nothing open records **how to turn it back on**. This issue is that record. It stays open until Dictus Pro actually ships.

Nothing is deleted or commented out. `PaywallView`, `ProBannerView`, `SubscriptionManager` and all entitlement checks are compiled and intact. The flag only decides whether a user can ever see or reach them.

## Where the flag lives

`DictusCore/Sources/DictusCore/PremiumFlags.swift`

```swift
public enum PremiumFlags {
    public static let paywallVisible = false
}
```

It sits in DictusCore rather than beside `SubscriptionManager` because entry points exist in **two targets**. It began as an `internal` constant in DictusApp, which DictusKeyboard cannot see — and that is exactly how the keyboard panel (#241) shipped a Pro pill that ignored the gate entirely. Anything Pro-facing added in either target must read this flag.

## Every gated entry point

| Site | What is hidden |
| --- | --- |
| `DictusApp/Views/HomeView.swift:46` | `ProBannerView` on the home screen |
| `DictusApp/Views/SettingsView.swift:85` | The "Dictus Pro" section in Settings (crown icon, first section) |
| `DictusApp/Views/SettingsView.swift:233` | The "Pro Features" section: locked rows, PRO pills, and the per-feature toggles |
| `DictusKeyboard/Views/ToolbarView.swift:126` | The "Dictus Pro" pill in the keyboard panel bar (#241) |

With the flag `false` the product looks like it has no subscription at all. With the panel bar specifically: close on the left, gear on the right, nothing between them.

## Preconditions before flipping it

Flipping this to `true` on its own ships a paywall with empty prices and a loading error, which is the failure #236 existed to prevent. All three must be true first:

1. **A real Pro feature exists.** Smart Modes (#79) is the intended first one. Without it the paywall sells nothing.
2. **App Store Connect setup is done** (#215): Paid Apps agreement signed, the subscription product created and in a state StoreKit can fetch. Until the product exists, `SubscriptionManager` cannot load prices in sandbox either.
[...]

Issue #216, opened 2026-07-24
Rework the paywall's subscribed state into a Pro hub


Pivii:
## Context

Since #78, the paywall handles the subscribed state by hiding the plan selector and CTA and showing a "Dictus Pro Active" card. Device feedback from the maintainer: the page feels unbalanced (large empty bottom half) and the active-status card carries little meaning on its own — see the screenshot discussion in the #78 session (2026-07-24).

## Idea to explore

Once subscribed, this screen could stop being a sales page and become the **Pro hub**:

- Feature cards become interactive: manage Smart Mode / History / Vocabulary toggles directly here (today they live in Settings > Pro Features; decide whether that section moves here entirely or stays as a mirror).
- Show subscription info: current plan (monthly/yearly), renewal date, and a "Manage subscription" entry (StoreKit `manageSubscriptionsSheet` or the App Store subscriptions deep link).
- Keep Terms/Privacy links (Apple requirement) and the brand hero.

## Open questions

- Does Settings > Pro Features section survive, redirect here, or disappear?
- Entry point naming: the Settings row currently says "Dictus Pro" for both states — fine, but the pushed screen should adapt its content by subscription state.
- Free users keep the current paywall unchanged (validated design, do not touch).

## Related

- #78 — Paywall UI (validated; this is the follow-up for the subscribed state)
- #54 — Premium strategy

Pivii:
## Decided 2026-08-24: this waits for at least two Pro features, and it is not a blocker

### Why it waits

The hub's content **is** the Pro feature list. Today there is exactly one Pro feature (Smart Modes, #79). Building the hub now means building it around a single card, then rebuilding it when History (#70) lands, then again when Custom vocabulary (#80) does. Three passes for one screen.

Waiting costs nothing, because nothing depends on it — see below.

**Do it when there are two or three Pro features to put in it.** Roughly half a day at that point.

### It is no longer a precondition for #279

#279 listed this issue as one of three things to settle before flipping `PremiumFlags.paywallVisible`, on the grounds that "a subscriber who reaches the paywall again should land somewhere sensible rather than on a purchase screen".

**That bar was already cleared by #78.** The paywall hides the plan selector in the subscribed state and shows a "Dictus Pro Active" card. What remains in this issue is that the screen looks unbalanced and the card carries little meaning — real, but cosmetic. #279's body has been corrected to say so.

### One piece is extracted to #215

[...]

Issue #215, opened 2026-07-24
App Store Connect setup for Dictus Pro subscription + TestFlight validation


Pivii:
Follow-up to #55 (merged in PR #214). The subscription infrastructure is on develop with Pro gating driven purely by StoreKit entitlements. Local StoreKit testing is fully validated (purchase, cancel, refund, delete, price increase consent, restore-link visibility). What remains is the real-store side, which is maintainer work in App Store Connect plus one TestFlight validation pass.

## The catalogue to create

Prices decided in #54, lifetime revised 2026-08-14 and again 2026-08-27 (PR #440). Product IDs must match `ProProductID` exactly, character for character, or the paywall loads nothing.

| Product ID | Type | Price | Notes |
|---|---|---|---|
| `solutions.pivi.dictus.pro.monthly` | Auto-renewable, group "Dictus Pro" | **4,99 €/month** | No introductory offer |
| `solutions.pivi.dictus.pro.yearly` | Auto-renewable, group "Dictus Pro" | **39,99 €/year** | **7-day free trial** (introductory offer), yearly only |
| `solutions.pivi.dictus.pro.lifetime` | **Non-consumable** | **79,99 €** during the founder window, **149,99 €** after | A non-consumable can never carry a trial or an introductory offer |

The founder window runs one month from the App Store release that ships the first real Pro feature (#79). Moving the price at the end of the window is a manual App Store Connect edit, and it has to happen on the same day `PremiumFlags.lifetimeFounderOfferEnd` says it does. Nothing in the app enforces it.

The lifetime scope sentence must appear in the product description before the first sale:

> The lifetime purchase covers every current and future Pro feature that runs on your device or on a server you provide. Any feature relying on Dictus infrastructure is a separate offering.

## App Store Connect tasks (maintainer)

- [x] Activate the Paid Applications agreement (Agreements, Tax, and Banking: banking + tax info). Nothing loads in sandbox/TestFlight without it. Bank validation can take days, start here.
- [x] Enroll in the App Store Small Business Program (approved 2026-08-25). Commission on paid apps and in-app purchases is 15% instead of 30%, for as long as yearly proceeds stay under 1M USD.
- [ ] Create the subscription group "Dictus Pro" and the monthly auto-renewable subscription, `solutions.pivi.dictus.pro.monthly`, 4,99 €/month. "Ready to Submit" is enough for sandbox/TestFlight.
[...]

PR #355, opened 2026-08-13
Add the lifetime plan the app never knew about (refs #350)

Pivii:
## What this was for

Pricing was locked on 2026-08-04 (#54) as **three** plans — monthly 4,99 €, yearly 39,99 € with a 7-day trial, and a **lifetime** non-consumable at 49,99 € during a founder window, 79,99 € afterwards. The app only ever knew two. `solutions.pivi.dictus.pro.lifetime` was absent from the identifiers `SubscriptionManager` requests, so the product could be created in App Store Connect (#215) and still be unreachable from the app. Nothing open tracked that gap.

This is the app-side half of that decision. It does not make anything visible: `PremiumFlags.paywallVisible` stays `false`, and flipping it is still #279.

## To test by hand

Nothing on this screen can be verified headlessly. The simulator has no touch injection, and the paywall is unreachable while the flag is `false`. Everything below needs a human.

**Step 0 — make the paywall reachable, locally only.** Set `PremiumFlags.paywallVisible = true` in `DictusCore/Sources/DictusCore/PremiumFlags.swift`. **Do not commit it.** Run from Xcode with the `DictusApp` scheme (it attaches `StoreKitConfig.storekit`; a `simctl` install does not). Reach the paywall from Settings → Dictus Pro.

1. **Open `DictusApp/Subscription/StoreKitConfig.storekit` in Xcode.** Expect: a "Dictus Pro Lifetime" non-consumable at 49.99 alongside the two subscriptions, and Xcode raising no parse complaint. *(The JSON schema is undocumented; this is the one thing written by pattern-matching. StoreKit does resolve it at runtime — see the evidence below — but the editor is the last word.)*
2. **Look at the plan selector.** Expect three rows in this order: Yearly (preselected, discount badge, "Soit 3,33 € par mois"), Monthly, Lifetime. Yearly still selected on first frame.
3. **Read the lifetime row.** Expect `49,99 €` with **no** `/mois` and no `/an` suffix. This was the bug: the old CTA fell through to `/month` for anything that was not the yearly ID.
4. **Read the sentence under the lifetime row.** Expect, in French: « L'achat à vie couvre toutes les fonctionnalités Pro, actuelles et futures, qui s'exécutent sur votre appareil ou sur un serveur que vous fournissez. Toute fonctionnalité reposant sur l'infrastructure Dictus relève d'une offre distincte. » Check it wraps fully and is not truncated on a 6.1-inch screen.
5. **Confirm there is no founder line.** Expect nothing between the price and the scope sentence: no date, no "bientôt", no empty placeholder.
6. **Tap the lifetime row.** Expect the CTA to read « Acheter pour 49,99 € » (not "S'abonner"), and the fine print under it to read « Achat unique » instead of « Annulable à tout moment ».
7. **Tap back to Yearly, then Monthly.** Expect « 7 jours gratuits, puis 39,99 €/an » and « S'abonner pour 4,99 €/mois ». *(This is the regression risk: the CTA sentence and the period suffix are now two separate localized strings composed together.)*
8. **Relaunch in English** — `xcrun simctl launch <udid> com.pivi.dictus -AppleLanguages '(en)'`, or the scheme's App Language. Expect "Lifetime", "Buy for 49,99 €", "One-time purchase", and the English scope sentence.
9. **Buy the lifetime** from the StoreKit test sheet. Expect the thank-you screen, then Pro active everywhere (Settings crown checkmark, Pro toggles unlocked).
10. **Reopen the paywall while owning the lifetime.** Expect the "Dictus Pro Active" banner and **no plan selector and no CTA** — no second purchase is offered. *(#216 still owns making that banner say "Lifetime" rather than a generic active state.)*
11. **Delete the app, reinstall, tap "Restore purchases".** Expect Pro to come back. For a non-consumable this is the entire product promise. On-device StoreKit test transactions are wiped with the app, so this specific check is only conclusive in sandbox/TestFlight — see #215 Phase 3.
12. **Temporarily set the founder date to a day in the future** to check the copy renders: `public static let lifetimeFounderOfferEnd: Date? = DateComponents(calendar: .current, year: 2026, month: 9, day: 12).date`. Expect « Offre fondateur jusqu'au 12 septembre 2026. » under the price, and "Founder offer until September 12, 2026." in English — the deadline only, **no amount**, and no mention of a later price.
13. **Now set that same constant to a day already past** (`year: 2026, month: 8, day: 1`). Expect the founder line to **disappear entirely**, leaving the lifetime row showing its price and the scope sentence only. This is the fix that stops a shipped build from still advertising a founder deadline weeks after the price rose in App Store Connect — the failure is invisible until the date passes, so it is worth seeing once by hand. **Then revert the constant to `nil`** — a test asserts it ships unset.
14. **Optional, the riskiest UX change: delete the `solutions.pivi.dictus.pro.yearly` entry from `StoreKitConfig.storekit`** and reopen the paywall. Expect two rows, **Monthly checked**, and the CTA reading « S'abonner pour 4,99 €/mois ». Before this PR's fix the rows were all unchecked while the CTA charged the monthly anyway. Restore the file afterwards.
15. **Revert step 0.**

## What changed

Four commits for the original issue, then five for the review round.

**`ProProductID` in DictusCore.** The identifiers were declared in DictusApp, which has no test target. App Store Connect identifiers are permanent — a typo cannot be renamed, only abandoned — so they now sit where the DictusCore suite can check them against the StoreKit configuration the app ships.

**The lifetime product is requested.** Identifier added to the set, `lifetimeProduct` accessor, non-consumable added to `StoreKitConfig.storekit` so it can be exercised locally at all. The entitlement path is untouched: `updateProStatus()` scans `Transaction.currentEntitlements` without filtering on product type, so a verified non-consumable already grants Pro, and restore lands on the same path.

**The period is derived, not matched.** The period belongs to the price, not to the sentence, so `priceLabel(for:)` produces "39,99 €/an" or a bare "49,99 €" from `Product.subscription`, and the CTA wraps that. `Subscribe for %@/month` and `Subscribe for %@/year` collapse into `Subscribe for %@`; there is no branch left to fall through. A product with no subscription reads "Buy for 49,99 €".

**The lifetime row and its copy.** Third row, yearly still preselected. The scope sentence always renders — a promise cannot be narrowed after someone has bought against it. The founder line renders only when `PremiumFlags.lifetimeFounderOfferEnd` holds a date, and it ships `nil`.

## Review round

**The founder line never expired.** It was gated on the constant being non-nil, never on the date still being ahead. Builds outlive the window: App Store Connect moves the price to 79,99 € on the announced day, but a user who has not updated keeps running the build that announced the offer and is still promised a founder price that is gone, on a date already past. `LifetimeFounderWindow` now owns that boundary. The window covers the whole of the named day, and since the constant's recipe yields the *start* of that day, the comparison runs against the start of the day after it — through the calendar, so a daylight-saving day still ends when the calendar says it does. The constant's doc comment now says nothing may read it raw.

**A partially missing catalogue was silent.** `loadProducts()` logged only the all-or-nothing case, so if the lifetime alone were mistyped, still Waiting for Review, or not cleared for a storefront, the two subscriptions would return, the lifetime row would simply not render, and nothing would be logged. That is this issue's own failure inverted. The partial case now names the absent identifiers.

**The CTA could charge a plan no row showed as selected.** `selectedProduct` degraded to `products.first`, which is the cheapest, so if the yearly alone failed to load every row drew an empty circle while the CTA read "S'abonner pour 4,99 €/mois" and charging it did exactly that. I reconciled rather than dropped the fallback: dropping it leaves a paywall where two plans loaded fine but nothing is selectable and the CTA sits dead on "...". The selection now moves onto the cheapest loaded plan, so a failed fetch can only ever demote the user, never promote them onto the 49,99 € lifetime.

**`priceLabel` ignored the period's length.** It switched on the unit alone, so any month-unit subscription rendered "/mois" — and App Store Connect sells 2-, 3- and 6-month durations, where "39,99 €/mois" advertises a third of what it charges. The rule moved into `ProPlanPeriod` in DictusCore to be testable, over a mirror of StoreKit's unit rather than the type itself: the keyboard extension links DictusCore under a ~50 MB ceiling and must not pull StoreKit in.

**The founder line named a price it could not state correctly.** It read "Offre fondateur jusqu'au 12 septembre 2026. 79,99 € ensuite." StoreKit gives the *current* price in the buyer's own currency through `displayPrice`, but the future price is not a product yet, so it could only be written into the localized string — a US buyer would read $54.99 on the row and "79,99 € afterwards" underneath, in a currency they will never be charged in. The amount on a non-euro storefront is not ours to state either: App Store Connect derives it from Apple's price matrix and readjusts it as rates move. The line now carries the deadline only. Naming the future price is more persuasive and it was the approved wording; it goes because it is only correct on the euro storefronts, and for a window lasting weeks correctness is worth more. The rejected alternative is recorded in the code and stays open: a second, never-offered product holding the future price so `displayPrice` renders it everywhere — not built, because it burns a permanent product identifier for a line that shows for weeks.

## Judgement calls beyond the issue's checklist

- **"Cancel anytime" follows the selected plan** ("Achat unique" for the lifetime). Not in the issue's scope list, but it is the same bug as the `/month` fall-through: copy describing a one-off purchase as recurring.

## Acceptance criteria

| --- | --- | --- |
| `lifetimeProductID` added and in `productIDs` | Done | `swift test`: `testIdentifiersMatchTheAppStoreConnectDecision`, `testEveryIdentifierIsRequestedFromStoreKit` |
| Non-consumable in `StoreKitConfig.storekit` | Done, and resolved by StoreKit at runtime | `testLifetimeIsANonConsumableOutsideTheSubscriptionGroup`; plus the control runs below |
| Period derived, none for a non-consumable | Done | `ProPlanPeriodTests` (6 tests) covers the rule including multi-month durations; `testLifetimeIsANonConsumableOutsideTheSubscriptionGroup` pins the input. Rendering itself is manual step 3. |
| Third row, yearly stays preselected | Done | Code; `selectedProductID` default unchanged. Visual: manual step 2. |
| Founder copy + scope sentence, FR + EN | Done | Compiled `fr.lproj/Localizable.strings` in the built app carries all five strings; `LifetimeFounderWindowTests` (8 tests) covers when the founder line may show. Visual: manual steps 4, 5, 12, 13. |
| Purchase / restore / already-owns-lifetime | **Not verified** | No headless path exists. Manual steps 9, 10, 11. |

**Runtime control runs.** `Product.products(for:)` drops an unknown identifier without throwing, and `loadProducts()` logs that. Driving the configuration through the simulator's StoreKit staging directory:

| Catalogue given to StoreKit | Logged |
| --- | --- |
| empty | `error=empty result (StoreKit configuration missing or product ID unknown)` |
| **only** the lifetime non-consumable | nothing — so `Product.products(for:)` resolved it |
| the two subscriptions, lifetime absent | `error=missing product IDs: solutions.pivi.dictus.pro.lifetime` |
| all three | nothing |

The second row closes the one real risk in this change — the `.storekit` JSON schema is undocumented and was written by matching the shape of the existing subscription entries. The third and fourth are the review's silent-partial-failure finding, fixed and shown firing on exactly the missing identifier with no false positive.

## Checks

- `xcodebuild build -scheme DictusApp` → `** BUILD SUCCEEDED **`. Builds `DictusApp`, `DictusKeyboard`, `DictusWidgets` and `DictusCore`. The build emits 27 warnings, none of them in any file this PR touches — they are pre-existing, in `TranscriptionService`, `ModelManager`, `KeyboardViewController` and others. *(An earlier version of this description said "0 warnings"; that was an incremental build that had not recompiled those files. Correcting it rather than leaving the stronger claim standing.)*
- `cd DictusCore && swift test` → 942 tests, 0 failures (22 new).
- `swiftlint lint --strict` → 0 violations in 180 files.
- `Localizable.xcstrings`: 4 added entries, 3 replaced, nothing else. A build was run over it and added no churn; trailing newline intact.
- Version and build numbers untouched.

refs #350

## Summary by CodeRabbit

* **New Features**
  * Added a one-time Dictus Pro Lifetime purchase option alongside monthly and yearly plans.
  * Added dynamic pricing, plan details, subscription periods, introductory offers, and lifetime purchase messaging.
  * Added founder-offer messaging with date-aware availability.
* **Improvements**
  * Improved handling when some purchase options are temporarily unavailable, automatically selecting an available option when needed.
  * Improved recognition of lifetime purchases and Pro access.
* **Tests**
  * Added coverage for plan labeling, product catalog details, lifetime entitlements, and founder-offer timing.


Pivii (commented):
## Pricing update based on RevenueCat benchmarks (2026-08-14)

I did a quick benchmark analysis (documented separately in Second Brain). The key finding:

**RevenueCat rule of thumb**: lifetime = 3–5× annual. Real-world range: 2× (Moonly) to 11.5× (Waking Up).

Our current 49,99€ lifetime is **1.25×** the 39,99€ annual — **below every real app in RevenueCat's dataset**. Even Moonly (the cheapest) is at 2.1×.

### What needs to change in this PR

1. **`StoreKitConfig.storekit`** — `displayPrice: "49.99"` → `"79.99"` (founder price = 2× annual)
2. **`testPricesMatchThePricingDecision()`** — the assertion for lifetime 49.99 must switch to 79.99
3. **Scope sentence stays** — the "local or self-hosted" boundary is solid
4. **Founder window logic is fine** — `nil` shipped, `PremiumFlags.lifetimeFounderOfferEnd` set only at release time

### Long-term

After the founder window closes (1 month after #79 ships), App Store Connect raises the lifetime to 119,99€ (3× annual) while this code stays the same: the price comes from StoreKit, not from the app.

### Why not 49,99€

> "Your most loyal subscribers are the most likely to buy a lifetime deal, so priced too low it cannibalizes the recurring revenue they would have paid." — RevenueCat

At 49,99€, one lifetime purchase replaces only 1.25 years of annual. At 79,99€, it replaces 2 years — still aggressive but defensible. At 119,99€ (post-window), 3 years = healthy zone.

Source: https://www.revenuecat.com/blog/growth/lifetime-subscriptions

Issue #351, opened 2026-08-13
Bring your own server: let the user point Dictus at their own transcription or LLM endpoint


Pivii:
> *This was generated by AI during triage.*

**Not before Dictus Pro ships.** This is `ready-for-human` for sequencing, not for complexity — the design below is settled, but the Pro release (#215 → #350 → #79 → #279) comes first. Do not start it in that window.

## The idea

Let a user point Dictus at a machine they control — a desktop, a homelab box, a rented GPU — and have transcription, LLM polish, or both run there instead of on the phone.

Raised by the maintainer on 2026-08-13. A Dictus-operated server was considered in the same conversation and **rejected for now**; see "What is explicitly out of scope" below.

## What the gain actually is

Not speed. Parakeet transcribes in 150-400 ms on device (#309); a network round trip plus audio upload plus remote inference does not beat that on a ten-second dictation, which is the dominant case.

The gain is **LLM polish quality**. Apple Foundation Models runs roughly a 3B model on the phone, with a context ceiling being guarded in #270 and a refusal mode that corrupts output in #349. A 32B or 70B model on the user's own hardware is a different product. Remote transcription is worth offering for completeness — large-v3 beats what fits in the keyboard's memory budget, and long audio benefits — but it is the secondary half.

**So the first deliverable is a remote polish engine, not a remote ASR engine.**

## Why this is small

Two facts collapse most of the perceived scope.

**1. `PolishEngineProtocol` was built for this.** `DictusCore/Sources/DictusCore/Polish/PolishEngineProtocol.swift:9` — three members, two existing implementations (`AppleFoundationModelsPolishEngine`, `PassthroughPolishEngine`). A remote engine is a third one that speaks HTTP. `PolishCoordinator`, the guardrails, the metrics ring, the `announcesProcessingStage` behaviour and the off-device `polish-harness` all apply unchanged. #268 already records this protocol as the migration seam.

**2. The protocol is already standardised, by someone else.** `POST /v1/chat/completions` and `POST /v1/audio/transcriptions` — the OpenAI request shapes — are what every self-hosted server speaks: faster-whisper server, whisper.cpp, FunASR, Ollama, LM Studio, vLLM. Speaking those two endpoints makes Dictus compatible with the whole ecosystem on day one.

**There is no Dictus server framework to write.** The maintainer's first instinct was to build one; it is not needed, and building one would mean maintaining a project that already exists five times over.

## Shape

- Advanced settings, **off by default**, behind Pro.
[...]

Issue #315, opened 2026-08-05
Apple FM polish intermittently fails while availability is reported as available


Pivii:
## Summary

The polish layer is intermittently failing even though Apple Foundation Models is reported as available and polish is enabled. When it fails, the debug export records `outcome = engineFailed` and the app appears to insert/fall back to the raw transcription.

This was noticed during the same session where animation slowdowns were reported, but this is a separate core polish reliability issue.

## Environment

From `dictus-polish-debug-20260805-174258.json`:

```json
{
  "appVersion": "1.8.0",
  "buildNumber": "25",
  "hardware": "iPhone16,2",
  "systemName": "iOS",
  "systemVersion": "26.5.2",
  "activeModelEngine": "PK",
  "activeModelID": "parakeet-tdt-0.6b-v3",
  "appleFMAvailable": true,
  "appleFMState": "available",
  "polishEnabled": true,
  "targetLanguage": "fr"
}
```

## Observed behavior

- Polish is enabled.
- Apple FM availability is `available`.
- Many polish events succeed earlier in the export.
- Later events start failing with `outcome = engineFailed`.
- Several failures are almost immediate (`engineMs` 4–10 ms, `latencyMs` 19–24 ms), which suggests a fast rejection/throw rather than a long model run.
- Raw and polished char counts are equal on failures, but `polished` is empty in the event payload; user-visible behavior appears to be raw insertion / no polish.

Reporter quote:

> "j'ai l'impression [...] que j'ai le polish qui ne marche pas"

## Export summary

From `dictus-polish-debug-20260805-174258.json`:

```text
events: 242
success: 196
engineFailed: 25
skippedShort: 13
exceededContextBudget: 4
skipped: 2
rejectedGuardrail: 1
cancelled: 1
```

Latency distribution:

```text
success:      n=196, min=688 ms, p50=1654 ms, p90=4842 ms, max=12528 ms
engineFailed: n=25, min=14 ms,  p50=929 ms,  p90=2942 ms, max=5936 ms
```

The last 15 events show a cluster of failures after several successful calls:

```text
2026-08-05T15:27:08Z success      lat=739  engine=725  rawLen=66
2026-08-05T15:27:27Z success      lat=3180 engine=3169 rawLen=252
2026-08-05T15:28:11Z success      lat=3775 engine=3764 rawLen=579
2026-08-05T15:28:44Z success      lat=4004 engine=3992 rawLen=419
[...]

Issue #301, opened 2026-08-03
The documented DictusCore test command fails: the iOS-Simulator destination builds polish-harness, which is macOS-only


Pivii:
## Summary

The test command recorded in `CLAUDE.md` and in `docs/agents/language-onboarding.md` does not run. Every agent handed that command loses a verification step, or silently substitutes another destination.

```bash
cd DictusCore && xcodebuild test -scheme DictusCore-Package \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
```

```
DictusCore/Sources/polish-harness/main.swift:75:8: error: 'SystemLanguageModel' is only available in iOS 26.0 or newer
DictusCore/Sources/polish-harness/main.swift:88:88: error: 'AppleFoundationModelsPolishEngine' is only available in iOS 26.0 or newer
** TEST FAILED **
```

## Cause

`DictusCore-Package` builds every target in the package, `polish-harness` included. That target is a macOS command-line harness, and its Foundation Models calls are guarded for macOS but not for iOS. Building it for an iOS destination surfaces the unguarded uses as availability errors. The failure is in the harness, not in the library or the tests.

## The macOS destination works

```bash
cd DictusCore && xcodebuild test -scheme DictusCore-Package -destination 'platform=macOS,arch=arm64'
```

```
Executed 586 tests, with 1 test skipped and 0 failures (0 unexpected)
** TEST SUCCEEDED **
```

This contradicts `docs/agents/language-onboarding.md`, which states that the package cannot build for macOS because it pulls in iOS-only SwiftUI symbols. Whatever was true when that line was written, the full suite passes on macOS today.

## Verification

Both runs above were executed on `develop` at `4ea8546`, with no branch applied. This is pre-existing and unrelated to any open PR.

## Found via

PR #300 (#286). The implementing agent hit the failure, fell back to the macOS destination, and reported the substitution rather than hiding it.

## What needs deciding

Two directions, and the choice is not obvious:

1. **Guard `polish-harness` for iOS** so the documented command works as written. Keeps one command for everything, and keeps availability annotations on a target that will never run on iOS.
2. **Make macOS the documented destination** and update `CLAUDE.md` plus `docs/agents/language-onboarding.md`. Cheaper, and honest about where these tests actually run — but it needs someone to confirm no test in the suite depends on iOS-only behaviour that macOS silently stubs out.

[...]

Issue #268, opened 2026-07-30
Can a backgrounded DictusApp reach the ANE for LLM inference? (D2 — reliability half of the second-backend spike)


Pivii:
> **History.** This issue was opened on 2026-07-30 as *"is a second local LLM backend viable on pre-Apple-Intelligence iPhones?"*. It carried two rationales — market **coverage** and Apple FM **reliability**. The spike ran on 2026-08-13 (PR #356, merged), D1 landed on 2026-08-21, and the coverage rationale is now **closed on measured evidence**. See [the decision comment](https://github.com/getdictus/dictus-ios/issues/268#issuecomment-5369728789).
>
> This body has been rewritten to describe only what is still open. The original framing is preserved in the issue history and in the report.

## The question

**Can a backgrounded DictusApp reach the Neural Engine to run an LLM?**

That is the whole issue. Everything else about a second backend has been measured and answered.

## Why it is the only question left

Dictus polishes while backgrounded — the user is in Notes, not in Dictus. Apple closes the GPU to backgrounded apps, so a candidate model runs on **CPU or the ANE**. On the same model, same prompt, same machine (M4 Pro, `3-long` fixture, 1,704-token prefill):

| `qwen3:1.7b` | cold | warm |
| --- | --- | --- |
| CPU-only | **7.92 s** | 2.70 s |
| GPU / ANE class | 3.82 s | **1.37 s** |

The pre-registered budget is p50 ≤ 5 s. The Apple FM field baseline is p50 **1,654 ms**. So on CPU this is not shippable, and on the ANE it is competitive — *and never rate-limited*, which is the point.

**The ANE demonstrably works backgrounded**: FluidAudio runs Parakeet on `.cpuAndNeuralEngine` on every single dictation, specifically to avoid GPU dispatch. What is unknown is whether an **LLM** can be made ANE-resident and hold a 1,704-token prefill there.

## The candidate

`qwen3:1.7b`, Q4_K_M — the one model the spike excluded on a memory ceiling that D1 later falsified.

- **1.9 GB resident** against ~3.33 GB of measured jetsam headroom on an 8 GB device
- **66/72** on the property-check suite, above the 63/72 no-LLM control
- Keeps every jargon token on the `2-bilingue` fixture exactly as Apple FM does, and fabricates nothing on `1-normal` **[measured 2026-08-21]**

Not "viable" — **not excluded**. Those are different claims.

## Scope — this is a measurement, not an integration

[...]

PR #356, opened 2026-08-13
Spike report: is a second local LLM backend viable on pre-Apple-Intelligence iPhones? (refs #268)

Pivii:
Spike report for #268. **Measurement only** — no production integration, nothing added to `DictusCore`, `DictusApp` or `DictusKeyboard`. The deliverable is `docs/research/268-second-llm-backend.md`.

## What this was for

You are trying to find out whether an LLM feature can be made reliable at all on this architecture, because the one you have silently stops working for twelve minutes at a time. #268 started as a market-cap question — most of the installed base cannot run Apple Intelligence, so most of the installed base cannot run the flagship Pro feature — and #315 turned it into a reliability question on 2026-08-13: Apple documents that `rateLimited` only happens to backgrounded apps, and DictusApp is backgrounded for every polish call by design.

So the question this spike had to answer was not "is there a nice small model", it was "is there any local LLM that works, in the background, on the hardware Dictus actually runs on, without destabilising the dictation pipeline".

## What is still unknown, and what it would take to know it

For a spike this replaces the test list.

**D1 — needs your iPhone, costs five minutes, and is the cheapest decisive measurement left.** Peak memory and jetsam headroom during a backgrounded dictation could not be measured off-device: a simulator has no jetsam limit and runs Core ML on the Mac. But **no LLM work is needed to get the first half of the answer**, because the instrumentation already ships.

1. On the current `develop` build, do **five ordinary dictations from the keyboard** — normal length, French, polish on, default Parakeet model.
2. Export the **persistent log** (not the polish debug JSON — the persistent log, which carries `transcriptionPerformance` and `modelPrewarmPeakMemory`).
3. Send the file.

That tells us how much headroom DictusApp actually has left with Parakeet resident and the audio engine warm. If it is under roughly 1.5 GB, the whole question is closed without anyone writing integration code.

**D2 — needs a physical iPhone and about a day of work.** Whether a Gemma 3 4B QAT model converted for the Neural Engine can hold this product's 1,704-token prompt fast enough while backgrounded. Only worth doing if you decide to chase the reliability rationale rather than the coverage one.

**D3 — needs a decision from you, not a measurement.** Coverage and reliability want different models and this spike cannot choose between them. Coverage wants something that runs on an iPhone 12; reliability wants something good enough to replace Apple FM on an iPhone 17 Pro. My recommendation below picks one, but it is your call.

## Findings

**1. The GPU is not available to a backgrounded app, and that invalidates every published iPhone LLM benchmark.** Apple's Core ML reference says to restrict a model to the CPU "if your app might run in the background"; UIKit's background guidance says "Don't commit any new Metal work to be processed"; and iOS 26 added a separate, entitlement-gated capability for background GPU work, scoped to `BGContinuedProcessingTask` and "not supported on all devices". Dictus stays alive on `UIBackgroundModes: audio`, which has no route to it. The shipping product already proves the permitted half: FluidAudio runs Parakeet with `.cpuAndNeuralEngine` "to avoid GPU dispatch", backgrounded, on every dictation.

Measured cost of that restriction on the same machine and model: Metal is 2.2–2.9× faster than CPU on decode and 2.5–7× on prefill.

**2. CPU-only latency already misses the budget on an M4 Pro, for a 1B model.** With the real `PolishNaturalPromptFR` system prompt plus the largest shipping fixture — 1,704 tokens of prefill, counted by the model's own tokenizer — cold and CPU-only: `gemma3:1b` 4.79 s, `llama3.2:3b` 13.03 s. An iPhone is a smaller configuration of the same cores, so these are floors. The pre-registered budget was p50 ≤ 5 s, derived from your own field baseline (p50 1654 ms, max 12528 ms).

**3. The only model inside the memory budget makes the text worse than doing nothing.** Resident set at 4096 context, CPU-only: `gemma3:1b` 1.2 GB, `qwen3.5:0.8b` 1.4 GB, `qwen3:1.7b` 1.9 GB, `llama3.2:3b` 2.9 GB, `gemma3:4b` 3.4 GB. Only the first is at the 1.2 GB line — and it scored 60/72 on the shipping fixtures against a **no-LLM control of 63/72**. It translates the jargon the French prompt exists to protect: "je te laisse **examiner** la pull request" where Apple FM and the 4B both keep "review".

**4. The smallest Qwen fabricated content.** `qwen3.5:0.8b` turned "J'espère que tu es passé un bon week-end" into "Je te souhaite une semaine très réussie" and invented "je me sentais calme tout au long de la période". For a dictation product that is a disqualification, not a score.

**5. Quality is not the blocker at 4B; everything else is.** `gemma3:4b` scored 72/72, above Apple FM's 69/72 on the same fixtures in the same session, and reproduced its jargon handling exactly. It needs 3.4 GB resident and 4.29 s CPU-only warm on an M4 Pro.

**6. The ANE route is the only one not excluded, and it is excluded on the coverage tier by its own documentation.** ANEMLL (MIT, active) records an "M1/A14 limitation: constrained to 512-context monolithic models". A14 is the iPhone 12, and this product's prompt is 1,704 tokens.

**7. Apple FM is not a fair memory comparison.** Apple publishes it: ~3B weights at **2 bits per weight** with QAT, KV cache at 8 bits, and the model ships and updates with the OS. It costs the app no download and no app-owned weights. A challenger pays for all three, at 4–8 bpw.

## Recommendation

**Coverage rationale — no.** Measured, not argued. The 4 GB / 6 GB tier is closed. I would close that half of #268 rather than leave it as something to retry, and reopen only for a specific new model with a specific new footprint.

**Reliability rationale — not answered here, and #351 answers it better.** Same `PolishEngineProtocol` seam, already specced, a 32B on hardware the user already owns, and **zero** added resident footprint — which is the constraint that kills every option measured above.

Order I would work in: get D1 → do #315's option 3 (detect the latch and tell the user) regardless, since it is independent and unambiguously an improvement → sequence #351 ahead of any further work here → leave D2 on the shelf.

## What I am not comfortable asserting

Any iPhone latency number — the Mac figures are floors with an unknown multiplier, and I have not converted them. That `gemma3:1b` is weak at French *because* it is 1B — the model card documents 140+ languages without differentiating by size; what is measured is this checkpoint, at this quantization, with this prompt. That the property-check suite ranks models — the 63/72 control proves it calibrates against a floor, so treat 66 vs 72 as "both behaving". That Core ML would hard-fail rather than degrade if asked for the GPU in the background — Apple's wording is "allow", not "require", and I did not close that. Exact per-model iPhone RAM — Apple does not publish it, so the report gives a threshold and points at the runtime gate that already exists.

## Code added

Throwaway, confined to the `polish-harness` target (macOS-only, out of every app target, not built by CI): `LocalHTTPPolishEngine.swift`, a `PolishEngineProtocol` engine speaking `POST /v1/chat/completions`, plus `--engine local --model <name>` on `show`/`eval`. It reads the shipping instructions through `AppleFoundationModelsPolishEngine.instructions(for:language:)` so the comparison is of models, not of prompts.

One note for whoever touches the harness next: `chat_template_kwargs: {enable_thinking: false}` is accepted and **ignored** by Ollama 0.32.6 — `reasoning_effort: "none"` is what actually works. An earlier pass ran the Qwen models with thinking silently on and was discarded.

## Verification

- `swiftlint lint --strict` — 0 violations in 178 files
- `cd DictusCore && swift build` — clean
- `cd DictusCore && swift test` — 920 tests, 0 failures
- `xcodebuild -scheme DictusCore-Package -destination 'generic/platform=iOS'` — succeeds (the harness target is compiled for iOS too, which is the #301 trap; the new file carries the availability annotation that keeps it building against an iOS 17 deployment target)

No device testing applies: this PR changes no app target and ships no feature.

https://claude.ai/code/session_014epcyo5E8D3KvuXQmZnWA4

## Summary by CodeRabbit

* **New Features**
  * Added support for running polishing through a local, OpenAI-compatible language model endpoint.
  * Added command-line options to select the local engine, model, and server URL.
  * Added validation and clear error handling for invalid configuration, responses, and connection failures.

* **Documentation**
  * Added research findings on local language-model performance, memory usage, latency, quality, and device feasibility.


Pivii:
## Amended 2026-08-21 — D1 came back and it changes the conclusion on one of the two rationales

New commit `3ac8650`, one file, documentation only. The original report above the addendum is **unedited**, so the pre-registered reasoning stays auditable against what actually happened.

**What changed.** D1 was collected on the maintainer's device and it does not confirm the report. Available jetsam headroom during a backgrounded dictation is **~3.33 GB** on an 8 GB `iPhone16,2` — 2.8× the 1.2 GB ceiling Finding 3 rejected three candidates against. That ceiling is defensible for the coverage tier and was misapplied to the reliability tier.

One candidate survives the correction: **`qwen3:1.7b`**, 1.9 GB resident, 66/72 — which Finding 4 never examined qualitatively because Finding 3 had already cut it on memory. Measured on the two fixtures that disqualified its neighbours, it keeps every jargon token on `2-bilingue` exactly as `apple-fm` does, and fabricates nothing on `1-normal`.

Finding 2's blank cold cell for that model is filled: **7.92 s CPU-only vs 3.82 s Metal** cold, 2.70 s vs 1.37 s warm. Warm reproduces the original table (2.97 / 1.30), so the sessions are comparable. That makes Finding 1's **admittedly unclosed** question about background GPU/ANE access load-bearing, where it previously was not — the "What I am not comfortable asserting" entry on `MLComputeUnits` being "allow" not "require" is now the hinge of the whole file.

**Net effect on the recommendation:**

- Coverage rationale — unchanged, still no, still should be closed.
- Reliability rationale — **no longer closed**. D2 is reframed around a 1.7B model rather than a 4B one.

Decisions and the reframed scope are recorded in #268.

**On reviewing this PR:** the addendum's own "What this addendum does not assert" section is the part to read adversarially. `qwen3:1.7b` is *not excluded*, which is not the same as viable — iPhone latency unmeasured, ANE conversion unattempted, and the jetsam margin with a resident KV cache alongside Parakeet is exactly what D2 has to show.

No change to any app target. Verification from the original PR body still holds; the new commit touches one Markdown file.

Issue #346, opened 2026-08-11
A learned word never reaches the suggestion bar, and blanks it in corrections mode


Pivii:
## Summary

A word the keyboard has learned never reaches the suggestion bar. Not by any of the three modes, not under any condition. Learning a word today does exactly one thing — it stops autocorrect from rewriting it — and the user is never offered it back.

Worse than absent: in `.corrections` mode the bar goes **empty** for a learned word, because both `spellCheck` overloads funnel through `TextPredictionEngine.swift:165` where `UserDictionary.isLearned` returns `nil`. Learning a word actively degrades the bar.

Decided in the #287 grilling session of 2026-08-11 (decision 5): a learned word gets **L1, offered** in addition to **L2, immune**. **L3, authoritative** — the keyboard rewriting other words *into* a learned word — stays out and belongs to #114.

## Why this is not a threshold problem

No value of `repetitionThreshold` produces a suggestion. This is a **missing consumer**, and it is why #287's original report reads as "all of the cost and none of the benefit". The three modes and what each would need:

| Mode | Source today | Reads the learned set? |
| --- | --- | --- |
| `.completions` | `UITextChecker.completions(...)`, reranked by `FrequencyDictionary` | **No.** Dictus also never calls `UITextChecker.learnWord`, so the system checker does not know the word exists |
| `.corrections` | `spellCheck` | Yes, and it returns `nil`, so the bar shows **nothing** |
| `.predictions` | trie n-grams, falling back to `frequencyDict.topWords` | **No** |

The hook that looks like it would solve this is already there and empty. `SuggestionState.learnWord(_:)` forwards to `injectUserWord(_:)`, which is a no-op in **both** engines and says so in its own comments (`TextPredictionEngine.swift:527-530`, `AOSPTrieEngine.swift:150-153`).

## Hard dependency

**Blocked on #287 block A landing first.** Block A is what removes the trie duplicates — `le`, `chat`, `pomme`, twenty of twenty-two measured entries. Shipping this consumer before that prune fills the suggestion bar with ordinary vocabulary the user never needed offered.

## Desired behaviour

- A learned word appears in the suggestion bar as a completion of its own prefix.
- A learned word no longer blanks `.corrections`. Being learned should mean "do not rewrite this", not "have nothing to say about this".
- A learned word is never an autocorrect **target** — the keyboard does not rewrite another word into it. That is L3 and it is #114's problem.

## The trap to avoid

[...]

Issue #326, opened 2026-08-07
German Umlautersatz spellings are in the dictionary as valid words, so the collapse rules that would fix them never run


Pivii:
## Summary

German autocorrect never converts the ASCII fallback spellings `fuer`, `schoen`, `koennen` and their kin into `für`, `schön`, `können`. The rules that would do it exist and are correct. They are never reached, because the misspelling is already in the dictionary as if it were a word.

Found on device while validating #321 (2026-08-07), with autocorrect debug logs on. Not caused by #321 — the umlaut scoring that issue added works, and the same logs prove it.

## What the logs show

```
AUTOCORRECT-SKIP word="fuer"   reason=already-valid
AUTOCORRECT-SKIP word="schoen" reason=already-valid
```

Against, in the same session:

```
AUTOCORRECT branch=accent orig="fur"  corr="für"
TRIE-CANDIDATES word="feur" winner="für"(freq=56787)
```

`fur` and `feur` correct. `fuer` and `schoen` do not. The difference is not the scorer — it is that the engine returns before the scorer runs.

## Mechanism

`TextPredictionEngine.swift:194` — the valid-word guard:

```swift
// Valid word guard: if the word exists in the trie dictionary, it's correct.
if aospTrieEngine.wordExists(wordToCheck) {
    return nil
}
```

`germanProfile.collapseRules` (`German.swift:30`) declares `ae → ä`, `oe → ö`, `ue → ü`, `ss → ß`. Its own comment states the intent:

> German Umlautersatz: standard ASCII transliterations Germans use where umlauts are unavailable or unwanted (…) a user who still types `koennen` expects `können`.

Those rules run inside `expandAccents`, which the guard returns before. So the feature is written, tested, shipped, and unreachable for exactly the words it was written for.

## Why the words are in the dictionary

`fuer` and `schoen` are not German words. Standard orthography has only `für` and `schön`. They are Umlautersatz — the ASCII substitution used where umlauts cannot be typed. The dictionary was built from a corpus containing transliterated text, so the transliterations were counted as vocabulary.

So the guard is not misbehaving. It is being fed a dictionary that asserts a misspelling is a word.

## Desired behavior

- Typing `fuer`, `schoen`, `koennen`, `Tuer`, `Maedchen` and the other Umlautersatz forms offers the umlaut spelling, the way `strasse` → `straße` already does.
[...]

Issue #287, opened 2026-08-01
The user dictionary learns every word on first use: repetitionThreshold is 1, so typos are learned permanently


Pivii:
> *This was generated by AI during triage.*

## Summary

`UserDictionary.repetitionThreshold` is `1`. Despite the name, no repetition is required: a word enters the personal dictionary the first time it is typed. A typo typed once is learned and kept.

The maintainer also observes that a word learned this way does **not** then appear in the suggestion bar. If that holds, the mechanism currently has all of the cost and none of the benefit: it accumulates entries, including wrong ones, without improving suggestions.

Reported on 2026-08-01 during device testing of PR #284. Independent of that PR, which rewrote three force unwraps inside the counter without changing the threshold or any downstream behaviour.

## Evidence

- `repetitionThreshold` is `1` in `UserDictionary`. The promotion path from pending to learned therefore fires on the first call.
- The DictusCore test suite has a case for the below-threshold branch that is **skipped**, with the reason recorded in the test: at a threshold of 1 there is no below-threshold call to observe. The pending counter exists but is structurally unreachable as a gate.
- `recordUsage` is called from the keyboard extension on every word boundary, so the learning path runs constantly during normal typing.
- Maintainer observation, device, 2026-08-01: typing an unusual word once results in it being learned, and it does not surface in the suggestion bar afterwards. Stated to be long-standing, not a regression.

## The two questions this issue has to answer

**1. Should the threshold be greater than 1, and what should gate learning?** A raw repetition count is the cheapest gate, but it is not obviously the right one: a typo repeated twice would still be learned. A frequency-weighted validity check — is this string plausible in the active language before we commit it — is the direction already recorded for #114. This issue should either adopt that or state why a simple count is enough. Deciding this needs a view on how much a wrong entry actually costs, which depends on question 2.

**2. What does a learned word actually do?** If learned words do not reach the suggestion bar, then the whole feature is currently inert on the benefit side, and the priority of question 1 changes completely: an inert mechanism that pollutes storage is a different problem from one that actively degrades suggestions. **Establish this first.** It is cheap to check and it determines whether this is a correctness bug or a dead feature.

## Related

- #114 — align autocorrect with AOSP LatinIME, where the frequency-weighted valid-word guard is already recorded as the intended direction. If that work lands first it may subsume question 1 entirely; this issue should be re-read against it before anyone starts.
- #80 — custom vocabulary, the deliberate, user-curated counterpart to this accidental learning.

## Out of scope

[...]

Issue #285, opened 2026-08-01
patch-fluidaudio-swift5.sh reports success without patching the checkout being built


Pivii:
## Summary

`scripts/patch-fluidaudio-swift5.sh` reports success when it has patched nothing that matters. It searches only `~/Library/Developer/Xcode/DerivedData`, so any build using a custom `-derivedDataPath` — a git worktree, or the exact invocation our own CI workflow uses — compiles against an unpatched FluidAudio checkout and fails inside the dependency. The failure reads as "the project is broken", not "you forgot a step", and it has now cost time more than once.

## What happens today

The script's search is hardcoded:

```bash
done < <(find "$HOME/Library/Developer/Xcode/DerivedData" \
  -path "*/SourcePackages/checkouts/FluidAudio/Package.swift" 2>/dev/null)
```

`FOUND` is set as soon as *any* checkout matches, anywhere on the machine, including long-dead ones. On a developer machine there are typically several stale `Dictus-*` derived data directories, so the guard always passes.

Observed on 2026-08-01 in the `dictus-wt/146` worktree, building with the same flags as `.github/workflows`:

```
===== PATCH FLUIDAUDIO =====
Already patched: ~/Library/Developer/Xcode/DerivedData/Dictus-fyfecdkprwqdhsccmqpdmlfvrfaj/.../FluidAudio/Package.swift
Already patched: ~/Library/Developer/Xcode/DerivedData/Dictus-albmnaffxnjmfvhkyqgzdeufexyx/.../FluidAudio/Package.swift
... (10 lines, all stale directories)
===== BUILD =====
** BUILD FAILED **
SwiftCompile normal arm64 Compiling RnntDecoder.swift, StreamingAsrManager.swift, ...
  (in target 'FluidAudio' from project 'FluidAudio')
```

The checkout actually being compiled was `dictus-wt/146/build/DerivedData/SourcePackages/checkouts/FluidAudio/Package.swift`, which the script never looked at. Applying the same edit there by hand and rebuilding gave `** BUILD SUCCEEDED **` with no other change.

## Why this keeps recurring

Three properties compound:

1. **The patch target is ephemeral.** It edits an SPM checkout, which is wiped by "Reset Package Caches", by deleting DerivedData, and sometimes by Archive. The script's own header says so. Every wipe silently re-arms the failure.
[...]

PR #360, opened 2026-08-14
feat(keyboard): offer a learned word in the completions bar (refs #346)

Pivii:
## What this was for

A word the keyboard has learned never reached the suggestion bar — not in any of the three modes, not under any condition. Learning a word did exactly one thing: it stopped autocorrect from rewriting it. The user was never offered the word back, which is why #287 read as "all of the cost and none of the benefit".

This gives a learned word **L1, offered**, on top of the **L2, immune** it already had. It stays out of **L3, authoritative** — the keyboard still never rewrites another word *into* a learned word. That is #114's.

Concretely: type the first letters of a word you taught the keyboard, and the word is there in the bar, waiting for a tap.

## Reworked after the first device pass

The first version put the learned word in the first slot of `.completions` only. On device that was nearly invisible, because `.corrections` outranks `.completions` and the middle of a word is where the corrector always has a candidate:

| typed | correction candidate | learned word visible |
|---|---|---|
| `Z` | none | yes |
| `Zo` | `Ai` | no |
| `Zor` | `Air` | no |
| `Zorg` | `Sort` | no |
| `Zorgl` | `Sorti` | no |
| `Zorglu` | none | yes |

It flashed at one letter, vanished for four keystrokes, returned at the end. It also explains the "zorbax was never learned" report: it *was* learned, it just never got a slot afterwards.

The rework does **not** give the learned word priority over `.corrections` — that row is the preview of what space will apply, and hiding the correction to show a learned word would make the preview lie. Instead the row's third slot changes hands: `[typed | correction | alternative]` becomes `[typed | correction | learned word]` when a learned word strictly extends the typed prefix. The alternative keeps the slot when no learned word matches. Corroborated by the maintainer's measurement of Apple's keyboard, which puts a learned word in the third slot at three characters.

**Slot 1 stays the corrector's, deliberately.** Apple also promotes the learned word to the applied slot at four characters. That is L3, deferred to #114 by decision 5 of #287, and it is the door ADR 0004 exists to keep shut.

## To test by hand

**Reset the learned dictionary first.** Settings → "Reset learned words". A stale `user-learned` entry shadows everything and has invalidated a device session before. Then turn on Developer → "Autocorrect debug logs".

1. **Teach the keyboard a word.** In Messages, type `zorglub` and press space. Autocorrect rewrites it. Press backspace once to undo. Expected: your word comes back, and the export carries `USERDICT-LEARN word="zorglub"`.
2. **The word survives the whole prefix now.** New message, type `Z`, `Zo`, `Zor`, `Zorg`, `Zorgl`, `Zorglu`, one letter at a time. Expected: `Zorglub` visible at **every** step — first slot when the bar shows completions, **third slot** when the bar shows a correction row like `"Zor" | Air | Zorglub`. This is the criterion the first pass failed; there must be no four-keystroke hole in the middle.
3. **The bold middle slot is never the learned word.** Same typing. Expected: slot 2 always holds the corrector's word (`Ai`, `Air`, `Sort`, `Sorti`), never `Zorglub`. This is the L3 line and it matters more than anything else in this list.
4. **Tapping the third slot inserts the learned word.** At `Zorg`, tap `Zorglub` in the third slot. Expected: the partial word is replaced by `Zorglub` plus a space.
5. **Space still applies the correction, not the learned word.** At `Zorg`, press space instead of tapping. Expected: `Sort` (whatever the bold slot showed) lands in the text. In the export, no `APPLIED` line ever names a learned word as the correction.
6. **Mot tapé en entier → barre vide.** Type `zorglub` in full. Expected: the bar goes empty — there is nothing longer to complete. Press space: ordinary next-word predictions appear.
7. **Casing follows what you typed.** `zorg` offers `zorglub`, `Zorg` offers `Zorglub`.
8. **Two learned words: the most recent wins.** Learn `zorbax` as in step 1, then type `zo`. Expected: `zorbax` takes the slot, and only one learned word ever appears.
9. **Autocorrect off still offers it.** Settings → autocorrect off, type `zo`. Expected: `zorglub` still offered. (Off blocks *learning*, not offering.)
10. **No language filter.** With a German entry from earlier testing, switch to French and type its first letters. Expected: offered. Deliberate — the store carries no language tag.

## What changed and why

Four commits.

**`LearnedWordCompletions` — a pure DictusCore type** (`DictusCore/Sources/DictusCore/TextCorrection/LearnedWordCompletions.swift`) with one selector and two layouts:

- `offer(for:in:)` — the single selector both layouts share. At most one learned word, **strictly** extending what was typed, most recently used wins, with the word itself as a total-order tiebreak so two entries stamped in the same second can't depend on `Dictionary` iteration order. Casing reconstructed from the typed prefix with the same test the accent path uses, because `UserDictionary` lowercases its keys.
- `merge(...)` — the `.completions` list: learned word at index 0, system completions filling the rest.
- `correctionsRow(...)` — `[typed | correction | learned-or-alternative]`. Built here, not at its two call sites, because `update` and `updateAsync` both assemble it and must not disagree for the same keystroke.

It lives in DictusCore because that is the only target `swift test` can see, and because #287's block A spent three device passes on logic that was correct only where it happened to be called from. `UserDictionaryPruneGate` is the shape it copies.

**`UserDictionary.learnedWordsByLastUsed`** — new public accessor. Deviation from the brief, which said to hand `allLearnedWords` to the merge: that map is word → *usage count*, and decision 4 ranks by `lastUsed`, which was private. Rather than pass the wrong number and ask the merge not to use it, a second accessor returns the stamp map, whose key set is the same by the invariant `stampEntriesMissingATimestamp()` repairs on every load.

**Two call sites in the keyboard.** `TextPredictionEngine.suggestions(for:)` for `.completions` — the single production source, which both `SuggestionState.update` and `updateAsync` route through. Two details there: the system checker's `nil` stops being an early return (a prefix it knows nothing about is exactly the case a learned word exists for), and the full ranked list goes to the merge and is clipped afterwards, so a learned word the checker also returns fills a slot instead of costing one. And the corrections branch in both `update` and `updateAsync`, which now calls `correctionsRow`. The async path snapshots the learned set on the background queue next to the completions, so the row and the completions are built from one view of the dictionary.

No cache on the learned set — it is in memory, holds tens of entries since #287, and this runs on the suggestion queue; a cache would add an invalidation to get wrong, since the other process writes to that store too.

**The dead hook is deleted** — `TextPredictionEngine.injectUserWord`, `AOSPTrieEngine.injectUserWord`, `SuggestionState.learnWord`, and both call sites. Both call sites had the real work in the `if` condition (`recordUsage`, `learn`), so those calls kept their effect; both are `@discardableResult`, so `--strict` stays green with no suppression.

**Why L3 needs no gate.** On space only `spellCheck` can apply anything, and `spellCheck` never sees this file. `.corrections` is the one mode that previews an automatic action, and it previews it in slot 1, which `correctionsRow` fills from the corrector and never from the learned set. Slot 2 and the completions row are reached only by a tap. The rule holds by construction. That is written at the top of the file as the thing to re-check before any future change lets a learned word reach `spellCheck`, the trie, or slot 1.

Nothing is injected into the trie, so #326's log-normalized frequencies are not involved at any point.

## Acceptance criteria

|---|---|---|---|
| 1 | Learned word offered as a completion of its own prefix, from the first letter | **Device-confirmed for the completions path; the corrections path is the rework and is unverified** | First pass on device: visible at `Z` and `Zorglu`, absent in between. That hole is what the third-slot change closes — manual step 2. |
| 2 | Typing it in full leaves the bar empty until space | **Device-confirmed (first pass)** | Reported behaving; unchanged by the rework, and still covered by `testAFullyTypedLearnedWordIsNotEchoedBack` and `testAFullyTypedLearnedWordDoesNotClaimTheCorrectionsSlot`. |
| 3 | At most once, first slot, casing of the typed prefix | **Device-confirmed for casing; one-word-max in the corrections row is unverified** | `testAtMostOneLearnedWordReachesTheCorrectionsRow`, `testTheCorrectionsSlotFollowsTheCasingOfTheTypedPrefix`. Manual steps 7–8. |
| 4 | Never applied by pressing space; `AUTOCORRECT-SKIP … reason=user-learned` present; no `APPLIED` names a learned word | **Device-confirmed (first pass), re-check needed after the rework** | Export showed only `Zor`→`Air` and `Zo`→`Ai` applied, neither from the learned set, and `AUTOCORRECT-SKIP … reason=user-learned` throughout. The rework touches the row the space key does **not** read (`handleSpace` calls `performSpellCheck`, never `suggestions`), and `testTheCorrectionSlotIsNeverTheLearnedWord` pins slot 1. Manual steps 3 and 5. |
| 5 | Autocorrect disabled still offers previously learned words | **Device-confirmed (first pass)** | Reported behaving. Structurally unchanged: neither layout reads `autocorrectEnabled`. Manual step 9. |
| 6 | `swift test` passes, with the five named cases covered | **Met** | `cd DictusCore && swift test` → `Executed 951 tests, with 0 failures (0 unexpected)`. 31 of them new (`swift test --filter LearnedWordCompletionsTests` → `Executed 31 tests, with 0 failures`), covering exact-match dropped, several matches resolved by `lastUsed`, slot placement in both layouts, the one-word maximum, and both casings. |
| 7 | `swiftlint lint --strict` clean | **Met** | `Done linting! Found 0 violations, 0 serious in 178 files.` |

Build, from the worktree root, all three steps with the explicit derived-data path (#285): `xcodebuild -resolvePackageDependencies` → `./scripts/patch-fluidaudio-swift5.sh build/DerivedData` (`FluidAudio checkout patched for the build using: …/346/build/DerivedData`) → `xcodebuild build -scheme DictusApp -destination 'platform=iOS Simulator,id=231991C2…'` → **BUILD SUCCEEDED**, with `DictusKeyboard.appex` embedded and validated. Re-run after the rework.

App-side launch smoke, headless: installed on the iPhone 17 simulator, `simctl launch` returned a pid and the process was still listed six seconds later. That is a launch smoke and nothing more — the simulator cannot enable this keyboard, so it exercises none of criteria 1–5.

## Things worth flagging

**The third-slot rule is a policy, not a measurement.** Apple shows a learned word in slot 3 at three characters; this shows it from the first, because our selector has no minimum prefix length (decision 5) and the volume that justified that is unchanged. If the bar feels hijacked in use, that use supplies the threshold — inventing one now would repeat the arithmetic-at-a-guess #326 already cost.

**Thread safety is unchanged, not good.** `UserDictionary.shared` is read from the suggestion queue by one more accessor, and by the main thread in the sync `update()` path — the same as the existing `isLearned` read, against `learn()` running on main. No new hazard in kind, and no fix attempted here.

https://claude.ai/code/session_01RUUyDpuVjZnLoXMdB2Ynn4

## Summary by CodeRabbit

* **New Features**
  * Suggestions now include recently used learned words alongside system completions.
  * Learned-word suggestions support prefix matching, capitalization, recency ranking, deduplication, and result limits.
  * Correction suggestions can include a learned word in the third slot while preserving available alternatives.
  * Learned matches remain available when system completions are unavailable.

* **Bug Fixes**
  * Unknown and restored words are now recorded consistently for future suggestions.
  * Duplicate and exact-match learned suggestions are excluded for more relevant results.


Pivii:
Two findings triaged, and a gap in the review itself.

**The sort direction is a false positive, and applying the suggested fix would break the bar.** `FrequencyDictionary.rank(of:)` returns the frequency **count**, and its own doc says so — *"Returns the word frequency count (higher = more common). Returns 0 if the word is not in the dictionary."* (`DictusCore/Sources/DictusCore/FrequencyDictionary.swift:68-73`). So descending order puts the most common completion first, which is what the code does and what is wanted. Flipping the comparator would surface the rarest completions first.

What is genuinely wrong is the **comment** that produced the finding: `TextPredictionEngine.swift` says *"lower rank = more common = first"*, contradicting `rank(of:)`'s own documentation. Both the comparator line and the comment predate this PR; the diff only moved them. Fixing the comment belongs in its own change rather than inside a suggestion-bar feature, but it should be fixed — this is the second reader it has misled, and the first one was a bot proposing to invert working behaviour.

**The thread-safety finding is real, pre-existing, and deliberately out of scope.** `UserDictionary.shared` was already read from the suggestion queue by `isLearned` while `learn()` runs on the main thread; `learnedWordsByLastUsed` adds a second reader of the same shape, not a new class of hazard. It deserves its own issue with #103 (iCloud KVS) in view, since any synchronization boundary drawn here has to survive that merge strategy.

**The gap:** this review covered `839842d`. The branch has since been reworked to `cd4db32`, which changes what the `.corrections` row contains — the learned word now takes the third slot, `[typed | correction | learned]`, while the applied slot stays with the corrector. That is the half of the feature with a safety property attached (see ADR 0004), and it is the half that has not been reviewed.

@coderabbitai review

Issue #342, opened 2026-08-08
A second dictation error inside the 3 s window is cleared by the first message's timer, so the user never reads it


Pivii:
## Symptom

When two dictations fail within three seconds of each other, the user reads the first message and never sees the second. The second is set, sits on screen for whatever is left of the *first* message's timer, and disappears — sometimes after less than a second, sometimes instantly.

Observed on device (`1.8.0 (25)`, rev `efedda0`), while device-validating #320. Tapping the mic twice in quick succession is enough: both attempts fail on "recording too short", and only one message is ever read.

```text
[13:42:27Z] KBD dictationMessageCleared reason=replaced displayedCount=3   <-- message A replaced
[13:42:27Z] KBD dictationMessageSet reason=appError owner=952A01A2 visible=true   <-- message B
[13:42:28Z] KBD dictationMessageCleared reason=appError-timeout displayedCount=0  <-- B dies after 1s
```

`displayedCount=0` on B. The `dictationMessage*` events added by #261 exist to answer "was the user told", and here they answer no. B lived one second because A's three-second timer was still running.

## Root cause

Every status message schedules an unconditional clear:

`DictusKeyboard/KeyboardState.swift:627`

```swift
assignStatusMessage(errorMsg, reason: "appError")
DispatchQueue.main.asyncAfter(deadline: .now() + 3) { [weak self] in
    self?.assignStatusMessage(nil, reason: "appError-timeout")
}
```

The closure clears **whatever message is current when it fires**, not the message it was scheduled for. It captures no identity, and there is no handle to cancel it. So a message set at `t+2` is destroyed at `t+3` by its predecessor's timer, after one second of life.

The same shape appears a second time, at `KeyboardState.swift:1493`:

```swift
assignStatusMessage(String(localized: "Recording interrupted. Please try again."), reason: "reconciled-\(source)")
DispatchQueue.main.asyncAfter(deadline: .now() + Self.interruptedMessageDuration) { [weak self] in
    self?.assignStatusMessage(nil, reason: "reconciled-timeout")
}
```

Both use three seconds, and the two cross-clear each other as readily as each clears itself: a `reconciled` message raised one second after an `appError` message inherits two seconds of life, and the log will name the wrong `reason` on the clear line.

## Why it is worth fixing

Not because the message is pretty. Because of what the second message usually is.

[...]

Issue #313, opened 2026-08-05
Error messages in the keyboard toolbar are truncated, styled outside the design language, and half of them are not written for a user


Pivii:
## Summary

Dictation errors reach the user as one line of red text in the keyboard toolbar. Two things are wrong with it: longer messages are cut off mid-sentence, and the treatment is a plain system red that belongs to no part of the design language. Underneath both is a third problem that has to be settled first: there is no contract for what an error message is, so the copy that arrives is whatever each call site happened to write, including raw CoreAudio error numbers.

Not urgent. Raised so the visual pass is done once, on top of a message layer that has been made coherent, rather than twice.

## What it looks like today

`DictusKeyboard/Views/ToolbarView.swift:109`

```swift
if let message = statusMessage {
    Text(message)
        .font(.caption)
        .foregroundColor(.red)
        .lineLimit(1)
        .frame(maxWidth: .infinity)
```

- **`.lineLimit(1)`** is the truncation. Anything past roughly forty characters at `.caption` in the toolbar width is tail-clipped. "No model downloaded. Open Dictus to download a model." is fifty-two.
- **`.foregroundColor(.red)`** is the system red, not a token from `assets/brand/dictus-brand-kit.html`. The brand's own error-adjacent colour is `#EF4444`, and nothing in the keyboard references it.
- The message takes the whole toolbar slot for its 3 s, hiding the hamburger and the suggestion bar, then vanishes. There is no way to see it again once it has gone.

## The message layer has no contract

Eleven `handleError` call sites in `DictusApp/DictationCoordinator.swift`. What they pass:

| what is passed | sites | consequence |
| --- | --- | --- |
| `String(localized:)` | 2 | correct — translated, reviewed |
| English literal (`"Microphone permission denied"`, `"No audio recorded"`, `"Recording too short"`) | 5 | English on a French device |
| French literal (`"Micro indisponible. Relancez l'application."`) | 2 | French on an English device |
| `error.localizedDescription`, raw | 2 | a CoreAudio error number can reach the screen |

The last row is the sharp one. #311 established that `AUIOClient_StartIO` surfaces as *"L'opération n'a pas pu s'achever. (com.apple.coreaudio.avfaudio erreur 2003329396.)"*. That path was given a written message in #312, but the two generic sites were left alone because they were out of that issue's scope. They can still put that string in front of a user, and it is both meaningless and far too long for one line.

So "make the errors prettier" cannot be done on its own: half the strings are not designed at all.

[...]

Issue #361, opened 2026-08-14
Move the polish call into the keyboard extension, where Apple FM actually works


Pivii:
> *This was generated by AI during triage.*

**Needs a grilling session before anyone builds it.** The measurement is done and the direction is settled; the open decisions at the bottom are not, and several of them change the shape of the work.

## Why

Apple refuses `LanguageModelSession` generations to backgrounded processes. DictusApp is backgrounded for every polish call by design, so the polish layer runs in the one state Apple deprioritises — measured at ~10% failure over a spread-out day and **45% during a dense working session**, latching for twelve minutes at a time with no recovery short of the process being killed. Full analysis in #315.

**#357 measured the way out.** Ten consecutive Apple FM generations from inside `DictusKeyboard`, fired ten seconds after the app's last instant refusal:

```text
07:28:52  <APP>  rateLimited   engineMs=10          <-- app refusing, appState=2
07:29:06  <KBD>  n=1/10   success  engineMs=4414
   …
07:29:50  <KBD>  n=10/10  success  engineMs=4825    <-- no degradation
          memBefore 15 MB → memPeak 22 MB → memAfter 19 MB
```

The keyboard extension is in the foreground at exactly the moment polish runs — the user has stopped speaking and is waiting for text. Moving the call there does not work around the limit; it stops meeting it.

Three things #357 also settled: the call costs **7 MB of peak delta**, there is **no latency penalty** (4.4-5.1 s on 462 characters, matching what the app produces on comparable input), and Apple FM inference does not run in the calling process.

## What has to move, and what does not

**Already shared — nothing to do.** Everything in `DictusCore/Sources/DictusCore/Polish/` is in the framework both targets link: `PolishEngineProtocol`, `AppleFoundationModelsPolishEngine`, `PassthroughPolishEngine`, `PolishPipeline`, `PolishGuardrail`, `PolishContextBudget`, `PolishPostpass`, `VerbalPunctuationPrepass`, `PolishMetrics`, `PolishMode`, `PolishFailureReason`, `PolishGlossary` and all nine prompts. The engine the probe called was the shipping one.

**App-side today, and the actual work:** `PolishCoordinator`, `PolishAvailability`, `PolishMetricsRing`, `PolishDebugExporter`, `PolishDebugView` — all in `DictusApp/Polish/`.

[...]

PR #373, opened 2026-08-22
Stop calling a refusing polish engine, and say so (refs #315)

Pivii:
Closes decision 14 of #315. Written as `refs #315` in the commits: the issue stays open until the maintainer closes it by hand.

## What this was for

Polish silently stops working. Apple's `rateLimited` is a per-call refusal applied to apps running in the background, and DictusApp is backgrounded for every polish call by design — the keyboard holds the screen. Once the refusals start, nothing brings polish back short of a fresh app process: the 2026-08-13 experiment showed twenty seconds of foreground did not refund the budget, and no amount of wall-clock ever has in any capture on the issue. The measured result is up to twelve minutes at a stretch where every dictation loses its punctuation and repair, the app keeps calling a dead engine, and the user is told nothing.

This PR does decision 14 and nothing else: **stop calling a refusing engine, and say so.** No recovery action, no button, no retry, no backoff — decision 7 killed the mechanism they would have relied on.

## To test by hand

Everything below needs a physical iPhone with Apple Intelligence and polish enabled. The keyboard extension cannot be enabled in any simulator (`docs/agents/simulator.md` §7), so the visible state is manual by construction.

1. Settings → polish **on**. Force-quit DictusApp, relaunch it once, then leave it. This is the fresh process the reset depends on. **Expect:** polish works normally, no notice in the keyboard toolbar.
2. Dictate normally from the keyboard, one dictation every one to two minutes, for about twenty-five minutes. That is the density that reproduced the refusals twice on 2026-08-13. Long dictations arm it faster than short ones.
3. Watch for the moment two dictations in a row come back unpolished. **Expect:** from the third dictation on, the text appears **noticeably faster** than a polished one — the engine is no longer being called at all, so the wait drops from seconds to nothing.
4. Look at the keyboard toolbar with the cursor idle (no word half-typed, no undo chip showing). **Expect:** to the right of the hamburger, in grey, `Le polissage est temporairement indisponible.` Nothing else — no cause, no remedy, no button.
5. Type a word. **Expect:** the suggestion bar takes the slot back and the notice steps aside. Stop typing and let the bar go idle. **Expect:** the notice returns. It has no timer; it is a state.
6. Tap the hamburger. **Expect:** the panel still opens — the notice shares the slot with it rather than replacing it.
7. Keep dictating for several more minutes. **Expect:** the notice never expires and never comes back on its own. This is the difference from a `statusMessage`, which would have vanished after three seconds.
8. Watch the recording overlay during one of these dictations. **Expect:** it goes straight from `Transcription...` to done. It must **not** flash `Traitement...` — no engine runs, so no processing stage is announced.
9. Force-quit DictusApp and relaunch it. Dictate again. **Expect:** polish works, and the notice is gone. This is the only reset there is.
10. Export the polish debug JSON from the hidden debug screen. **Expect:** `"outcomes"` contains `"engineUnavailable": N`, one per dictation skipped in step 3 onward, and those events carry no `failureReason` (nothing failed — we declined to ask). The two `rateLimited` that armed it are still there as `engineFailed`.
11. Export the persistent log. **Expect:** exactly one line `polishEngineUnavailable engine=apple-fm reason=rateLimited consecutiveRefusals=2`, at the transition, and none after it.

## What changed

**Detect** — `PolishAvailabilityGate` (DictusCore). Two consecutive `rateLimited` results from one engine flip it. Two, not three: every `rateLimited` run captured on the issue arrives in a run of at least two, and a single isolated one has never been observed. A run is broken by any verdict proving the engine served the call — a success, a guardrail rejection, a failure naming another reason. Outcomes where the engine was never reached or was cut short (cancelled, context overflow, the skips) neither count nor reset, so a flash dictation in the middle of a run cannot hide it. The gate is scoped to the engine that refused, so Apple Intelligence being switched off mid-process does not carry the block onto the passthrough backend.

The rule lives in DictusCore because DictusApp has no test bundle. The **state** lives on `PolishCoordinator`, which already owns every other per-process polish concern.

**Stop calling** — `PolishPipeline.transform` takes the gate and exits before the newline encode and before the context guard, returning the new `engineUnavailable` outcome. `resolvedOutput` already hands back the deterministic floor for every non-success, so what the user receives is unchanged. `announceEngineStage` reads the same gate value, so no `.processing` stage is raised for a call that is not happening.

**Reset** — a fresh process, and nothing else. There is no timer, no foreground observer and no retry anywhere in the new code; the gate is a plain value on the coordinator, and a new process gets a new one. The App Group flag the keyboard reads is cleared in `PolishCoordinator.init()` — the same object that owns the gate, so what is on disk and what is in memory cannot describe different states.

**Say so** — a second display path beside `statusMessage`, not a change to it. That path carries a three-second timer (the one behind #342) and this condition can last the whole app process. The notice sits beside the hamburger, in the toolbar's centre slot, shown only when that slot is otherwise idle. Above the suggestions it would suppress completions and corrections for the whole process, which is the keyboard's core job; in place of the hamburger it would make the panel unreachable for the same duration. Sharing the slot costs visibility while the user is mid-word, and that is the cheapest of the three prices. Secondary grey rather than the red of `statusMessage`: nothing failed for the user, the dictation still arrives. Single line inside the existing 52 pt bar — no height constraint touched (#166 and family).

**Record it** — a new `engineUnavailable` outcome, deliberately distinct from `engineFailed`: nothing failed, we declined to ask, and folding the two would inflate the failure rate this issue has been tracking since 2026-08-05. Plus one `polishEngineUnavailable` log line on the transition. No new metric *field*, so events persisted by shipped builds decode unchanged.

**Engine defect 1** — `polish()` ended with `defer { Task { await cache.drop(key) } }`. The drop now happens on both explicit exits *and* names the session it is removing (`SessionCache.dropIfStillCurrent`). Awaiting alone was not enough: the window is the unconditional key-based removal, which spans the whole 4-5 s `respond()` suspension, so a `prewarm()` for the next recording can install a warmed replacement under the same key while the first call is still suspended. Identity comparison is what closes it. `prewarm()` keeps its unconditional drop, since it is deliberately displacing whatever is there. **No automated coverage:** `SessionCache` is file-private and a real `LanguageModelSession` needs the FoundationModels SDK, so testing the identity behaviour would mean widening its visibility for the test alone.

**Engine defect 2** — `prewarm()` measured, see below. Not removed.

## Acceptance criteria

| --- | --- | --- |
| Two consecutive `rateLimited` from `apple-fm` → unavailable | ✅ | `testTwoConsecutiveRefusalsStopTheEngine`, `testOneRefusalDoesNotStopTheEngine`, `testThresholdIsTwo` |
| Consecutive means consecutive | ✅ | `testASuccessBetweenTwoRefusalsResetsTheRun`, `testAnotherFailureReasonResetsTheRun`, `testOutcomesThatNeverReachedTheEngineLeaveTheRunAlone` |
| While unavailable the engine is not invoked | ✅ | `testTransformDoesNotCallAnUnavailableEngine` — the stub calls `XCTFail` if reached |
| Latency drops to roughly zero | ✅ | `testUnavailablePathCostsNothingComparedToACall`: same 300 ms engine, same input; available arm > 250 ms, unavailable arm < 50 ms |
| Deterministic floor unchanged | ✅ | `testUnavailableStillReturnsTheDeterministicFloor` (FR NBSP applied), `testUnavailableAutoModeReturnsTheAutoFloor` |
| Reset on a fresh process only, no timer / re-arm / retry | ✅ | `testFreshGateAllowsTheCall`; `grep -rnE "Timer\|asyncAfter\|Task.sleep\|didBecomeActive\|willEnterForeground"` over both new files → no match |
| Visible as a persistent state, exact copy, FR + EN | ⚠️ manual | `Localizable.xcstrings` entry added; the rendering is manual steps 4-7 above |
| Visible in the polish debug export | ⚠️ partly | `engineUnavailable` seeded in the `outcomes` block and in the debug screen's outcome list; the export itself is manual step 10 |
| Visible in the persistent log | ✅ | `testUnavailableLogLineNamesTheEngineAndTheReason` pins name, subsystem, level and the whole line |
| Persisted events from shipped builds still decode | ✅ | `testEventsFromShippedBuildsStillDecode`, `testOutcomeSurvivesTheMetricJSONRoundTrip` |
| `cd DictusCore && swift test` green | ✅ | `Executed 992 tests, with 0 failures (0 unexpected)` |
| `swiftlint lint --strict` clean | ✅ | `Done linting! Found 0 violations, 0 serious in 186 files.` |
| Three targets build | ✅ | `DictusApp` and `DictusKeyboard` schemes, `** BUILD SUCCEEDED **`; DictusCore via `swift build` and as the app's local package |

## The `prewarm()` measurement

The brief expected this to be unmeasurable headlessly. It is not: the iOS Simulator reports `SystemLanguageModel.default.availability == available` and real generations run.

Six alternated rounds, same French input (~200 chars), 1.5 s gap before each call — the prewarmed arm does exactly what the app does (prewarm at recording start, polish ~1.5 s later), the cold arm calls `polish()` with no prewarm:

```
round 0: warmed=2013 ms  cold=2071 ms
round 1: warmed=2003 ms  cold=2041 ms
round 2: warmed=1958 ms  cold=2031 ms
round 3: warmed=2020 ms  cold=2085 ms
round 4: warmed=1995 ms  cold=2021 ms
round 5: warmed=2084 ms  cold=2021 ms

warmed median = 2013 ms   range [1958, 2084]
cold   median = 2041 ms   range [2021, 2085]
```

**`prewarm()` buys about 28 ms on a 2 s call — 1.4%, inside the spread of the two arms.**

Read this with its limits. It ran in the Simulator, which proxies Apple Intelligence from the host Mac rather than using the phone's ANE, and the test process is in the **foreground** — which is precisely the condition Apple's caveat excludes (*"particularly if your app is running in the background"*). So this is prewarm's **best case**, and even there it is worth 1.4%. It cannot prove prewarm does nothing on a backgrounded phone, but it is hard to argue it does more there than here.

The removal decision is the maintainer's; nothing was removed. If it is worth a field confirmation, the phone version is two builds — `develop`, and `develop` with the `session.prewarm()` line commented out — ten dictations of similar length each, comparing median `engineMs` from the debug export.

## Not comfortable with / worth knowing

- **A stale notice window.** The gate is in memory and resets with the process, as specified. The App Group flag that the keyboard *displays* is cleared by the next DictusApp launch. So if iOS kills the backgrounded app while polish is unavailable, the flag stays true until the app next launches — which is the next dictation. During that gap the keyboard shows a notice for a budget that has already been refunded. It self-heals on the next dictation and I chose not to add machinery for it, because every fix I could see (a timer, a foreground observer, a process token) is either the re-arm the brief forbids or belongs to #313.
- **The toolbar slot is contested.** The notice loses to suggestions and to the undo chip, which means it is hidden in the seconds right after a dictation — the moment it is most relevant. Deliberate, and the reasoning is in the code and above, but it is the judgement call in this PR most worth disagreeing with.
- **Nothing contradicted the issue.** The brief matched the code as found.
- **Two review fixes landed after the maintainer's device validation** (`ac0451c`, `ac87740`), both listed above. Neither touches the gate, the skip path, the notice, the log line or the export, so nothing validated on `b375cea` changed.

## Summary by CodeRabbit

- **New Features**
  - Added safeguards that temporarily disable polish after repeated engine rate-limit refusals.
  - Added a visible “Polish is temporarily unavailable.” notice in the dictation toolbar.
  - Added shared availability status between the app and keyboard.
  - Added unavailable-engine outcomes to polish diagnostics and logs.

- **Bug Fixes**
  - Improved cleanup after successful or failed polish requests.
  - Prevented unnecessary polish calls when the engine is unavailable.

### DIFF
--- a/.out-of-scope/foreground-the-app-for-polish.md
+++ b/.out-of-scope/foreground-the-app-for-polish.md
@@ -1,0 +1,15 @@
+# Foregrounding the App to Escape the Apple FM Rate Limit
+
+Dictus will not bring DictusApp to the foreground in order to run a polish or Smart Mode call. Not per dictation, not periodically, not automatically.
+
+## Why this is out of scope
+
+Apple documents that `rateLimited` only occurs for apps running in the background, and a controlled test on 2026-08-13 (#315) showed the rule is applied per call: the same process, seconds apart, is refused backgrounded and served in the foreground. So foregrounding does work. It is rejected on cost, not on effect.
+
+DictusApp is backgrounded by design. The keyboard extension holds the screen, the app records and transcribes behind it, and the user never sees DictusApp during normal use. The one moment the product does foreground the app — cold-start dictation — is the worst moment in it, and it has three open issues to itself (#23 no API exists to return the user, #264 the home-screen flash, #311 the stranded dictation on an early swipe-back). Doing that after every dictation would apply the product's worst interaction to its most common one.
+
+A periodic variant was considered — foreground every twenty or thirty minutes to reset the budget — and fails on its own terms: the test showed a foreground visit does not refund the budget at all. Only a process restart does, and an app cannot restart itself.
+
+## What is in scope instead
+
+Moving the call to a process that is already in the foreground: the keyboard extension. Measured separately. Failing that, honest degradation, and a backend not subject to foreground priority (#268, #351).
--- a/DictusApp/Polish/PolishCoordinator.swift
+++ b/DictusApp/Polish/PolishCoordinator.swift
@@ -35,6 +35,15 @@ public final class PolishCoordinator {
     private let metricsRing = PolishMetricsRing()
     private var inflight: Task<PolishPipeline.Result, Never>?
 
+    /// Whether polish is still calling its engine (#315).
+    ///
+    /// Deliberately in memory and nowhere else. Apple's background rate limit is
+    /// only refunded by a fresh process, so "reset on a fresh process" and "do not
+    /// persist" are the same instruction, and this property is the whole of the
+    /// implementation. Anything that re-armed it — a timer, a foreground
+    /// observer, a retry — would be theatre against the measurements on #315.
+    private var availabilityGate = PolishAvailabilityGate()
+
     private init() {
         self.defaults = AppGroup.defaults
         #if canImport(FoundationModels)
@@ -46,6 +55,12 @@ public final class PolishCoordinator {
         #else
         self.appleFMEngine = nil
         #endif
+        // The fresh-process reset, on the flag the other process reads (#315).
+        // Done from the same object that owns `availabilityGate` above, so what
+        // is on disk and what is in memory cannot describe different states.
+        // This singleton is created from `DictusApp.init()`, so "fresh process"
+        // and "this line ran" are the same moment.
+        PolishAvailabilityChannel.clear()
     }
 
     /// Resolve the engine for this call. Re-checked on every `polish()` so a
@@ -326,14 +341,16 @@ public final class PolishCoordinator {
         // delegated to `PolishPipeline` so the eval harness runs identical code.
         let preprocessMs = Int(Date().timeIntervalSince(methodStart) * 1000)
 
+        let gate = availabilityGate
         let task = Task {
             await PolishPipeline.transform(
-                preprocessed: preprocessed, engine: currentEngine, target: target, mode: mode
+                preprocessed: preprocessed, engine: currentEngine, target: target, mode: mode, gate: gate
             )
         }
         inflight = task
 
         let bundle = await task.value
+        recordAvailability(bundle, engine: currentEngine)
         // True total, methodStart → here. Any gap vs (pre+engine+post) is
         // Swift Task scheduling / actor-hop overhead — itself worth seeing.
         let totalMs = Int(Date().timeIntervalSince(methodStart) * 1000)
@@ -441,14 +458,17 @@ public final class PolishCoordinator {
         announceEngineStage(currentEngine, to: onEngineWillRun)
         // Pre-engine work in auto mode: detection + detected-language pre-pass.
         let preprocessMs = Int(Date().timeIntervalSince(methodStart) * 1000)
+        let gate = availabilityGate
         let task = Task {
             await PolishPipeline.transform(
-                preprocessed: preprocessed, engine: currentEngine, target: contextLanguage, mode: .auto
+                preprocessed: preprocessed, engine: currentEngine, target: contextLanguage,
+                mode: .auto, gate: gate
             )
         }
         inflight = task
 
         let bundle = await task.value
+        recordAvailability(bundle, engine: currentEngine)
         let totalMs = Int(Date().timeIntervalSince(methodStart) * 1000)
         let returned = PolishPipeline.resolvedOutput(
             bundle, preprocessed: preprocessed, target: contextLanguage, mode: .auto
@@ -512,9 +532,39 @@ public final class PolishCoordinator {
     /// worth announcing (#267). Both polish paths reach the engine by different
     /// routes and each has to make this call; the check lives here so neither can
     /// make it differently.
+    ///
+    /// The availability gate (#315) is read here for the same reason it is read
+    /// inside the transform: while polish is unavailable there is no wait to
+    /// announce — the transform returns in about a millisecond without touching
+    /// the engine — and a `.processing` stage raised for it would flash a label
+    /// and a new animation for a single frame. Both decisions come off the one
+    /// gate value, so they cannot drift apart.
     private func announceEngineStage(_ engine: PolishEngineProtocol, to callback: (() -> Void)?) {
+        guard availabilityGate.allowsCall(engine: engine.identifier) else { return }
         guard engine.announcesProcessingStage else { return }
         callback?()
+    }
+
+    /// Feed one completed transform outcome to the availability gate (#315), and
+    /// on the single call that flips it, say so where both the user and a future
+    /// capture can see it.
+    ///
+    /// Both polish paths call this with the result they just got, so the rule
+    /// runs once per engine-facing dictation whichever route it took.
+    private func recordAvailability(_ bundle: PolishPipeline.Result,
+                                    engine: PolishEngineProtocol) {
+        let becameUnavailable = availabilityGate.record(
+            outcome: bundle.outcome,
+            reason: bundle.failureReason,
+            engine: engine.identifier
+        )
+        guard becameUnavailable else { return }
+        PersistentLog.log(.polishEngineUnavailable(
+            engine: engine.identifier,
+            reason: bundle.failureReason?.slug ?? "-",
+            consecutiveRefusals: PolishAvailabilityGate.consecutiveRefusalsBeforeUnavailable
+        ))
+        PolishAvailabilityChannel.markUnavailable()
     }
 
     /// Log one metrics event and append it to the debug ring.
--- a/DictusApp/Polish/PolishDebugExporter.swift
+++ b/DictusApp/Polish/PolishDebugExporter.swift
@@ -139,7 +139,7 @@ enum PolishDebugExporter {
         var outcomes: [String: Int] = [
             "success": 0, "rejectedGuardrail": 0, "skipped": 0,
             "skippedShort": 0, "skippedAutoMode": 0, "cancelled": 0, "engineFailed": 0,
-            "exceededContextBudget": 0
+            "engineUnavailable": 0, "exceededContextBudget": 0
         ]
         var failureReasons: [String: Int] = [:]
         for e in entries {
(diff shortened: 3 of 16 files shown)


Pivii:
## Device validation, 2026-08-22 evening: decision 14 reproduces end to end

Build `1.8.0 (26)`, `rev b375cea@HEAD`, iPhone 15 Pro Max, iOS 26.6. Installed over the wire from the PR head, both the app and the keyboard extension bundle stamped `b375cea`.

The gate armed on its own during ordinary use, which is the point: nothing was forced or simulated.

```text
20:52:21  engineFailed            rateLimited   engineMs=50    <-- slow onset, the known pattern
20:52:32  engineFailed            rateLimited   engineMs=9     <-- instant refusal
20:52:32  polishEngineUnavailable engine=apple-fm reason=rateLimited consecutiveRefusals=2
20:52:40  engineUnavailable                     engineMs=0     latencyMs=3
20:52:48  engineUnavailable                     engineMs=0     latencyMs=2
20:52:58  engineUnavailable                     engineMs=0     latencyMs=2
20:53:28  engineUnavailable                     engineMs=0     latencyMs=5
20:55:35  engineUnavailable                     engineMs=0     latencyMs=7
20:55:58  appLaunched version=1.8.0                            <-- fresh process
20:56:04  success                               engineMs=891
```

### What each acceptance criterion did on the device

| Step | Result |
| --- | --- |
| Two consecutive `rateLimited` arm the gate | Armed at 20:52:32, on the second refusal, not the third |
| The engine is no longer called | `engineMs=0` on all five skipped dictations |
| Latency drops to roughly zero | **2 to 7 ms**, against ~900 ms on the polished calls either side |
| The notice is visible, exact copy, secondary grey | Confirmed by screenshot: `Le polissage est temporairement indisponible.` beside the hamburger, nothing else |
| The notice yields to typing and returns at rest | Confirmed by the maintainer |
| The hamburger panel still opens | Confirmed by the maintainer |
| The notice never expires on its own | Held for the whole outage, roughly three minutes |
| No `processing` stage is announced | From 20:52:40 the keyboard goes `transcribing -> ready` directly. Before the gate it was `transcribing -> processing -> ready` |
| Reset on a fresh process only | 20:55:58 launch, 20:56:04 success |
| `engineUnavailable` in the debug export | `"engineUnavailable": 5`, and those events carry no `failureReason` |
| Exactly one log line at the transition | One `polishEngineUnavailable`, none after |

### The reset happened the way it happens in the field

The relaunch at 20:55:58 was not a manual force-quit. iOS terminated the backgrounded app around 20:55:50, and the log shows it from the keyboard's side: `coldStartDarwinFallback elapsedMs=521` at 20:55:57, which is the keyboard calling an app that is no longer there. So the reset was exercised on the path users actually take, not only on the deliberate one.

### The stale notice window, measured and accepted

The PR flagged this and the device run sharpened it. The gate lives in memory and dies with the process; the App Group flag the keyboard *displays* is cleared by the next `PolishCoordinator.init()`. When iOS kills the app mid-outage, the flag stays true until the app next launches, which is the next dictation. The maintainer saw exactly that: the notice was still on the keyboard after the process was gone, and cleared on the first recording.

So the window is not seconds, it is **until the next dictation**, which can be arbitrarily long if the user types without dictating. Recorded correctly here rather than understated.

**Decision: leave it.** The direction of the error is the safe one. A stale notice is a false positive, and the user has no action attached to it to be misled into. The failure this issue exists to kill is the false negative, where polish is dead and nothing is shown. Every fix considered is worse than the defect:

- **Ping the app from the keyboard on each appearance** (Darwin plus timeout, the mechanism already used for cold start). It works, and it costs a round trip every time the keyboard shows, to correct a cosmetic lie. It is also the re-arming machinery decision 14 forbids.
- **A heartbeat in the App Group.** DictusApp is backgrounded and suspended most of the time and cannot beat reliably.
- **Clear the flag on `appWillTerminate`.** Does not fire when iOS jetsams a backgrounded app, which is the case that produces the window. This run is the proof: the app died at 20:55:50 announcing nothing.

If it is ever worth doing, it belongs to #313, which owns the coherence of that display layer.

### On the copy not naming force-quit

Raised again during validation, and the answer from the 2026-08-13 session stands: a force-quit is the only user action that restores polish, and it cannot be asked of a user as routine. The user is inside another app on the keyboard, and the remedy would be to leave that app, find in the switcher an app they never open, kill it, and come back, mid-message, roughly every dozen dictations, since a fresh process spends its new budget the same way.

One correction to the brief's wording. It justified the copy with "the user has no action available". That is imprecise: an action exists, it is simply not printable. The decision is unchanged, the argument above is the accurate one.

The real remedy is #361, not better copy.

### Not this PR

Three unrelated failures in the same window: `guardrailViolation` twice and one `unsupportedLanguageOrLocale` on a 214 character French dictation. The maintainer was dictating with a television running, so Apple FM refusing the content is the expected behaviour of its own guardrail. Nothing opened.

Pivii:
@coderabbitai review

Issue #393, opened 2026-08-24
polish-harness cannot run a Smart Mode, so #79's validation requirement cannot be met


Pivii:
Found reviewing PR #389 (Smart Modes block A).

#79 requires it: *"Every shipped mode is harness-validated as visibly different from free polish."* PR #388 showed what that requirement is worth — it measured the Email mode, found a prompt that held the no-hallucination line across 190 calls, and **still cut the mode** on evidence nobody had before the measurement.

Notes and Translate ship in block A **unmeasured**. PR #389 flags this itself. What it does not say is that the measurement cannot be run.

## The blocker

`DictusCore/Sources/polish-harness/main.swift:190` and `:207` — `runOnce` and `runOnceAuto` both construct `.polish(mode)`. There is no path that builds a `.smart(...)` task, so no invocation of the harness can exercise a Smart Mode's prompt, its per-mode contract, or its per-mode user-turn framing.

`--instructions` now takes a `(PolishTask, SupportedLanguage) -> String` closure, but nothing routes a mode through it.

So the gap in #389's acceptance table is not "the run was not done". It is **tooling work, then a run.**

## Why it matters more than it looks

PR #389's own review found a defect in the Translate prompt that a measurement would have caught and a reading nearly did not: the example block was constant while the target varied, so the "→ English" instructions demonstrated an English input producing a French output — the rule being broken, in context, inside the prompt meant to enforce it. Fixed in `b59c43c`, found by eye. The next one may not be.

## Scope

- A way to select a Smart Mode from the command line, by catalogue identifier, so `notes` and `translate.en` are addressable
- The mode's own acceptance contract applied to the output, not the free-polish bands — otherwise Notes is rejected for condensing and Translate for changing language, which is the whole point of per-mode contracts
- The A/B form working against a mode, so a prompt candidate can be compared with the shipping one
- Fixtures that put pressure on each mode: for Notes, a long rambling dictation with no structure; for Translate, input already partly in the target, and input in a language outside the four

## Acceptance criteria

- [ ] `polish-harness` can run a named Smart Mode over a fixture file
- [ ] The per-mode contract is what judges the output
- [ ] Notes and Translate → EN are each measured against free polish, with the raw outputs committed as #388 did
- [ ] The result is recorded in #79 and each mode is either confirmed or cut, on the same bar Email was held to

Pivii:
## First field evidence: the Notes prompt drifts to English on French input

[...]

Issue #392, opened 2026-08-24
Smart Modes have no Pro entitlement check anywhere, including at resolution


Pivii:
Found reviewing PR #389 (Smart Modes block A).

Smart Mode is the first premium capability and the paywall trigger (#54, #55, #78). **Nothing in the code consults an entitlement.**

`SmartModeAvailability.armability` and `SmartModeStore.resolveArmedMode()` never call `FeatureGate.isAvailable(.smartMode)`, which already exists at `DictusCore/Sources/DictusCore/Subscription/ProFeature.swift:70`.

## Why this is not simply "block C's job"

The paywall's feature list *is* block C, and block A's doc comment half-says so. Two things make it worth its own issue rather than a line in the block C description.

**1. `armability` is named "the arming policy".** It answers one question — may a mode be armed — and it answers it on Apple Foundation Models availability alone. Someone building the long-press fan in block B can reasonably build it on that property and ship a paid capability with no gate, and the naming would not warn them.

**2. `resolveArmedMode()` is the natural enforcement point for a lapsed subscription, and it lives in block A.** It is what every dictation calls to turn the stored identifier into a mode. Today an armed mode survives a cancellation indefinitely: the user keeps their Smart Modes after they stop paying, until something else clears the value.

## The decision this needs

Where entitlement is enforced, and what the user sees:

- **At arming** — the fan shows modes disabled with "Dictus Pro" as the reason, alongside the Apple Intelligence reasons `SmartModeUnavailableReason` already carries. Consistent with the existing failure-early design, and the reason type is already shaped for it.
- **At resolution** — a mode armed while subscribed stops applying when the subscription lapses. Needed regardless, or a cancelled user keeps the feature.
- **Both**, which is likely the answer, with the question being what the second one says when it fires mid-use.

Note the interaction with #79's paywall decision: Pro is sold on devices without Apple Intelligence, where Smart Mode cannot run at all and Pro sells on other features. So "not entitled" and "not capable" are two different states that must produce two different messages, and `SmartModeUnavailableReason` currently models only the second.

## Acceptance criteria

- [ ] `FeatureGate.isAvailable(.smartMode)` is consulted before a mode can be armed
- [ ] An armed mode stops applying when entitlement lapses, without waiting for the user to open the app
- [ ] "Not entitled" and "device not capable" are distinguishable, with different copy
- [ ] `armability`'s doc comment states what it does and does not gate
[...]

Issue #391, opened 2026-08-24
Two Smart Mode error paths do not follow the rules the success paths do


Pivii:
> **Half of this is already on `develop`, verified 2026-09-05.**
>
> The keyboard recovery path was fixed: `KeyboardPolishCoordinator.recoverPendingIfNeeded` now checks `pending.smartMode`, logs `smartModeRefused`, calls `finish(text: nil, insert: false)` and announces the failure. The comment above it names #391 and names the other two insertion sites, so the fourth acceptance criterion is met too.
>
> **What remains is the second gap only**, and it is unchanged: in `DictusApp/DictationHandoff.swift`, the Smart Mode failure path returns at the `guard let text = outcome.text` before reaching the `removeObject` calls at lines 139-149, so a stale `lastTranscriptionSmartMode` / `lastTranscriptionPolicy` survives exactly the failure case the cleanup was written for.
>
> Remaining criteria: the in-app failure path clears the same App Group keys the success path clears, and a test covers it.

---

Found reviewing PR #389 (Smart Modes block A). Not reachable in the shipped build — nothing can arm a Smart Mode until block B — so this is a defect block B inherits rather than one that ships today.

## The rule, and where it is not applied

`PolishTask.isSmart` states it: **a Smart Mode must never silently insert untransformed text.** #79 calls that outcome "the worst outcome available" — the user asked for bullets, or for English, and would get neither without being told.

Block A applies the rule at two of the three insertion sites. The third is the keyboard's recovery path:

`DictusKeyboard/KeyboardPolishCoordinator.swift:190`

```swift
finish(text: DictationTail.apply(pending.raw, policy: pending.policy), insert: true)
```

`recoverPendingIfNeeded` never consults `pending.smartMode`. With "Translate → EN" armed, this inserts the untranslated French.

## Why it is filed rather than fixed in #389

The method's own doc comment (lines 155-168) records that this path has never fired in four device sessions, and that the `documentIdentifier` identity rule closes it from both sides. Both are true. But the guarantee is weaker than it reads: the rule is stated three times in prose and enforced at two of three sites, and the site that skips it is the one nobody watches because it never runs.

## Second gap, same shape

[...]

Issue #269, opened 2026-07-30
Custom Smart Mode: let the user write their own instructions


Pivii:
## Context

Captured during the #79 Smart Modes design session (2026-07-30).

Smart Modes v1 ships a fixed set of built-in modes. This issue covers the natural next step: letting the user write their own mode — a name, an icon, and their own instructions.

The motivation is concrete. Pierre tested Superwhisper's Email mode and Dictus Desktop's, and both disappoint in the same way: they invent greetings and sign-offs the user did not ask for. A mode that reformulates *someone's* email style well is hard to ship for everyone, because email style is exactly what varies per person. A custom mode moves that decision to the person who knows the answer.

Superwhisper ships this, and its own documentation states the flip side: instructions are empty by default, and without them "AI will not know what task to perform and may give unexpected results" (https://superwhisper.com/docs/modes/custom).

## Prerequisite already handled in v1

The #79 session decided that a Smart Mode is **data, not an enum case** — identifier, prompt, acceptance contract, display name, icon, pinned flag, stored in the App Group. v1 ships only built-ins, but they are already data. So this issue adds an editor, not a refactor.

## Scope

- [ ] An editor in DictusApp: name, icon, instructions
- [ ] Custom modes appear in the same list as built-ins and can be pinned to the keyboard fan (4 slots total, shared with built-ins)
- [ ] Reasonable limits on the number of custom modes and instruction length (the instruction text competes with the input for the model's context window)
- [ ] Guidance in the editor, since an empty or vague prompt produces confusing output

## The consequence to accept explicitly

**A custom mode cannot have a meaningful acceptance contract.** Every built-in mode ships with one — expected length band, expected output language — and that is what the runtime guardrail checks (see #79). For arbitrary user instructions we know none of those things.

So a custom mode runs with the loosest possible guardrail: non-empty output, a very wide length ratio, no language check. This is the one Smart Mode path with no safety net, which is precisely why it was kept out of v1: the paid feature should not ship an unguarded path on day one.

## Open questions

- Does the user get to preview a custom mode on sample text before pinning it? (The app can run the engine on a fixture — the same thing `polish-harness` does off-device.)
- Are input/output examples worth supporting, as Superwhisper does, or is that too much for a phone-sized editor?
- Should a custom mode be able to declare its own length band, so it gets *some* guardrail?

## Related

- #79 — Smart Modes v1 (defines the mode data model this builds on)

PR #394, opened 2026-08-24
Smart Modes block B: the keyboard arms a mode (refs #79)

Pivii:
Block B of #79: the keyboard. Until this branch, `SmartModeStore.arm(_:)` had no caller in production and dictation behaved exactly as it did before block A. It has one now.

Targets `develop`. `refs #79` — close by hand.

## The two decisions this block owed

Both were settled with Pierre before any code, and recorded in [the #79 comment](https://github.com/getdictus/dictus-ios/issues/79#issuecomment-5394059056).

**The fan holds four entries, so `maximumPinnedModes` drops from 4 to 3.** #79 contradicted itself — its acceptance criterion said four modes were pinnable, its geometry paragraph measured four *entries* including Normal. The geometry paragraph wins because it is a measurement: from the real `KeyMetrics` values, four entries give 46.7 pt a row on an iPhone SE and five give 37.4 pt, under Apple's 44 pt minimum. This target is a blind release at the end of a downward drag, under the thumb covering the row, not a tap on a visible button. **The acceptance criterion in the issue body is amended to three.**

**A context overflow degrades per mode.** `.exceededContextBudget` is the only non-success decided before the engine is called, so no transformation was attempted and the wrong-transformation risk fail-closed exists to prevent is absent. But the floor is not equally safe for every mode: Notes inserts it (same words, same language, just not restructured — and Notes is the mode built for the long dictations that overflow around 4,500 characters), Translate cannot, because the floor *is* the wrong language. `SmartMode` carries the answer as a field rather than deriving it from the contract, so #269's custom modes have to state it. Every other non-success stays fail-closed for every mode. Both paths show a message.

## What ships

- **The gesture.** Long-press the mic, the keys are replaced by a vertical fan, drag to a row, release. The release arms the mode *and* starts the recording. Releasing back on the mic aborts silently.
- **The armed mode is visible in three places**: the mic pill turns `#8B5CF6`, the toolbar's centre slot names it at priority 4, and the recording overlay names it in the empty centre of its top bar — the last being what makes a sticky mode safe, since by then the mic is gone.
- **The centre slot is a priority table**, lifted out of the SwiftUI body into `ToolbarCentreSlot` in DictusCore, where it is tested. Six competitors, not five: #315's polish-unavailable notice sits between the suggestions and the armed mode's name, because when polish will not run the mode will not either.
- **Failures speak.** Three sentences, because three different things happen: too long and degraded, too long and refused, anything else. The mode name is a colon-label rather than the sentence's subject, which fixes the rough edge `DictationHandoff` flagged for this block — translate modes are called "→ EN", and "→ EN could not transform this text" does not read. The app's copy moved to the same shape.
- **Half of #391**: `recoverPendingIfNeeded` inserted the raw without consulting `pending.smartMode`, which under "→ EN" means typing the untranslated French. It was the third insertion site and the only one that skipped the rule, because it is the one nobody watches.

## Three defects the simulator found, once the fan was reachable

Worth reading before the diff, because two of them are about SwiftUI rather than about this feature.

1. **The gesture was destroyed at the moment it succeeded.** SwiftUI identity is positional, and the toolbar was built inside the `switch` over `KeyboardAreaMode` — so opening the fan changed the branch, rebuilt the toolbar, and took the long-press recogniser with it. Six consecutive runs opened the fan and highlighted no row. The toolbar now lives outside the switch. **Nothing about the layout moves**: still 52 pt in every presentation, `heightConstraint` untouched (#166), the same `GeometryReader`s below it — which now measure the area rather than the whole hosting view and no longer need their clamped subtractions.
2. **A release without movement produced nothing.** A `SequenceGesture` only delivers `onEnded` once its second gesture has been recognised, and a `DragGesture` that never receives a movement update never is. Holding the mic and letting go left the fan on screen over the keys. A second `DragGesture(minimumDistance: 0)` covers that case and only that case.
3. **A fan closed by its backstop left the gesture inert.** The flag was cleared only on release. It is cleared at the next press instead.

The backstop itself is new and defensive: SwiftUI does not call `onEnded` on a *cancelled* touch, and the fan hides the keys, so a fan that outlives its finger is a keyboard nobody can type on. Eight seconds without a drag update closes it.

## Verified on the simulator

iPhone 17, iOS 26.5, Full Access granted, Dictus selected by long-pressing the globe.

| | |
|---|---|
| Fan opens on long-press | ✅ `smartModeFanOpened entries=3`, hosting 52 → 268 |
| Drag arms the row under the finger | ✅ `smartModeArmed mode=notes`, then `mode=translate.en` |
| Release starts the recording | ✅ DictusApp came to the foreground |
| Drag back onto the mic aborts | ✅ `smartModeFanAborted highlight=none` → `to=keys` |
| Mic pill purple + centre slot names the mode | ✅ screenshot |
| Discovery hint when nothing is armed | ✅ screenshot |
| Idle backstop closes a stuck fan, gesture recovers | ✅ `smartModeFanTimedOut` then a fresh `smartModeFanOpened` |

`cd DictusCore && swift test` — 1189 tests, 0 failures. `swiftlint lint --strict` — 0 violations. Both targets build.

## What still needs a device

- The gesture under a real thumb: whether 46.7 pt rows are hittable blind on an SE, and whether 0.35 s is the right press duration.
- Haptics — the selection tick on crossing rows, the refusal on a disabled one.
- Memory with several modes cycled in one keyboard process. This is #79's own open item and it is not answerable in a simulator, where `physicalMemory` returns the Mac's.
- The failure messages, which need Apple Intelligence and a real overflow.
- A smoke test that the extension still launches — a green build has never proven that here.

## Not in this branch

- **#393 must land before this reaches a user.** Notes and Translate are unmeasured, and the harness cannot yet run a Smart Mode. #388 showed what that measurement is worth: a prompt that held on 190 calls was still cut on evidence nobody had beforehand.
- **#392, the Pro entitlement check.** Nothing here consults `FeatureGate`. Block C.
- **The in-app degraded notice.** The keyboard says the mode did not apply; DictusApp does not, because `handleError` is the only thing there that shows a sentence and it ends the dictation as a failure — which a dictation that produced text is not. Inventing a non-fatal notice surface belongs with the block that owns that target's UI.

## Summary by CodeRabbit

- **New Features**
  - Added long-press Smart Mode selection from the keyboard microphone, including fan navigation, highlighting, haptics, and mode indicators.
  - Added Smart Mode discovery hints, improved toolbar prioritization, badge icons, and expanded layout support.
  - Notes can return raw text when processing exceeds context limits while reporting the degraded result.
  - Added language-code badges for Translate modes.
  - Limited pinned Smart Modes to three.

- **Bug Fixes**
  - Improved Smart Mode failure messages and prevented insertion when no valid output is available.

- **Documentation**
  - Added guidance for reliably testing SwiftUI gesture interactions.


Pivii:
## Device validation 2026-08-24 — one regression found and fixed, `b20ea6e`

**The gesture had no threshold at all.** `5215704` opened the fan on touch-down, so a plain tap could no longer start a dictation: the maintainer had to tap once to dismiss the fan and again to record. The log is unambiguous — **21 mic taps, 21 fan opens, zero direct recordings.**

A `LongPressGesture`'s value means "a press is in progress", not "the press has lasted long enough", so `onChanged` delivers `.first(true)` the instant the finger lands. The 0.35 s was written down and never applied. Opening now happens on `.second`, which a `SequenceGesture` delivers only once its first gesture has *succeeded*.

**This corrects the PR description above.** It claims the gesture was verified on the simulator, and it was — but only for holds and drags, because every test batch began with a `sleep`. A short tap was never sent, so the one input that distinguishes "long press" from "any press" was the one input never tried. The simulator can tell them apart perfectly well; the test set was wrong, not the tool. Both cases now run: a short `axe tap` produces `keyboardMicTapped` with no `smartModeFanOpened` behind it, and a 3 s stationary hold opens the fan.

Also in `b20ea6e`: fan rows centre their icon and label as one group. The highlight capsule spans the full width and a label pinned to its left edge read as text sitting on a pill rather than as the pill's own label.

### What the device session confirmed

| | |
|---|---|
| Long-press opens the fan, haptics included | ✅ |
| Drag to a row arms it and starts the recording in one gesture | ✅ |
| Abort by returning to the mic | ✅ `smartModeFanAborted highlight=none`, six times |
| Plain tap records directly | ✅ after `b20ea6e` |
| Mic pill colour, overlay naming the mode | ✅ |
| Keyboard extension launches | ✅ |

16 Smart Mode dictations ran end to end: 9 Translate → EN all successful, 6 Notes with one guardrail rejection. **That rejection was the pipeline working** — the Notes prompt returned English bullets for a French dictation, the `sameAsInput` check caught it, nothing was inserted and the failure message fired. It is prompt drift on an unmeasured mode, filed as the first field evidence on #393.

### Still open on this PR

Two design questions from the same session, neither blocking a merge:

- the armed-mode purple against an app whose identity is grey, blue and white;
- whether the fan should become 2 columns × 3 rows, which costs width rather than row height and would carry #269's custom modes.

Both are with a design pass; whatever comes out of it is a follow-up, not a change to this branch.

Issue #397, opened 2026-08-24
Smart Modes cannot be armed under VoiceOver — the fan is gesture-only


Pivii:
Found during the #79 visual coherence pass (`claudedocs/79-smart-mode-visual-coherence.md`), outside the two questions it was asked.

## A VoiceOver user cannot arm a Smart Mode

Arming is a long-press on the mic pill followed by a drag onto a fan row and a release. Two things make that unreachable under VoiceOver:

- **VoiceOver owns the gesture.** It intercepts the long-press-drag before the app's recogniser sees it; the standard escape is a rotor action or an activation, and this gesture has neither.
- **The fan takes no touches by design.** `SmartModeFanView` sets `.allowsHitTesting(false)`, deliberately and correctly — the drag belongs to the mic's recogniser from the long press to the release, and a row that accepted a touch would break it. The consequence is that the rows are not accessibility elements either, so there is nothing to swipe to and nothing to activate.

So the fan is not a degraded experience under VoiceOver. It is absent.

## Why this is not only an accessibility issue

**It changes what the app's mode list is for.** #79 describes pinning as choosing which modes the keyboard's fan holds — a convenience. If the gesture is the only way to arm, then for a VoiceOver user there is no way to arm at all, and the app becomes the *required* arming surface rather than a pinning surface. That is a different screen with a different job, and `SmartModeListView` (on `feature/79-smart-modes-app`) does not do it today: it pins and orders, it does not arm.

The two branches have to agree on the answer, which is why this is filed rather than fixed in either.

## Also missing, and cheap

`AnimatedMicButton` has **no `accessibilityLabel` at all**. VoiceOver announces it as an unlabelled button in every state, in the app and in the keyboard. Whatever is decided above, the button needs a label and a value that names the armed mode.

## The decision this needs

Where a VoiceOver user arms a mode. Three shapes, and the first two are not exclusive:

- **An arming control in the app's mode list** — a row action, or a distinct "armed" selection separate from the pinned set. Makes the app the accessible path and leaves the gesture as the fast path.
- **An accessibility action on the mic pill** — a custom rotor action per pinned mode, so arming stays in the keyboard where the dictation happens. Closer to parity, more work.
- **Making the fan rows real elements when VoiceOver is running**, with the gesture bypassed in that case. Largest change, and it fights the design that makes the drag work for everyone else.

## Acceptance criteria

- [ ] A VoiceOver user can arm and disarm every pinned Smart Mode without performing a drag
[...]

Issue #396, opened 2026-08-24
The mic pill is near-invisible during transcription on a light keyboard (1.30:1)


Pivii:
Found during the #79 visual coherence pass (`claudedocs/79-smart-mode-visual-coherence.md`), outside the two questions it was asked.

## The measurement

`AnimatedMicButton.buttonFillColor` returns `.dictusAccentHighlight.opacity(0.5)` — `#6BA3FF` at 50% — for `.transcribing` and `.processing`. Composited over a light iOS keyboard backdrop (~`#D1D3D9`):

| | ratio |
|---|---|
| pill shape against the keyboard | **1.30:1** |
| its white `mic.fill` glyph against the pill | **1.95:1** |
| the same glyph on opaque `#3D7EFF` | 3.73:1 |

WCAG asks 3:1 for a non-text UI component. The shape is essentially invisible and the glyph inside it is barely better.

## Why it matters more than the number

This is the state the user looks at for the **longest wait in the product** — transcription plus, since #361, the LLM stage, which #267 exists because it can run 3-6 s and materially longer for a Smart Mode. The one moment the button most needs to say "something is happening" is the moment it fades out.

It is also the state a frustrated user taps. `isTappable` is false there, so the tap does nothing — a near-invisible button that ignores you is the worst possible reading of a wait.

Dark keyboards are unaffected; the problem is specific to a light host.

## Proposed fix

Fill with opaque `#3D7EFF` and let the existing shimmer sweep carry "in progress". The shimmer is already the state's signal — the 0.5 opacity adds nothing that the motion does not, and costs the shape its edges.

The ring is a separate question and can stay.

## Acceptance criteria

- [ ] The pill's shape clears 3:1 against both a light and a dark keyboard in `.transcribing` and `.processing`
- [ ] The glyph stays legible in both
- [ ] The shimmer still reads as motion against the new fill
- [ ] The success flash and the recording red are unchanged

Pivii:
Closing: the scene this issue measures does not exist.

**The mic pill is never rendered in `.transcribing` or `.processing` inside the keyboard.** `DictationStatus.ownsKeyboardArea` (`KeyboardAreaMode.swift:73`) returns `true` for those two statuses — `processing` added deliberately in #267 so the overlay holds through the LLM stage — and `KeyboardRootView.swift:247` replaces the entire area, toolbar included:

```swift
if presentedMode == .recording {
    // Recording overlay fills the full area (toolbar + keyboard space).
[...]

Issue #79, opened 2026-04-08
Smart Mode — Local LLM integration (Apple Foundation Models + open-source)


Pivii:
## Context

Part of the Dictus Pro milestone (v1.5). This is the **killer feature** of the Pro tier — intelligent reformulation of dictated text using on-device LLM.

See #54 for the full premium strategy.

## Goal

After a user dictates text, they can apply a "Smart Mode" that reformulates it using a local LLM. Everything stays on-device — no cloud, no data leaves the phone.

### User flow

1. User dictates text (existing flow) → raw transcription appears
2. User taps a Smart Mode template (e.g., "Email", "SMS", "Notes")
3. LLM reformulates the text according to the template
4. Reformulated text replaces or is offered alongside the original

### Pre-configured templates (v1.5)

| Template | What it does |
|----------|-------------|
| **Email** | Formal tone, greeting + body + sign-off structure |
| **SMS** | Casual, concise, keeps the intent |
| **Notes** | Bullet points, structured, removes filler |
| **Summary** | Condenses long dictation into key points |

### LLM Backends (priority order)

1. **Apple Foundation Models** (iOS 26+, iPhone 15 Pro+)
   - Free, no download needed, Apple manages the model
   - Best option when available — fast, integrated, no memory overhead
   - Limitation: requires iPhone 15 Pro or later + iOS 26

2. **Open-source models** (fallback for older devices)
   - Gemma 4 (Google), Mistral small, or similar compact models
   - User downloads the model (like they download Whisper models today)
   - Runs via Core ML or similar on-device inference framework
   - Needs research: which models fit in memory, inference speed, quality

## Architecture considerations

- **Model Manager**: extend existing WhisperModelManager pattern for LLM models (download, select, delete)
- **Smart Mode engine**: new module in DictusApp (not keyboard extension — too memory-heavy)
- **Template system**: structured prompts per template, with language awareness (FR/EN)
- **Feature gating**: behind `SubscriptionManager.isPro` (free during beta)

## Memory budget

- Apple Foundation Models: managed by iOS, no impact on our memory
- Open-source models: typically 1-4 GB on disk, ~500MB-2GB RAM during inference
- This runs in DictusApp (not the 50MB keyboard extension)
- Need to profile actual memory usage and gate by device capability

## Research needed

- [ ] Apple Foundation Models API availability and capabilities (iOS 26)
- [ ] Which open-source models work well on-device for text reformulation in French
[...]

PR #398, opened 2026-08-24
Armed Smart Mode: ring instead of fill (refs #79)

Pivii:
**Stacked on #394** — it targets `feature/79-smart-modes-keyboard`, not `develop`, so the diff here is only the colour change. Merge after #394.

Follows the visual coherence pass in `claudedocs/79-smart-mode-visual-coherence.md`, written after the 2026-08-24 device session.

## The observation, and its one cause

The purple armed pill is genuinely useful — you can see at a glance that a mode is armed — and it reads as foreign in an app whose identity is grey, blue and white. Both halves are right.

The armed pill was **2,016 pt² of full-saturation `#8B5CF6`, permanently**, because the mode is sticky. Every other purple in the feature is already quiet:

| Surface | Purple dose |
|---|---|
| Fan row highlight | `#8B5CF6` @ 0.22 capsule, purple label |
| Toolbar armed label | 13 pt purple text + `sparkles` |
| Overlay armed badge | 14 pt purple on `#8B5CF6` @ 0.14 |
| **Mic pill, armed** | **`#8B5CF6` @ 1.0, 2,016 pt², permanent** |

One surface out of four was shouting. It is a dose problem, not a token problem.

## What changes

The signal moves from the pill's **fill** to the pill's **ring**, which already exists — `AnimatedMicButton` draws a 66 × 46 capsule behind the 56 × 36 body. A 3 pt stroke is roughly 300 pt²: same token, same meaning, about a seventh of the ink.

It also lands the signal on the **edge**, which is where `SmartModeFanView` already argues it must be for this feature — the thumb that armed the mode is sitting on the pill.

**Armed rings do not breathe.** The idle pulse says "tap me", which is about the button; armed is a statement about a setting. Colour and motion then both separate the two states, which matters when colour alone cannot be relied on.

**Opaque, and this is measured rather than chosen.** At the 0.3 end of the idle breath, `#8B5CF6` composited over a light keyboard is **1.35:1** — invisible. At full opacity it is **2.83:1** there and **4.23:1** against white, better than `#3D7EFF`'s 3.73:1. Purple was never the contrast problem; a breathing purple ring would have been.

`tint` becomes `ringTint`, and `buttonFillColor` stops consulting it.

## What this rejects, and why

**Grey mic for Normal, blue for Smart Mode.** Three reasons, descending:

1. Dictus already renders a faded mic and it means *"you cannot dictate"* — `fullAccessBar` at `.opacity(0.4)`, and `MicButtonDisabled.swift` exists for it. A grey resting mic sits one visual step from the disabled one, in the state 100% of users are in before they buy Pro. The default must not look broken.
2. The mic is the reference point the rest of the bar is calibrated against. `DictusColors.dictusPillIconSecondary` says so in its own comment: the other icons are deliberately 50% *because* they must not compete with the mic. Removing its colour does not make it quieter, it removes the calibration.
3. It inverts the meaning. Blue would go from "ready to dictate" to "ready to dictate differently", making the saturated state the rarer one — backwards for a signal whose job is "this is the control you came here for".

**Per-mode blues.** Two blues are only distinguishable side by side, and these never are: one pill at a time, seen peripherally, through a translucent keyboard, over an arbitrary host. `#3D7EFF` and `#2563EB` differ by 0.10 in relative luminance. The mode's *name* is already on screen at priority 4 and in the recording overlay, and names scale to any catalogue size.

**Recording red, anywhere.** Untouched in both modes.

## Verification

Simulator screenshot: blue pill, purple ring, `✦ → EN` in the centre slot. 1189 tests, 0 failures. `swiftlint lint --strict` clean.

**Needs a device look**, and it is the whole point of the change: whether ~300 pt² of ring is still legible at a glance in peripheral vision, which is the job the 2,016 pt² fill was doing. If it is not, the honest fallback is a thicker ring rather than a return to the fill.

## Not here

The document raised two findings outside its brief, both filed: #396 (the transcribing pill measures 1.30:1 on a light keyboard, and its glyph 1.95:1) and #397 (Smart Modes cannot be armed under VoiceOver at all). Neither is in this PR.

### DIFF
--- a/DictusCore/Sources/DictusCore/Design/AnimatedMicButton.swift
+++ b/DictusCore/Sources/DictusCore/Design/AnimatedMicButton.swift
@@ -22,29 +22,55 @@ public struct AnimatedMicButton: View {
     public let status: DictationStatus
     public let isPill: Bool
 
-    /// The button's resting colour, and the colour of its idle glow.
+    /// The colour of the button's **ring**, when the status has not claimed it.
     ///
     /// A parameter since #79: the keyboard's pill signals the armed Smart Mode by
     /// turning `.dictusSmartMode` purple, which costs zero width in the most
     /// contested 32 pt of the UI and stays visible while the user is typing.
     ///
-    /// It only ever replaces the resting colour. Recording stays red and the two
-    /// post-recording stages stay in their washed accent, because those describe what
-    /// the phone is doing right now and outrank a setting — a red mic must mean
-    /// recording on every screen of this product.
-    public let tint: Color
+    /// ### Why the ring and not the fill
+    ///
+    /// It was the fill, and that is what a device session found wrong (2026-08-24).
+    /// The armed pill was 2,016 pt² of full-saturation purple, permanently, because
+    /// the mode is sticky — while every other purple in the feature is already quiet:
+    /// a 0.22 capsule behind a fan row, a 0.14 capsule behind the overlay badge, 13 pt
+    /// of label text. One surface out of four was shouting, and it read as foreign in
+    /// an app whose identity is grey, blue and white.
+    ///
+    /// A 3 pt stroke around the 66 × 46 ring is roughly 300 pt² — the same token and
+    /// the same meaning at about a seventh of the ink. It also puts the signal on the
+    /// **edge**, which is where `SmartModeFanView` already argues the signal has to
+    /// be: the thumb that armed the mode is sitting on the pill.
+    ///
+    /// Removing purple altogether was considered and rejected. `#8B5CF6` measures
+    /// 4.23:1 against white — better than `#3D7EFF`'s 3.73:1 — so it is not a
+    /// contrast problem, and the brand kit names this colour "Smart mode". The dose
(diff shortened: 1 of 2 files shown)


Pivii:
Rejected on device, 2026-08-24.

The purple ring reads as an outline stuck onto the pill rather than as a signal, and the maintainer's position on purple hardened rather than softened under the change: it should leave the feature entirely, including the toolbar centre-slot label this PR never touched.

`claudedocs/79-smart-mode-visual-coherence.md` was right that the problem was the **dose** and wrong in its conclusion. Less purple is not the answer; a different signal is. Whoever reopens this should start from that constraint rather than from the document's recommendation.

Also reverted on `feature/79-smart-modes-keyboard` (`9d1f4bf`): the fan backdrop from the same pass. `dictusGlass(in: Rectangle())` produced a lighter block with hard edges above and below — it sat *on* the keyboard instead of being part of it. The host-text-showing-through problem it was fixing is real and still open; the fix is not an opaque panel.

Issue #408, opened 2026-08-25
Swap Turbo to large-v3-v20240930_turbo_632MB: better accuracy, 39% smaller, same iPhone gate


Pivii:
## Context

Acceptance A of #171 is done: four Whisper variants and Parakeet v3 measured on real French dictation on 2026-08-25. Full corpus, verbatim outputs and caveats in https://github.com/getdictus/dictus-ios/issues/171#issuecomment-5409496092.

The variant we ship, `openai_whisper-large-v3_turbo_954MB`, won no clip and hallucinated on three of five. `openai_whisper-large-v3-v20240930_turbo_632MB` won or tied on four of five.

| | shipped `_954MB` | candidate `_632MB` |
|---|---|---|
| Clips won or tied (of 5) | 1 | 4 |
| Hallucinations | 3 clips (`谢谢`, `Merci.`, `j'ai un goy qui switch in english`) | 1 clip (`Merci.`) |
| French sentence with English loanwords | translated to English | transcribed correctly |
| Download size | 1,052,848,880 B | **645,668,913 B** |
| iPhone gate in `argmaxinc/whisperkit-coreml/config.json` | iPhone15-18, iPad15,7-8 | **identical** |
| RTF on iPhone | 2.70x (#171, measured) | **unmeasured** |

Both variants ship the same four artefacts (`AudioEncoder.mlmodelc`, `TextDecoder.mlmodelc`, `MelSpectrogram.mlmodelc`, `TextDecoderContextPrefill.mlmodelc`), so `directoryPatterns` and `requiredPaths` need no new shape.

The 39% size cut also shortens Core ML compilation, which lightens the timeout in #406 without touching the budget. It does not replace #406.

## Fix

In `DictusCore/Sources/DictusCore/ModelInfo.swift`, replace the Turbo entry:

- `identifier`: `openai_whisper-large-v3-v20240930_turbo_632MB`
- `sizeBytes`: `645_668_913` — the exact recursive byte sum of `<identifier>/` as served, computed the same way as every other entry (the same method reproduces the existing `1_052_848_880` constant exactly, which is how it was validated)
- `displayName`: `Turbo` (unchanged)
- `accuracyScore` / `speedScore`: see below
- The Phase 37 comment block explaining why the non-quantized variant was the historical root cause stays true and stays; extend it with why this variant replaces the `_954MB` one, citing the #171 comment.

Then follow the identifier through every site that hard-codes the old string. At minimum:

[...]

Issue #406, opened 2026-08-25
The 120s prewarm timeout was calibrated on Parakeet and fires on Turbo, whose own documented compile is ~2 min


Pivii:
## Symptom

TestFlight feedback, 2026-08-25 09:28 CEST. iPhone 15 Pro Max (iPhone16,2), iOS 26.6, mobile data, 5.3 GB free disk.

> Turbo model download successfully completed, but optimizing the model aborts after 120 seconds. Please consider increasing the timeout to 240 seconds, if any.

User-facing error: `Model optimization did not complete within 120s`.

## The codebase already contradicts itself on this number

`DictusApp/Models/ModelManager.swift:288`:

```swift
// 120s is well above the observed normal worst case (~17s for Parakeet
// Encoder on this device) so we don't false-positive on legitimately
// long prewarms.
let prewarmTimeoutSeconds = 120
```

`DictusApp/Views/ModelLoadingOverlay.swift:10`:

> Whisper turbo (~954 MB) takes **~2 min** of one-off Core ML compilation **on a 15 Pro Max**

The guard was calibrated against Parakeet's ~17 s and never re-calibrated for Turbo. It now sits exactly on Turbo's nominal compile time on the very device the comment names. Any variance pushes it over: 98% full disk, thermal throttling, a newer iOS. The reporter hit all three conditions plausibly and got the timeout on a top-tier device.

Introduced in `b3df680` (#104, Phase 37).

## Why not simply raise the global value to 240s

A single global timeout has to serve two opposite jobs:

- Turbo on a fast device: a legitimately slow compile that must be allowed to finish.
- Whisper Small on an unsupported A13 (#362): a compile that will never finish, where the guard is the only thing that stops an infinite spinner.

Raising the global value to 240 s makes every #362-class device wait twice as long before being told it failed.

## Fix

Move the timeout into the model catalogue as a per-model value.

- Add a `prewarmTimeoutSeconds` field to `ModelInfo` (`DictusCore/Sources/DictusCore/ModelInfo.swift`), defaulting to 120.
- Set `openai_whisper-large-v3_turbo_954MB` to 300 (2.5x its documented ~2 min, so the guard catches a genuine hang without firing on a slow-but-real compile).
- `ModelManager` reads the value from `ModelInfo` instead of the local constant.

Leave every other model at 120 s. Medium (1.53 GB) has not been reported as timing out, but it is the next candidate if evidence appears; do not change it speculatively.

## Acceptance

[...]

Issue #405, opened 2026-08-25
A prewarm timeout deletes the downloaded model, so Retry re-downloads the whole payload


Pivii:
## Symptom

TestFlight feedback, 2026-08-25 09:28 CEST. iPhone 15 Pro Max (iPhone16,2), iOS 26.6, de-AT, **mobile data (5G)**, 5.3 GB free disk.

> Turbo model download successfully completed, but optimizing the model aborts after 120 seconds.

Screenshots show `Turbo — OPTIMIZING / Checking the model...` at 09:17, then `Error — Model optimization did not complete within 120s` at 09:23. The gap suggests at least one retry.

## Root cause

`DictusApp/Models/ModelManager.swift`, in the `catch` of `downloadWhisperKitModel`:

```swift
if downloadPhaseCompleted {
    // Prewarm failure: clean up. ANE compilation failures (E5 bundle
    // errors) leave behind unusable cached files that prevent
    // re-download from working correctly.
    cleanupModelFiles(identifier)
}
```

The cleanup is correct for an **E5 bundle compilation failure**: those leave a corrupt Core ML cache and a retry in place would fail identically. It is wrong for a **timeout**, where the downloaded files are intact and only the compile ran out of clock.

The code already distinguishes the two cases a few lines above, purely to log:

```swift
} catch let err as ModelManagerError {
    if case .prewarmTimeout(let s) = err {
        PersistentLog.log(.modelPrewarmTimeout(name: identifier, timeoutSeconds: s))
    }
    throw err
}
```

So the discrimination exists; the `catch` just does not act on it.

## Impact

Turbo is `sizeBytes: 1_052_848_880` (1.05 GB). Every timeout costs the user a full 1 GB re-download. The reporter was on cellular. This also compounds #362, where an iPhone 11 loops on Whisper Small: each loop pays the download again.

## Fix

Skip `cleanupModelFiles` when the failure is `ModelManagerError.prewarmTimeout`. Leave the cleanup in place for every other prewarm failure. The download-failure path already has a no-cleanup policy for the same reason (#210), so this aligns the prewarm path with it.

The existing `Retry` affordance on the model card (#235) then resumes at the compile step instead of restarting the download. The overflow menu keeps `Delete partial download` for the case where a user does want a clean slate.

## Acceptance

- [ ] A prewarm timeout leaves `<modelsDirectory>/<identifier>` and the WhisperKit repo directory on disk
[...]

Issue #362, opened 2026-08-20
iPhone 11 onboarding loops or crashes while optimizing unsupported Whisper Small


Pivii:
## Environment

- App version/build: unknown TestFlight build, confirmation requested
- Revision/branch: unknown
- iOS: unknown, confirmation requested
- Device: iPhone 11, A13, 4 GB RAM
- Model: `openai_whisper-small`

## User-visible failure

During first-run onboarding, the model download completes and the blocking preparation screen remains on:

- Model: `Small`
- Phase: `OPTIMIZING`
- Rotating copy such as `A few more seconds...`

The user cannot leave onboarding. The app sometimes crashes during this phase. Deleting and reinstalling the app reproduces the same cycle.

There is no close, skip, fallback, retry, or log-export control on this blocking screen.

## Root cause

Dictus currently chooses the onboarding model only from physical RAM:

```swift
capabilities.physicalMemoryGB >= 6
    ? "parakeet-tdt-0.6b-v3"
    : "openai_whisper-small"
```

An iPhone 11 has 4 GB RAM, so Dictus always selects Whisper Small.

However, Argmax's authoritative `whisperkit-coreml/config.json` support matrix lists A12/A13 iPhones, hardware families `iPhone11,*` and `iPhone12,*`, as supporting only:

- `openai_whisper-tiny`
- `openai_whisper-tiny.en`
- `openai_whisper-base`
- `openai_whisper-base.en`

Whisper Small is not supported for this device tier. The screenshot is specifically past download and inside `WhisperKit(config)` prewarm/Core ML compilation, which matches the unsupported-model failure boundary.

Source: https://huggingface.co/argmaxinc/whisperkit-coreml/raw/main/config.json

## Why reinstalling cannot help

A fresh install repeats the same deterministic recommendation from the same hardware identifier and RAM tier. It downloads Small and reaches the same Core ML optimization step again.

## Immediate fix

- [ ] Recommend multilingual Whisper Base on A12/A13 iPhones instead of Small.
- [ ] Add regression coverage for real identifiers such as `iPhone12,1` (iPhone 11) and `iPhone11,2` (iPhone XS).
- [ ] Keep Small for A14-class 4 GB devices where Argmax lists it as supported.
- [ ] Gate unsupported Whisper variants from device-specific model selection, rather than gating Turbo by RAM only.

## Recovery and diagnostics follow-up

- [ ] A blocking onboarding preparation screen must eventually offer a safe fallback or retry after failure.
[...]

PR #421, opened 2026-08-25
Give each model its own Core ML prewarm budget (refs #406)

Pivii:
Targets **`integration/model-turbo-timeout`**, not `develop`. The base already carries PR #407 (#405) and PR #410 (#408); this is the third of the three and they are only meaningful together on device.

refs #406

## What this was for

A TestFlight tester finished downloading Turbo and then lost it: the Core ML optimization step was cut off by a 120 s guard, and because #405 was not in that build, the ~1 GB he had just pulled over mobile data was deleted with it.

That 120 s was never a Turbo number. It was calibrated in Phase 37 (#104) against the ~17 s a *Parakeet Encoder* compile took on an iPhone 15 Pro Max, and never revisited — which left it sitting almost exactly on Turbo's own documented ~2 min compile time on that very device. Any variance tips it over.

On 2026-08-25 the maintainer hit the same guard on his own iPhone 15 Pro Max, using the new 39%-smaller variant from #408. That is what settles it: the size cut alone does **not** bring the compile inside the old budget, so #406 is a prerequisite for #408 rather than something #408 softens.

## To test by hand

This is a combined build — #405 + #406 + #408 in one pass. Device only: Core ML ANE compile timing cannot be reproduced on a simulator, so nothing below is covered by CI.

Before you start: delete Turbo from Settings so the download starts clean.

1. Settings → Models → tap **Turbo**. Expect the download to announce and pull **~645 MB**, not ~1 GB. (That is #408.)
2. Watch the card flip to **OPTIMIZING**. Expect it to sit there for **longer than two minutes** — that is the point of the change, not a hang.
3. Expect it to **finish** and the card to reach **Ready**, with Turbo now the active model. Previously this step died at 120 s with `Model optimization did not complete within 120s`.
4. Export the debug log and find the line `modelCompilationCompleted name=openai_whisper-large-v3-v20240930_turbo_632MB durationMs=…`. **Send me that `durationMs`.** It is the first time anyone will have measured a Turbo compile on device — the guard has always cut it short before, on both variants. It goes back on issue #406 so the constant stops being a guess.
5. Dictate a sentence and confirm the transcription comes back normally on Turbo.
6. **If it still times out** at step 3: the error will now read `within 300s`, not `within 120s` — that is how you know this branch is the one running. Then check the card: it must offer **Retry**, and tapping Retry must go **straight back to OPTIMIZING with no progress bar and no download**. That is #405 working: the 645 MB stays on disk and a retry costs only time. Send me the log line `modelPrewarmTimeout name=… timeout=300s` and whether a `modelCleanupPerformed` line follows it (it must not).
7. Optional, if you have a device still holding the old 1.05 GB Turbo: install and confirm dictation keeps working with no surprise download. It shows as **Turbo (Legacy)** under Downloaded.

## What changed and why

**The budget moved into the catalogue, per model.** `ModelInfo` gains `prewarmTimeoutSeconds`, with `ModelInfo.defaultPrewarmTimeoutSeconds = 120` that every entry inherits by omission. `ModelManager.downloadWhisperKitModel` resolves the model's own value instead of a local constant.

**The global value was deliberately not raised.** One number has to serve two opposite jobs: Turbo needs a long compile allowed to finish, while Whisper Small on an unsupported A13 (#362) never finishes at all, and there the guard is the only thing that ends an infinite spinner. Doubling the global punishes the second to serve the first.

**Turbo declares 300 s, and that is a bounded guess, not a measurement.** Nobody has ever seen a Turbo compile finish on device — the guard always cut it first, on both variants — so the real duration is unknown. The only figure in existence is the "~2 min on a 15 Pro Max" prose in `ModelLoadingOverlay.swift`, and that describes the `_954MB` variant #408 replaced. 300 is 2.5× it. The catalogue comment says all of this in the entry itself, including the instruction not to shrink 300 to hug the first reading that arrives (one device, one thermal state, one disk-pressure condition; the TestFlight reporter had 5.3 GB free of 254 GB).

**The fix produces its own replacement measurement.** `downloadWhisperKitModel` already logs `modelCompilationCompleted(name:durationMs:)` on success; nothing here touches that path. The first compile to complete under the wider budget is the first real reading of a Turbo compile.

**The stale comment was rewritten, not left standing.** The `120` in `ModelManager.swift` carried a comment calibrating it against Parakeet's ~17 s — the exact thing this issue is about. It is replaced by the reasoning above rather than left next to a value it no longer describes.

## Acceptance criteria

|---|---|---|
| `ModelInfo` carries a per-model prewarm timeout, default 120 | Done | `ModelInfo.swift:78` `public let prewarmTimeoutSeconds: Int`; `:89` `defaultPrewarmTimeoutSeconds = 120`. `testEveryNonTurboModelKeepsTheDefaultBudget` asserts all six non-Turbo entries inherit it. |
| Turbo's entry declares 300 | Done | `ModelInfo.swift:331` (`_632MB`) and `:368` (`_954MB`). `testTurboBudgetExceedsItsDocumentedCompileDuration` passes. |
| `ModelManager` reads it from the catalogue; the hard-coded `120` is gone | Done | `ModelManager.swift:309-310` reads `ModelInfo.forIdentifier(identifier)?.prewarmTimeoutSeconds ?? ModelInfo.defaultPrewarmTimeoutSeconds`. `grep -rnI 'prewarmTimeoutSeconds *= *[0-9]' DictusApp DictusKeyboard DictusCore/Sources` returns nothing. |
| `.modelPrewarmTimeout` still logs the timeout actually applied | Done | Single-source chain, unchanged in shape: the resolved value goes to `withPrewarmTimeout(seconds:)`, which throws `.prewarmTimeout(seconds: seconds)` with that same value; the `catch` binds it as `s` and logs `timeoutSeconds: s`. There is no second constant that could disagree. The user-facing string (`"…did not complete within \(seconds)s"`) reads the same source, so it will now say `300s` for Turbo. Argued from the diff — a real log line only exists after a device run (step 6 above). |
| A unit test in `DictusCore` asserts Turbo's timeout exceeds its documented compile duration | Done | `testTurboBudgetExceedsItsDocumentedCompileDuration`, written as `> documentedCompileSeconds (120)` rather than only `== 300` so it survives the number being re-derived from a real measurement. Negative control run: with the catalogue reverted to 120 it fails (`XCTAssertGreaterThan failed: ("120") is not greater than ("120")`); at 300 it passes. |

### Verification run

- `cd DictusCore && swift test` → `Executed 1221 tests, with 1 test skipped and 0 failures`. Base was 1216; the 5 added here are the difference, and no existing test regressed.
- `swiftlint lint --strict` → `Found 0 violations, 0 serious in 216 files` (green before and after).
- `xcodebuild build -scheme DictusApp -destination 'platform=iOS Simulator,id=1720B34A…' -derivedDataPath build/DerivedData` → `** BUILD SUCCEEDED **`, after `./scripts/patch-fluidaudio-swift5.sh build/DerivedData` (#285).
- Simulator work stayed headless: an already-booted iPhone 17 Pro, no window raised.

## Judgment calls and things to push back on

**The deprecated `_954MB` entry also got 300, which is one step past the brief's "leave every other model at 120".** It is unreachable in practice — `available(on:)` no longer offers it, so no download and therefore no prewarm can start for it. But it is the exact model the TestFlight report timed out on at 120 s, so declaring 120 next to it would leave the catalogue asserting a budget the field has already disproved for that model. Say the word and I will drop it back to the default.

**The Parakeet path still has no timeout at all** — not a regression, it has never had one. `downloadParakeetModel` calls `parakeetEngine.prepare()` with no deadline guard, so a FluidAudio compile that hangs hangs forever. I deliberately did not add one: it is outside #406's acceptance, and introducing a new failure mode on the onboarding default model is not something to do as a side effect. Its catalogue value is therefore declarative-only today, and the field's doc comment says so. Worth its own issue.

**Why 300 s is safe to declare even though #362 exists**: both Turbo variants are gated out on A12/A13 by the Argmax support matrix and on sub-6 GB devices by RAM, so a #362-class iPhone 11 can never be handed the wider budget — it keeps 120 s on Small. `testTheWiderBudgetIsUnreachableFromTheDevicesThatNeedTheShortOne` pins that, so a future gating change cannot silently hand a five-minute spinner to the devices that most need a short one.

**Untested here**: that the compile actually completes within 300 s. Nobody knows what a Turbo compile costs, which is the whole point. If the device run at step 3 still times out, 300 was too low and the reading from step 6 tells us by how much.

Head sha: `0d8828667d53903493540df0e7f7f313530fa45d`

https://claude.ai/code/session_0155NfNULhoE24Yz5dGPjFTo

### DIFF
--- a/DictusApp/Models/ModelManager.swift
+++ b/DictusApp/Models/ModelManager.swift
@@ -294,9 +294,22 @@ class ModelManager: ObservableObject {
             // download to it on 2026-08-25, and the maintainer reproduced it the same
             // day on the smaller variant from issue #408. Raising the global instead
             // would have punished the opposite case: Whisper Small on an unsupported
-            // A13 (issue #362) never finishes compiling, and this guard is the only
-            // thing that ends that spinner. `ModelInfo.prewarmTimeoutSeconds` lets the
-            // two models disagree.
+            // A13 (issue #362) takes far longer than it has any business taking.
+            // `ModelInfo.prewarmTimeoutSeconds` lets the two models disagree.
+            //
+            // WHAT THIS GUARD DOES NOT DO (issue #427): it does not interrupt the
+            // compile. `withPrewarmTimeout` races a sleep against `WhisperKit(config)`
+            // in a task group, and a task group cannot return until every child has
+            // finished — `cancelAll()` is a request, and a Core ML compile neither
+            // checks cancellation nor offers a suspension point where it could. So the
+            // error is thrown on time and surfaces only once the compile has finished
+            // anyway. Measured 2026-08-26: a 5s budget reported failure after 212s, on
+            // a compile that had by then completed and warmed the cache.
+            //
+            // So do not read this budget as protection against a hang. It bounds
+            // nothing; it reports, afterwards, that the work took longer than a number.
+            // #362 in particular is NOT protected by it, whatever an earlier version of
+            // this comment claimed.
             //
             // Whatever this resolves to is the number that reaches the user: it is
             // carried by the thrown `.prewarmTimeout(seconds:)` into both the
--- a/DictusApp/Models/ModelManager.swift
+++ b/DictusApp/Models/ModelManager.swift
@@ -284,10 +284,30 @@ class ModelManager: ObservableObject {
             // Phase 37: guard the CoreML prewarm against indefinite hangs.
             // On iPhone ANE, some WhisperKit model variants fail to compile and the
             // async init never returns (E5 bundle load failure — issue #104,
-            // 2026-04-22 iPhone 15 Pro Max). 120s is well above the observed normal
-            // worst case (~17s for Parakeet Encoder on this device) so we don't
-            // false-positive on legitimately long prewarms.
-            let prewarmTimeoutSeconds = 120
+            // 2026-04-22 iPhone 15 Pro Max).
+            //
+            // WHY the number comes from the catalogue (issue #406):
+            // it used to be a flat 120 written here, justified by the ~17s a Parakeet
+            // Encoder compile took on a 15 Pro Max. That was a reading from a different
+            // engine, it was never revisited, and it ended up sitting almost exactly on
+            // Turbo's own compile time — so a TestFlight tester lost a completed Turbo
+            // download to it on 2026-08-25, and the maintainer reproduced it the same
+            // day on the smaller variant from issue #408. Raising the global instead
+            // would have punished the opposite case: Whisper Small on an unsupported
+            // A13 (issue #362) never finishes compiling, and this guard is the only
+            // thing that ends that spinner. `ModelInfo.prewarmTimeoutSeconds` lets the
+            // two models disagree.
+            //
+            // Whatever this resolves to is the number that reaches the user: it is
+            // carried by the thrown `.prewarmTimeout(seconds:)` into both the
+            // `.modelPrewarmTimeout` log line and the error text on the model card, so
+            // the debug log always states the budget that was actually applied.
+            //
+            // The fallback covers an identifier the catalogue does not know, which this
+            // path should never see — `downloadModel` had to resolve the model to route
+            // it here by engine in the first place.
+            let prewarmTimeoutSeconds = ModelInfo.forIdentifier(identifier)?.prewarmTimeoutSeconds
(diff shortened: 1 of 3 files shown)


Pivii:
Merged into `develop` via the integration branch (84d94ac). Closing here because the PR targeted `integration/model-turbo-timeout`, which has served its purpose.

Issue #428, opened 2026-08-26
A model that never finishes compiling locks the user out of the whole app


Pivii:
## What happens

On an iPhone 15 Pro Max (iOS 26.6) running `c74a649`, opening Dictus lands on the model loading screen and stays there. No settings, no model list, no way to pick a different model. The app is unusable. Force quitting and reopening returns to the same screen. Reinstalling does not help, because the state lives in the App Group container, which survives an install.

The maintainer hit this while testing Turbo and could not even export a debug log, because the export lives behind the screen he cannot leave.

## Measured on the device

Read remotely from the device container, not inferred:

- `dictus.activeModel` = `openai_whisper-large-v3-v20240930_turbo_632MB`
- The model is **fully downloaded**: all four `.mlmodelc` bundles present, 615 MB of weights, `config.json` present. Downloading is not the failure.
- Turbo has **no `*.mlcomputeplan.json`**, for any of its modules. `openai_whisper-medium`, in the same container, has all three. Core ML has never completed a compile for this variant on this device.
- App Group state holds `dictus.modelLoadState = "loading"` and `dictus.modelReady = true` at the same time.
- No crash report for DictusApp, and the jetsam log at 10:23 today shows DictusApp at 5903 pages (~94 MB) and `backboardd` as the largest process. It is not being killed. It does not finish.

Rewriting `dictus.activeModel` to `openai_whisper-medium` from the Mac while the app was terminated wrote successfully, and the app restored `turbo_632MB` on the next launch: iOS `cfprefsd` still held the cached value.

## Why it is a trap and not a slow load

Four independent code facts, each read in the source. Together they form a self-sustaining loop.

**1. The screen replaces the app rather than covering it.** `DictusApp/Views/MainTabView.swift:66-115` is a three-way `if`, not an overlay. When `preparingModelID != nil`, the `TabView` is never constructed, so Settings and the model list do not exist to be reached. `ModelLoadingOverlay.body` (`DictusApp/Views/ModelLoadingOverlay.swift:66-125`) contains no `Button`, no `NavigationLink`, no gesture and no timeout. Its only exit is `setModelLoadState` being called with `.ready` or `.idle`.

[...]

Issue #427, opened 2026-08-26
The prewarm timeout cannot interrupt the compile it guards: a 5s budget reported failure after 212s


Pivii:
## Measured

iPhone 15 Pro Max, iOS 26.6, throwaway build `72fe5b9@HEAD` with Turbo's `prewarmTimeoutSeconds` forced to **5**, 2026-08-26:

```
09:19:53  modelCompilationStarted  name=…turbo_632MB
09:23:25  modelPrewarmTimeout      name=…turbo_632MB timeout=5s      <-- 212 s later
09:23:25  modelDownloadFailed      error=Model optimization did not complete within 5s
```

A 5-second budget produced its timeout error **3 minutes 32 seconds** after the work started — approximately the full cost of the compile it was supposed to cut short (measured at 236 s in #406).

## Why

`ModelManager.swift`:

```swift
@MainActor
func withPrewarmTimeout<T: Sendable>(seconds: Int, _ operation: @escaping @Sendable () async throws -> T) async throws -> T {
    try await withThrowingTaskGroup(of: T.self) { group in
        group.addTask { try await operation() }
        group.addTask {
            try await Task.sleep(nanoseconds: UInt64(seconds) * 1_000_000_000)
            throw ModelManagerError.prewarmTimeout(seconds: seconds)
        }
        // First to finish wins. Cancel the other before returning.
        let result = try await group.next()!
        group.cancelAll()
        return result
    }
}
```

**A task group cannot return until every child task has finished.** `cancelAll()` is a cooperative *request*; a child that never checks `Task.isCancelled` and never hits a cancellation-aware suspension point simply keeps running. `WhisperKit(config)` performs a long Core ML compile that does neither.

So the sleep child throws on schedule at 5 s, and the throw then waits, inside the group's implicit drain, for the compile child to finish anyway. The error surfaces only once the thing it was guarding has already completed.

The comment shows the author expected `cancelAll()` to end the operation. It does not, for this operation.

## What follows from it

**1. The first attempt actually succeeded.** The compile ran to completion and populated the Core ML / E5 bundle cache. That is visible in the retry two minutes later:

```
09:28:44  modelCompilationStarted
09:28:48  modelCompilationCompleted duration=3636ms
```

3.6 s against a cold 236 s. We discarded a completed compile and told the user it had failed.

[...]

Issue #167, opened 2026-05-09
Cold-start keyboard dictation fails while model is loading


Pivii:
## Summary

Cold-start dictation from the iOS keyboard can fail on the first attempt with a user-visible “model not loaded / model loading” style error, even though the app successfully launches and warms the engine immediately afterward.

This is a regression from the expected cold-start flow: after tapping the Dictus mic from a host app, the app opens, the user swipes back to the host app, and dictation should start. Instead, the first cold-start attempt transitions to `failed`; the next warm-start attempt works normally.

## Environment / build

From attached logs:

- iOS: 26.3.1
- App: 1.7.0 (16)
- Revision: `773cb03@fix/163-turbo-long-audio-truncation`
- Device: iPhone16,2 / 8 GB RAM
- Model: `parakeet-tdt-0.6b-v3`

## Reproduction

1. Reset Dictus logs.
2. Kill the Dictus app to simulate a cold start.
3. Open a host app, e.g. Apple Notes.
4. Show the Dictus keyboard.
5. Tap the Dictus mic button.
6. Dictus opens via cold-start URL.
7. Swipe back / return to the host app.
8. Observe the keyboard error: model is not loaded / model loading.
9. Tap the mic again without killing the app.
10. Warm-start dictation works normally.

## Observed behavior

First cold-start attempt fails because dictation starts while model loading is in flight:

```text
06:52:26 KBD keyboardMicTapped
06:52:26 KBD coldStartDarwinFallback elapsedMs=518 status=requested
06:52:26 APP appLaunched version=1.7.0
06:52:27 APP modelLoadStateChanged from=ready to=loading reason=init-preload
06:52:27 APP coldStartURLReceived isColdStart=true isEngineDead=false hasBeenActive=false
06:52:27 APP dictationStarted fromURL=true appState=1 engineRunning=false
06:52:27 APP dictationDeferred reason=model load in flight (state=loading)
06:52:27 APP statusChanged from=idle to=failed source=coordinator
06:52:27 APP liveActivityFailed context=rejectedTransition error=idle->failed
06:52:27 APP engineWarmUpSuccess context=didBecomeActive
06:52:27 APP modelLoadStateChanged from=loading to=ready reason=didBecomeActive-success
06:52:27 APP appWhisperKitLoaded model=parakeet-tdt-0.6b-v3
```

Then a second attempt succeeds as warm start:

```text
06:52:33 KBD keyboardMicTapped
[...]

PR #429, opened 2026-08-26
Clear a stuck model load, and bound the one that made it (#428)

Pivii:
refs #428

One interrupted model load could lock a device out of its own app on every launch, forever. This PR removes that, and bounds the wait that produced it. **It does not add an escape hatch — one was built, reviewed four times, and cut.** See "Why there is no escape hatch" below.

| | commit | what it does |
|---|---|---|
| **A** | `68642a6` | a fresh process can no longer believe a load is in flight |
| **C** | `d7c2401` | the launch preload gets a deadline that actually returns |
| | `432fdf8` · `9be4e4b` · `1b06dd5` · `6eabea1` · `65ff33d` | the machinery underneath, hardened across four reviews |
| | `809a63f` | the escape hatch removed |

## A — a fresh process has no load in flight

`modelLoadState` is persisted in the App Group and had one writer, on the load path. A compile interrupted by a force-quit or a jetsam left `"loading"` behind with no owner and nothing to correct it. The keyboard reads that value and answers every mic tap by opening Dictus with `intent=prepare`, and that intent makes `MainTabView` replace the whole tab bar with the preparation screen.

`ModelLoadState.clearStaleLoadingState(in:)` runs at the top of `DictusApp.init`, above every reader. `ready` is deliberately left alone: believing it costs one lazy load, believing `loading` costs the whole app.

**Confirmed working on the affected device**, build 52db06f:

```
19:46:13Z modelLoadStateChanged from=loading to=idle  reason=stale-at-launch
19:46:13Z modelLoadStateChanged from=idle to=loading  reason=init-preload
19:49:35Z modelLoadStateChanged from=loading to=ready reason=init-preload-success
```

That first line is A firing. The second is the transition that used to be invisible — `setModelLoadState` returns silently when the value already matches, so a process starting on a stale `loading` said nothing at all; the device log had six consecutive `appLaunched` lines with no state line after any of them.

## C — the launch preload has a deadline that returns

It used to `await ensureEngineReady()` naked. The budget is the per-model catalogue value via `ModelInfo.preloadDeadlineSeconds(for:)` (Turbo 300s, else 120s, #406).

`withPrewarmTimeout` is deliberately neither reused nor extended: #427 established that a task group cannot return while a child runs and that `cancelAll()` is a request a Core ML compile ignores. The device log shows that failing in the wild — `modelPrewarmTimeout timeout=5s` logged **212 seconds** after the compile it names started. This deadline is arbitrated between two independent tasks instead. Expiry frees the UI and nothing else; it never fails a model.

## The machinery underneath

Kept deliberately, not by omission.

**One lock for the Neural Engine.** The ANE cannot compile two models at once. It had two locks that did not know about each other — `ModelManager.isPrewarming` for the download path, `initTask` for `ensureEngineReady` — and nothing covering the pair. **That hole predates this branch**: `ModelManager` is instantiated more than once (onboarding builds its own), so an instance-owned flag never serialised the two instances either. There is now one lock, on the singleton, and both paths take it. This is strictly better than what `develop` has today.

**The load epoch and the publish gate.** The deadline needs an answer to "may a load that lands after the app stopped waiting still install its engine?" — and the answer is yes when the model is still the active one. Getting that wrong cost a working dictation (see below), which is why the gate asks about the model rather than only about the epoch.

## Two regressions this work produced, and how they were caught

Worth reading, because both were introduced by fixes and neither was visible to reading.

**The deadline threw away a load that succeeded.** A keyboard cold start parks on the launch preload's task by the #167 rule; the deadline expired at 120s and bumped the epoch; the compile then *succeeded* at 150s and was discarded, failing the parked recording. That dictation transcribed normally before this branch existed. Fixed in `432fdf8`.

**The stage watchdog threw away a finished recording.** `stopDictation` enters `.transcribing`, arming a 30s watchdog, then calls `ensureEngineReady()` — which since the lock can queue behind a compile for ~200s. The watchdog fired and cancelled a dictation whose audio was already captured. `develop` never had this; the lock created it. Fixed in `65ff33d` by teaching the watchdog that a stage queued on hardware is not a stalled stage.

**And one that was invisible to reading but obvious to execution.** The lock's poll used `try?` on its sleep, so a cancelled waiter stopped sleeping and spun the main actor for the length of a compile. Measured, not argued:

```
iterations in the second after cancellation:
  try?      (the defect) : 133,109
  try await (the fix)    : 1
```

## Why there is no escape hatch

An escape from the preparation screen was built (a 45s offer, "choose another model", landing on the Models tab). It went through four independent review passes. **Every serious finding in all four was in the escape or in something the escape forced. Zero were in A or C.**

That is a property of what the escape is — a new user action taken mid-flight against a resource that cannot be interrupted, which then has to reconcile with every path that touches model state — rather than a run of bad luck. The rule for cutting it was set before the fourth review reported; the fourth review found two HIGH; it was cut.

**What that costs the user, without softening:** a compile that outruns its budget leaves them on a preparation screen with no way off it until the deadline returns the app — 300s for Turbo, 120s otherwise.

**What it does not cost them is the bug.** The lockout is gone either way, because A is what removes it: no launch can begin in the stuck state. A wait that always ends, on one launch, is a different animal from a state that survives force-quits and reinstalls, which is what broke the affected device. On the three cold readings we have — 236s, 212s, 202s against a 300s budget — the deadline is not expected to fire at all on that hardware; it is the backstop for something slower than anything we have measured.

## On the compile timings in the log

Three cold readings of `turbo_632MB` on the same iPhone 15 Pro Max: **236s**, **still running at 212s**, and **202s start to finish**. Any load that finds the Core ML bundle cache afterwards takes **seconds** — 3636ms measured. So `modelCompilationCompleted duration=` is not a cold-compile figure unless you know the cache was cold.

**The cache lives in the app container, so every development install throws it away.** The first launch after any install pays a full cold compile. Expect that during the manual tests — it is the reading, not a fault.

**Turbo is not broken.** It compiles and dictates correctly on the affected device. The lockout was the whole bug.

## What this branch's guarantees actually rest on

Read this before approving.

**The Neural Engine lock, the load epoch and the publish gate have no automated coverage of any kind, and cannot have any in this repo.** All of it lives in `DictusApp`, which has no test target; `DictusCore` is the only suite, and CI never runs it either. The green tests below say nothing about the concurrency work — they say the catalogue, the state machine and the copy contract still hold.

**With the escape cut, the lock has no device evidence at all.** The only sequence that made two compiles contend went through the escape: download Turbo, escape during its compile, select another model. Without it, the preparation screen covers the Models tab for the whole compile, and the only other route into `ensureEngineReady` is a dictation, which requires leaving the app and so suspends the compile you were trying to contend with. Step 9 below records what that test would have been and why it can no longer be run.

So the lock rests on: four adversarial reviews, one standalone harness (below), and the reasoning in its doc comments. Not on a device pass, and not on a test.

**The harness**, replicating the lock's logic line for line because it cannot be imported:

```
1. exclusivity under 5 contenders : maxConcurrent=1, gate not left held
2. throwing holder x3             : gate not leaked
3. cancelled waiter               : did not spin, did not take the gate
4. cancel-at-free race x200       : waiter took the gate anyway 0 times
5. termination, 8 contenders      : 8/8 proceeded, gate released
ALL CHECKS PASSED
```

That is evidence about the lock's *shape*. A replica can be faithful to the wrong thing.

## Verification

At `809a63f`:

- `cd DictusCore && swift test` — **1250 tests, 1 skipped, 0 failures**
- `swiftlint lint --strict` — **0 violations in 218 files**. No baseline, nothing disabled.
- `DictusApp`, `DictusKeyboard`, `DictusCore` — all **BUILD SUCCEEDED**, headless simulator destination

The branch adds 14 tests net: the stale-state reset (8), the per-model deadline (3), and how a preparation ended (3). Two build warnings sit on lines this branch does not touch.

## Tests manuels sur iPhone

À faire sur un appareil physique avant le merge. Le simulateur ne compile pas les modèles comme le fait le Neural Engine.

À savoir : le cache Core ML vit dans le conteneur de l'app, donc **chaque installation de développement le jette**. Le premier lancement après une install paie une compilation complète (environ 200 secondes pour Turbo sur un 15 Pro Max). C'est normal.

**1. Le cas qui a produit l'issue : chargement bloqué, app relancée**

1. Réglages > Modèles, choisir **Turbo** et le télécharger.
2. Pendant la phase « OPTIMISATION » (la compilation, pas le téléchargement), **tuer l'app** : double-clic Accueil, balayer vers le haut.
3. Rouvrir Dictus depuis l'écran d'accueil.
4. **Attendu** : l'app s'ouvre normalement sur ses onglets. Accueil, Modèles et Réglages sont tous accessibles.
5. Dans le log, vérifier `reason=stale-loading-cleared-at-launch` juste après `appLaunched`, puis `from=idle to=loading reason=init-preload` juste après.

**2. Le clavier n'est plus bloqué non plus**

1. Après l'étape 1, ouvrir Notes ou Messages.
2. Passer au clavier Dictus, appuyer sur le micro.
3. **Attendu** : la dictée démarre, ou affiche une vraie erreur. Elle ne doit pas renvoyer indéfiniment vers l'écran de préparation.

**3. Une dictée n'est pas perdue quand une compilation occupe le Neural Engine** (la régression corrigée en `65ff33d`)

1. Lancer le téléchargement d'un modèle et attendre le début de la phase « OPTIMISATION ».
2. Quitter l'app, ouvrir Notes, dicter une phrase avec le clavier Dictus, appuyer sur stop.
3. **Attendu** : le texte finit par arriver. Il peut prendre longtemps — jusqu'à la fin de la compilation — mais il ne doit **pas** être perdu ni se terminer par une erreur.
4. Dans le log, chercher `action=stageWatchdogDeferred` : c'est la ligne qui prouve que le chien de garde a attendu au lieu d'annuler.

**4. Une préparation saine n'est pas cassée**

1. Choisir un modèle déjà téléchargé et déjà compilé une fois.
2. **Attendu** : préparation en quelques secondes, écran fermé tout seul avec la coche « Modèle prêt ».

**5. La deadline (long, et c'est la seule preuve pour C)**

1. Lancer une **première** compilation de Turbo, app ouverte au premier plan.
2. Si elle dépasse **5 minutes** (300 s), l'app doit rendre la main : écran fermé, onglets accessibles.
3. **L'écran ne doit PAS afficher « Modèle prêt »** en se fermant : aucun modèle n'est chargé à ce moment-là.
4. Chercher `ModelPreload ... action=deadlineExpired seconds=300`.
5. Sur les mesures actuelles (202 s à froid), cette étape ne devrait pas se déclencher. Une expiration ici voudrait dire que 300 s est trop court pour cet appareil.

**9. Le verrou du Neural Engine : ce test ne peut plus être exécuté**

Ce test existait pour couvrir la seule pièce de ce travail qui n'a aucune couverture automatisée. Il faisait entrer deux compilations en collision : télécharger Turbo, sortir de l'écran de préparation pendant sa compilation, puis choisir un autre modèle — le second chargement devait attendre le premier, ce que prouvait la ligne `NeuralEngine ... action=queuedBehindCompile heldBy=prewarm:...`.

**La sortie de l'écran a été retirée, donc cette séquence n'existe plus.** L'écran de préparation couvre l'onglet Modèles pendant toute la compilation, et le seul autre chemin vers `ensureEngineReady` est la dictée, qui oblige à quitter l'app et suspend donc la compilation qu'on cherchait à concurrencer. **Le verrou part en production sans aucune preuve sur appareil.** C'est écrit ici plutôt que caché derrière une étape qui ne marcherait pas.

⚠️ Si jamais une panne de ce type se produisait : deux compilations Core ML simultanées produisent une erreur de classe « E5 bundle », et d'après tout ce que ce dépôt a observé, **elle ne se présente pas comme une erreur mais comme un chargement qui ne revient jamais** — la même signature que le bug corrigé ici. La ligne `queuedBehindCompile` est ce qui les distingue.

## Follow-ups, deliberately not in this PR

1. **Nothing records that a model's compile has never succeeded on this device**, so the app retries the same compile on every launch. Fourth point of #428, still open.
2. **The model card says nothing specific while a load is queued** behind another compile. Open question: should it, and what should it say?
3. **The preparation screen has no way out again.** That is the deliberate outcome of cutting the escape, not an oversight — but if a device slower than anything measured ever blows the 300s budget, this is where it will hurt.

https://claude.ai/code/session_0155NfNULhoE24Yz5dGPjFTo

## Summary by CodeRabbit

* **New Features**
  * Improved model preparation and launch loading with model-specific deadlines.
  * Added clearer handling when a model finishes loading after the app stops waiting.
  * Audio setup now occurs consistently during foreground warm-up.

* **Bug Fixes**
  * Stale “loading” states are reset when the app launches.
  * Improved coordination when preparing multiple models.
  * Model preparation indicators now better reflect active downloads and prewarming.

* **UI Updates**
  * Removed the delayed escape option from the model-loading overlay.
  * Overlays now dismiss cleanly when preparation times out.


Pivii:
**Pushed `432fdf8`** — review finding 1 (HIGH).

An abandoned load returned instead of throwing, so every caller parked on the shared `initTask` read it as a loaded engine. Both engine paths now throw `SpeechModelError.loadAbandoned`, and the wrapping preserves it rather than flattening it into "the model could not be loaded".

Verified at this commit: 1248 tests / 1 skipped / 0 failures, swiftlint 0 violations in 217 files, DictusApp + DictusKeyboard + DictusCore all BUILD SUCCEEDED.

Findings 2, 3, 4, 5, 6, 7 still open, next.

Pivii:
**Pushed `9be4e4b`** — review findings 2, 3, 4, 5, 6 and 7. All seven are now closed.

One commit because they interlock: 2, 3 and 5 are consequences of what "abandoned" was allowed to mean, and 3, 4 and 7 all turn on when the 45 second clock starts.

The behaviour change worth flagging: **the escape no longer takes the init lock away.** It could not — the abandoned compile is still on the Neural Engine, which cannot compile two models at once, so "choose another model after escaping" was starting a second compile on top of the first. A later load now queues behind the abandoned one. The user gets their app back immediately; they get their next model when the hardware is free. Manual step 4 has been reworded accordingly.

Also folded in: the 202s cold reading from 2026-08-27, which makes the 300s budget's headroom a measurement (33%) rather than an estimate, and confirms the compile completes rather than hangs.

Verified at this commit: **1248 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 217 files**, DictusApp + DictusKeyboard + DictusCore all BUILD SUCCEEDED, headless.

Correction to earlier numbers in this PR: the branch adds **16** tests, not 17 (A 8, B 5, C 3 — counted from the commits). The reviewer's 1240 was a measurement of `68642a6` (A alone), not of `d7c2401`; the arithmetic checks out against three independent runs — develop 1232, A 1240, B 1245, C 1248.

Pivii:
**Pushed `eb3b753`** — comments only, no behaviour change.

Records the finding-2 decision where the next person will meet it: right above the wait loop, which reads like an obvious candidate for "simplification". The trade, in the source: a bounded, measured wait (202s cold Turbo compile, measured on the affected device) beats an E5-class failure nobody here has ever reproduced deliberately and therefore could not debug.

Also states B's promise plainly, since the original spec line promised more: **the escape frees the app immediately; it does not free the Neural Engine.** The PR body and manual step 4 now say the same thing — step 4 carries an explicit warning that its two-minute wait is the feature, not a defect, so it does not get filed as one.

Verified at this commit: 1248 tests / 1 skipped / 0 failures, swiftlint 0 violations in 217 files, all three targets BUILD SUCCEEDED headless.

Pivii:
**Pushed `a37903b`** — comment only, no behaviour change.

Records why `abandonedModel` is process-scoped rather than persisted, as a rule and not a convenience: what made #428 unrecoverable was state that outlived the process that wrote it, so a per-process memory of "the user walked away from this model" is the deliberate opposite. Persisting it would buy a little polish and re-introduce the exact failure mode — a flag no live process owns, quietly deciding what the app may do. That trade is available and it is the wrong one; the argument is now where someone will find it.

The PR body also gains a **follow-ups** section, with the open question this branch did not ask: during the step-4 wait the model list says nothing, which is the same confusion #428 exists to remove, one screen to the right. Whoever runs the device pass on step 4 should note what the list actually showed — that is the input the question needs.

Verified at this commit: 1248 tests / 1 skipped / 0 failures, swiftlint 0 violations in 217 files, all three targets BUILD SUCCEEDED headless.

Pivii:
**Pushed `182a383` and `30f8d89`** — all six findings from the second review are closed.

`182a383` — **finding 1**, the only regression on this branch. The publish gate asked "has the epoch moved?" and used it to answer two different questions. An engine is wrong to publish when it is for a model the user moved off; it is not wrong merely because the app stopped waiting. A load still carrying the active model now publishes whatever the epoch says, and logs `publishedLateLoad`.

`30f8d89` — **findings 2 to 6.** Finding 2 was structural, so it is fixed structurally: **the Neural Engine now has one lock instead of two.** `ModelManager.isPrewarming` is gone; both the download prewarm and `ensureEngineReady` take a single gate on the coordinator singleton. Patching the specific door (a card tapped on an uncovered list) would have left a third one — the two locks not knowing about each other is what let the same hazard be found twice. This also closes a hole nobody had filed: onboarding builds its own `ModelManager`, so an instance-owned flag never serialised the two instances against each other.

Also in `30f8d89`: the escape clock is cleared on the error dismissal (3); abandoning records the model the screen was preparing rather than whatever is active (4); the epoch moves only on abandonment, with a per-run latch settling the launch preload's race (5); both `engineWarmUpAttempt` positions are explicit and named (6); and the ordering comment in `DictusApp.init` is corrected — `@StateObject` takes an `@autoclosure @escaping` closure, so the coordinator is built lazily at first `body`, not as a stored-property default. Right conclusion, wrong reason, now fixed.

Verified at `30f8d89`: **1248 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 217 files**, DictusApp + DictusKeyboard + DictusCore all **BUILD SUCCEEDED** headless.

Pivii:
**Pushed `d466ee5` (merge) and `1b06dd5` (audit fixes).** Current SHA: `1b06dd5`.

**Merge first, as instructed.** Two conflicts, both the same shape — this branch extracted a block into a helper, develop changed the logging inside it. Resolved by keeping the extraction AND carrying develop's change in; either default would have been a silent revert. Three #313 obligations that did not conflict and would have shipped wrong:
- `.loadAbandoned` conforms through `DiagnosableError` (came through clean).
- Its two catch sites log through `DictationFailureMessage.diagnostic(for:)`, not `localizedDescription`.
- **The copy was a fault and is now a notice.** "The model preparation was interrupted. Tap the microphone again to retry." described damage. Nothing is broken — a load the app stopped waiting for is the expected outcome of a deadline or an escape. It now reads **"The model is still getting ready. Tap the microphone again."**, in `noWordsDetected`'s register. Old key retired, new one translated. `DictationErrorCopyTests` passes.

**Audit findings 1, 2, 3, 5** in `1b06dd5`. Finding 1's spin is fixed at the source: `try await` on the sleep, as `develop` always had, plus cancellation checked before taking the lock and again before the compile task is created. Finding 2 is fixed twice over — the foreground guard now skips only the model load so the audio session is always reconfigured (#123), and the abandon memory is cleared when a load of that model actually publishes. Finding 3 is fixed at the invariant rather than the call site: whatever `liveActivePrepModel` returns is the model the screen names, the model whose progress it draws, and the model the escape abandons, so per-model evidence now beats the global flag. Finding 5's early release is in.

**The PR body has a new section before Verification: "What this branch's guarantees actually rest on."** It states plainly that the lock, the epoch and the publish gate have no automated coverage and cannot have any in this repo, that the 1252 green tests say nothing about them, and that the lever if that is not enough is a device pass rather than another read.

Verified at `1b06dd5`: **1252 tests / 1 skipped / 0 failures** (baseline moved with the merge), swiftlint **0 violations in 218 files**, all three targets **BUILD SUCCEEDED** headless.

Pivii:
**Pushed `6eabea1`** — third-review findings A, B, C and D. Current SHA: `6eabea1`.

**C is the one that mattered**, and you were right to call it out as the headline claim contradicted rather than as a defect. `selectModel` cleared `preparationDismissedByUser`, so escaping and then tapping a lighter model — the one obvious next action — re-covered the app for the whole of that queued load. The promise held for about four seconds. It no longer clears it: once someone has said they want the app rather than the waiting screen, that answer holds for the session, and the model card carries the load's state instead. `downloadModel` still clears it, because starting a download is a new explicit request.

**A** — both arms of the launch preload now check the epoch as well as their one-shot latch, which is what the other two load paths already did; the deadline task is cancelled once the load resolves instead of sleeping out its budget to no-op. **B** — #174's "a freshly downloaded model becomes active" now yields to a choice the user made afterwards, on both engine paths; without it, escaping a Turbo prewarm and picking Small silently restored Turbo and loaded it into RAM, and manual step 4 asserts the opposite. **D** — the two reasons that mean "the app stopped waiting" are shared constants now, and the screen dismisses without the "Model ready" checkmark when it sees one.

**D also produced the first testable thing in three reviews** — 3 new DictusCore tests covering both gave-up reasons, every resolved-load reason that must not be mistaken for one, and the literals themselves. The branch now adds 19 tests.

**The PR body's headline claim has been re-read against the code, as asked.** It now says the app "stays reachable", says plainly that this was the broken half, and manual step 4 gained two explicit checks: the app must not be re-covered, and Turbo must not have become active again.

Verified at `6eabea1`: **1255 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 218 files**, all three targets **BUILD SUCCEEDED** headless.

Pivii:
**Pushed `ad65aec`.** Current SHA: `ad65aec`.

Findings A, B, C and D were already fixed in `6eabea1` (our messages crossed). This commit adds what was missing from C: the rule it was implementing, written down.

**The second invariant**, stated where the dismissal flag lives, next to the first one:

> A screen the user has dismissed may only come back on work the user asked for AFTER dismissing it, and never on work that was already running when they dismissed it.

The comment also records that the implementation is **deliberately stronger** than the invariant — once dismissed, selection never re-presents this session, even after the abandoned compile finishes and the chosen model genuinely starts compiling, which the invariant would allow. The finer rule is not implementable today: at the moment the screen decides, `modelLoadState` reads `.loading` both for "queued behind an abandoned compile" and for "actually compiling". Separating them needs exactly the information the open follow-up about the model card is asking for.

### The gate, run rather than read

Two reviews found defects on this branch that were invisible to reading and obvious to execution. So the Neural Engine gate was driven through the states reading cannot settle, in a standalone harness replicating its logic line for line (it cannot be imported — no test target):

```
1. exclusivity under 5 contenders : maxConcurrent=1, gate not left held
2. throwing holder x3             : gate not leaked
3. cancelled waiter               : did not spin, did not take the gate
4. cancel-at-free race x200       : waiter took the gate anyway 0 times
5. termination, 8 contenders      : 8/8 proceeded, gate released
ALL CHECKS PASSED
```

And the spin fix measured directly against the defect it replaced:

```
iterations in the second after cancellation:
  try?      (the defect) : 133,109
  try await (the fix)    : 1
```

133k reproduces the audit's 105k independently, on different hardware. The fix is 1.

This is evidence about the gate's *shape*, not about the shipped binary — the harness is a replica, and a replica can be faithful to the wrong thing. What it does rule out is the class of defect that has bitten this branch twice.

Verified at `ad65aec`: **1255 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 218 files**, all three targets **BUILD SUCCEEDED** headless.

Pivii:
**Pushed `65ff33d` and `809a63f`.** Current SHA: `809a63f`. The rule fired and the escape is cut.

`65ff33d` — fourth-review findings 2, 1, 3, 5, all of which survive the cut.

**Finding 2 was the urgent one and the second regression this branch produced.** `stopDictation` armed a 30s watchdog and then called `ensureEngineReady()`, which since the lock can queue behind a compile for ~200s — so the watchdog cancelled a dictation whose audio was already captured. `develop` never had this; the lock created it. Fixed by teaching the watchdog that a stage queued on hardware is not a stalled stage: it re-arms and logs `stageWatchdogDeferred` rather than throwing the recording away. Nothing about status, the keyboard handshake or the Live Activity changed.

Also: the prewarm epoch is captured before the unbounded wait rather than after it (1); the Parakeet path sets `.prewarming` before queueing, as the WhisperKit path does, so the card stops saying "Downloading" at 100% (3); and a load that published its engine no longer logs `supersededByNewerLoad` while leaving the state at `.idle` (5).

`809a63f` — **the cut.** Removed: the offer and button, `preparationDismissedByUser`, the shared wait clock, the reveal timer, `allowsEscape`, `abandonPreparation`, `abandonInFlightModelLoad`, both strings, and their tests. 437 lines out, 25 in.

**Kept, as a decision:** the Neural Engine lock, the load epoch and the publish gate. `ModelPreparationEscape` is renamed `ModelPreparationOutcome` and keeps only what has nothing to do with the user leaving — how a preparation ended, so the screen stops announcing "Model ready" after a deadline. `user-left-preparation-screen` is gone from the gave-up reasons; the deadline is the only way left for the app to stop waiting.

**PR title and body rewritten** for what this is now. The body says plainly why there is no escape hatch, what the fallback costs without softening it, and that with the escape gone **the lock ships with no device evidence at all** — step 9 now records what that test would have been and why it cannot be run.

Verified at `809a63f`: **1250 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 218 files**, all three targets **BUILD SUCCEEDED** headless.

Pivii:
**Pushed `e541080`.** Current SHA: `e541080`.

**The CodeRabbit finding on `DictationCoordinator.swift:811` was correct and is fixed.** `waitingForNeuralEngine` wrapped the whole of `ensureEngineReady()`, so a dictation that found the engine free, took it, and then hung in its own compile kept the deferral raised — the watchdog re-armed every 30s and the dictation never recovered. That traded "cancels a dictation that was merely queued" for "can never cancel a dictation at all", which is an unbounded hang with no recovery.

The deferral is now raised **inside the queue wait** and comes down the instant the gate is taken. Two flags, because they answer different questions: one marks that a dictation is inside the engine load, and `acquireNeuralEngine` uses it to decide whether a wait it is about to enter belongs to one.

### On the other five comments — four are stale, one was not

I checked rather than assuming. Confirmed stale:

| comment | why it no longer applies |
|---|---|
| `DictationCoordinator.swift:1457` (6d368d5) — propagate an abandoned outcome | fixed in `432fdf8`; both paths now `throw SpeechModelError.loadAbandoned` |
| `+ModelLoad.swift:55` (6d368d5) — don't release serialization mid-compile | fixed in `9be4e4b`; `releaseEngineInitLock` is gone and the only `initTask = nil` left is the identity-checked `clearInitTask` |
| `ModelManager.swift:549` (6d368d5) — preserve the replacement model | references `abandonPreparation()`, removed by the cut (0 refs in source) |
| `ModelManager.swift:206` (9be4e4b) — reset the preparation wait clock | references `preparationWaitStartedAt`, removed by the cut (0 refs in source) |

**But the sixth was live, against `809a63f`, and is not stale:** `+ModelLoad.swift:143`. The deadline can land between the engine being published and `warmUp()` returning, so `abandonedModel` gets set *after* `clearAbandonedModel` already ran for that load. The late-success branch then published `.ready` while the model was still marked abandoned, and every later foreground warm-up skipped valid work for a model that was loaded and running. Applied in this commit.

Also removed a stale reference in a file header to `releaseEngineInitLock`, which no longer exists.

Verified at `e541080`: **1250 tests / 1 skipped / 0 failures**, swiftlint **0 violations in 218 files**, all three targets **BUILD SUCCEEDED** headless.

Issue #433, opened 2026-08-27
A model whose compile was interrupted is listed as not downloaded while its files occupy the disk


Pivii:
## What happens

Download a model and interrupt the app while it is compiling — force quit, or let iOS end the process. The files are fully on disk. The model then appears under **available models**, not under **downloaded models**, so from the user's side it looks like the download never happened.

It did. The weights are there and they occupy the space. The user has no way to see that, and no way to reclaim it: you cannot delete a model the app does not believe you have.

Reported and reproduced by the maintainer on 2026-08-27, build `809a63f`, iPhone 15 Pro Max.

## Measured

```
09:30:39Z modelDownloadCompleted  name=openai_whisper-small
09:30:39Z modelCompilationStarted name=openai_whisper-small
09:30:45Z appLaunched version=1.8.0          <- force quit during the compile
...
09:31:18Z modelDownloadCompleted  name=openai_whisper-small   <- second attempt, instant
09:31:18Z modelCompilationStarted name=openai_whisper-small
09:31:46Z modelCompilationCompleted name=openai_whisper-small duration=27413ms
```

The second `modelDownloadCompleted` lands 0 seconds after the request because nothing was downloaded — the files were already there. The maintainer saw the progress bar jump straight to 100%, which is the app confirming it had the model all along while listing it as one the user does not have.

Small is about 500 MB. Turbo is 645 MB and Medium is 1.5 GB, so the invisible cost scales with exactly the models a user is most likely to abandon mid-compile.

## Where it comes from

`dictus.downloadedModels` is written by `ModelManager.persistState()`, and the identifier is added to that list only after the whole download-and-prewarm sequence completes. A compile interrupted between the two leaves the files present and the list unaware. `WhisperModelRepository.installedModelFolderURL` already answers a different question correctly — it treats a model as installed on the presence of a `.mlmodelc` directory, which is true here.

So the app holds two disagreeing views of the same fact and shows the user the wrong one.

## What the fix has to decide

The straightforward direction is that a model whose files are complete belongs in the downloaded list whatever its compile state, so it can be seen and deleted. That is not free of questions, and they should be answered rather than assumed:

- **Downloaded and unusable is a third state.** Listing it as downloaded and letting the user select it walks them into the compile they interrupted. Whether it needs its own presentation is a product call.
[...]

PR #435, opened 2026-08-27
List a model whose compile was interrupted (refs #433)

Pivii:
refs #433

## What this was for

You downloaded Small, force quit while it was compiling, reopened Dictus, and the model was back under **Available** as if the download had never happened. It had. The 500 MB were on disk, invisible, and unreclaimable: the delete affordance only exists on a card the app believes it has, so there was no way to see the space was gone and no way to get it back. Turbo is 645 MB and Medium is 1.5 GB, so the invisible cost scales with exactly the models a user is most likely to abandon mid-compile.

The three open questions in the issue were settled on 2026-08-27 and are implemented as decided: no third visual state, completeness established on the full expected set of files, Parakeet out of scope (recorded as a comment on the issue).

## To test by hand

Everything below needs a physical device. A simulator cannot compile Core ML for the ANE, so it cannot produce the interrupted compile this issue is about.

**Use `Small`, not Turbo.** Small is the model in your own report, its compile took 27 s on your 15 Pro Max (per the log in the issue), and 27 s is a comfortable window to force quit into. Turbo would make you sit through a full cold compile — every dev install throws away the Core ML bundle cache, so the first launch after installing this branch already costs about 3 min 30 for Turbo before you even start.

1. Install this branch on the iPhone. Open Dictus, go to **Modèles**, and make sure **Medium** (or any model other than Small) is downloaded and active, and that **Small** is NOT downloaded. Delete Small first if it is there.
   → Expected: Downloaded holds one model, Small sits under Disponibles.
2. Tap **Small**. Wait for the progress bar to reach 100% and for the screen to say it is optimizing / preparing.
   → Expected: the download finishes, then the preparation screen appears.
3. **While it is still optimizing**, force quit Dictus (swipe up from the app switcher). You have roughly 27 s from the moment the preparation screen appears.
4. Reopen Dictus and go to **Modèles**.
   → Expected, and this is the fix: **Small now appears under Téléchargés**, not under Disponibles. Before this branch it appeared under Disponibles.
5. Tap the **···** menu on the Small card.
   → Expected: **Supprimer le modèle** is offered and is enabled (Medium is the active model, and there is more than one model, so neither guard blocks it). This is the affordance the issue said did not exist.
6. Tap **Supprimer le modèle** and confirm.
   → Expected: Small disappears from Téléchargés and returns to Disponibles, and the disk space is actually freed. Check Réglages iOS → Général → Stockage iPhone → Dictus before and after if you want the number.
7. Redo steps 2 and 3, then reopen, and this time **tap the Small card** instead of deleting it.
   → Expected: the preparation screen appears and runs the compile that was interrupted, with no re-download (it should reach the compile essentially instantly, because every file is already on disk). When it finishes, Small becomes the active model.
8. Export the debug log (Réglages → Debug) and look at the launch in step 4.
   → Expected: exactly one line `modelReconciledFromDisk name=openai_whisper-small`, and no such line on any later launch.

**Second scenario, the onboarding one.** Only if you want to cover it; it needs a wipe.

9. Delete Dictus from the phone and reinstall this branch so onboarding runs from zero. On the model page, tap **Installer le modèle**, then force quit while it optimizes.
10. Reopen Dictus.
    → Expected: onboarding still shows **Installer le modèle**, NOT **Continuer**. Tapping it must resume at the compile without re-downloading. What must never happen is onboarding waving you through to a keyboard whose mic answers "Aucun modèle téléchargé".

**One boundary worth knowing about, not a bug to hunt.** If the interrupted model is the *only* model on the device, it now appears under Téléchargés but still cannot be deleted: `deleteModel` refuses to remove the last model, which is a deliberate product rule I did not touch. The user sees the space and can prepare the model, but reclaiming it means deleting the app. Your reported case (Medium present, Small interrupted) is fully covered. Say the word if you want that guard revisited and I will open a separate issue.

## What changed

**`DictusCore/Sources/DictusCore/WhisperModelRepository.swift`** — a strict completeness check and the reconciliation itself, both pure and testable.

The existing `isModelInstalled` treats one `.mlmodelc` directory as proof of installation, and that is not enough here. `ModelRepoDownloader` creates each file's parent directory *before* downloading the file, so a download interrupted inside `weights/weight.bin` — 92% of the payload, and therefore where an interruption almost always lands — leaves all three bundle directories present and every one of them empty. `hasCompleteDownload` descends into each bundle instead, requiring `coremldata.bin`, `model.mil` and `weights/weight.bin` in all three, plus the root `config.json`.

That list was verified, not assumed: the repository tree of all seven catalogue variants (tiny, base, small, small_216MB, medium, both turbos) was read on 2026-08-27, and every one carries exactly those entries in each bundle. `model.mlmodel` is deliberately absent because the two turbos do not ship it.

The *set of bundles* is read off the variant folder rather than hardcoded, so it covers whatever the variant actually shipped — the turbos' fourth bundle included — and cannot go stale the next time the repository adds one. The three universal bundles remain as a floor, because on-disk enumeration cannot see a bundle that was never created at all. See "Review fixes" below for why that matters.

It is a **sibling** of `isModelInstalled`, not a hardening of it, and the reason matters. That helper has two callers — the dictation load path and `WhisperKitEngine.prepare` — and both read `nil` as "send the user to the model manager to download it". A false negative there tells someone holding a perfectly good 1.5 GB model to fetch it again. A false negative in the reconciliation only reproduces the bug that already exists. Strictness is cheap on one side and expensive on the other, so it went on the cheap side.

**`DictusApp/Models/ModelRepoDownloader.swift`** — the required-path list now comes from `WhisperModelRepository`. The downloader's final tripwire and the reconciliation must name the same paths, because the reconciliation is only allowed to trust files the download promised to place.

**`DictusApp/Models/ModelManager.swift`** — `loadState` reconciles before anything reads the bookkeeping, and `persistState` writes an honest `modelReady`.

The reconciliation never writes `activeModel`, which is the load-bearing constraint. Which exposed a definition that had never had to be right: `persistState` wrote `!downloadedModels.isEmpty` into `SharedKeys.modelReady` rather than the `isModelReady` that also requires an active model. The two could not disagree while the only writer of the list elected an active model in the same breath. They can now — a device whose first ever download was interrupted comes up with a populated list and no active model — so the flag now carries the definition it always claimed. Its three readers (the launch preload, `startDictation`, the foreground warm-up) all take it as "there is a model to load".

**`DictusApp/Onboarding/ModelDownloadPage.swift`** — treated the presence of the recommended identifier as "this page is done". After reconciliation that would wave the user through to a keyboard whose mic answers "no model downloaded". It now also requires that some model actually finished preparing.

**`DictusCore/Sources/DictusCore/LogEvent.swift`** — `modelReconciledFromDisk(name:)`, one line per model, only on the launch that repairs the disagreement.

## Review fixes (after the first round)

Three defects were found by review and fixed on this branch. All three were the same mistake in different places: two disagreeing definitions of "complete", with the weaker one believed.

**1 — The tripwire named directories, not files** (CodeRabbit, commit `07a9301`). `requiredDownloadPaths` returned the three `.mlmodelc` directories, and Phase 4 of `ModelRepoDownloader` checks them with `FileManager.fileExists`, which answers true for a directory. Since the downloader creates a file's parent directory before fetching the file, the tripwire passed on an `AudioEncoder.mlmodelc/` that existed and was empty. It now returns the leaf files, and `hasCompleteDownload` consumes that same list instead of rebuilding it.

**2 — The required set was three bundles, and the fourth is the one that matters** (commit `1cfa4c3`). The turbos ship `TextDecoderContextPrefill.mlmodelc`, which the first version of this fix excluded on the argument that only two variants of seven have it. Download order makes that unsafe: `listRequiredFiles` walks the repository breadth first, draining every bundle's small top-level files before any bundle's `weights/`, and the prefill bundle is enqueued last. Its `weights/weight.bin` — 12 MB on turbo_632MB, about 98 MB on the 954MB legacy variant — is the final file of the whole download. Interrupt it there and the three checked bundles are complete while the fourth is hollow; WhisperKit gates `prefillData` on the directory existing and then throws loading it. The user would have got a card in Téléchargés, already `.ready`, that failed every selection with no way to re-download. The required set now reads the bundles off the variant folder.

**3 — The reconciliation ran on view appearance, not at launch** (same commit). It sat in `loadState()`, which `HomeView` calls on `.onAppear` and on the onboarding-completed notification, and `ModelManagerView` calls on `.onAppear`. An ordinary download sits in exactly the repaired state — files complete, identifier not yet appended — for its entire prewarm, 27 s for Small and about 3 min 30 for Turbo, so switching tabs during a normal compile adopted the in-flight model early and logged a repair for a download nobody interrupted. It now runs from `init` behind a once-per-process flag; the flag rather than init alone, because onboarding builds its own `ModelManager`.

A fourth, the identical directory-versus-leaf hole on the Parakeet tripwire, is out of scope here by the Parakeet exclusion and is tracked in #438.

## Acceptance criteria and evidence

The issue carries no checklist, so these are its stated requirements plus the three settled decisions and the named risk.

| --- | --- | --- | --- |
| 1 | A model whose files are complete is listed as downloaded | **Verified** | Simulator, app running. Planted a complete Small folder and a partial Medium one, launched: log shows `modelReconciledFromDisk name=openai_whisper-small`, App Group holds `dictus.downloadedModels = ["openai_whisper-small"]` |
| 2 | An interrupted *download* is NOT listed | **Verified** | Same run: Medium had all three bundle directories and `config.json` but no weights, and was not added. Unit test `testDownloadInterruptedInsideABundleIsNotComplete` |
| 3 | Completeness uses the full expected set, not one `.mlmodelc` | **Verified** | `testRequiredDownloadPathsMatchTheBundlesTheCheckLooksFor`, `testAMissingWeightFileAloneMakesTheDownloadIncomplete`, `testAMissingConfigFileMakesTheDownloadIncomplete`. Expected set read from the live repository tree of all 7 catalogue variants |
| 4 | No third visual state | **Verified by construction** | The reconciliation only appends to `downloadedModels`; `ModelManagerView` splits its sections on that list and `modelStates`. No new state, no new view |
| 5 | The reconciliation never sets `activeModel` | **Verified** | After 5 app launches with a reconciled model present, `dictus.activeModel` is still absent from the App Group plist. `unlistedCompleteDownloads` returns `[String]` and has no other output (`testReconciliationReportsOnlyIdentifiersAndNeverElectsAnActiveModel`) |
| 6 | `modelReady` cannot go true on a device with no usable model | **Verified** | Same run: `dictus.modelReady = false` with `downloadedModels` non-empty and no active model |
| 7 | Already-listed models are unchanged / idempotent | **Verified** | 5 `appLaunched` lines in the log, exactly 1 `modelReconciledFromDisk`. `testCompleteFilesAlreadyInTheListChangeNothing` |
| 8 | Parakeet exclusion stated and recorded | **Done** | Stated above and in the code comment; comment posted on #433 |
| 9 | The model can actually be deleted after reconciliation | **Not verified** | Needs the device. Steps 4-6 of the manual list |
| 10 | Selecting it resumes at the interrupted compile | **Not verified** | Needs the device. Step 7 |
| 11 | The card is drawn under "Téléchargés" | **Not verified** | Step 4. `axe` would not attach to the simulator on this machine, so the Models tab could not be reached headlessly; the section split is a direct function of criterion 1, but no screenshot backs it |

Commands run in the worktree:

- `cd DictusCore && swift test` — **1278 tests, 1 skipped, 0 failures** (16 new in `WhisperModelRepositoryTests`, which goes from 10 tests to 26)
- `swiftlint lint --strict` — **0 violations, 0 serious in 219 files**
- `xcodebuild -scheme DictusApp -destination 'platform=iOS Simulator,id=<iPhone 17 Pro>' -derivedDataPath build/DerivedData build` — **BUILD SUCCEEDED**, producing `DictusApp.app`, `DictusKeyboard.appex`, `DictusWidgets.appex` and `DictusCore.o`

## One correction to the issue's context

The brief warned that `SharedKeys.modelReady` is read by the keyboard extension. It is not, on this revision: `DictusKeyboard` reads only `modelLoadState` (`KeyboardState.swift:1423`). `modelReady` has three readers and all three are in DictusApp. The risk was still real, just narrower than described, and it is handled — see criterion 6.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model download verification to recognize only fully completed downloads, including variant-specific files and bundles.
  * Restored completed Whisper models that were missing from the app’s download list.
  * Prevented incomplete or interrupted downloads from being treated as ready.
  * Fixed onboarding so interrupted model installation can resume instead of being skipped.
  * Improved model readiness reporting when downloaded files are present but no active model is loaded.
  * Reduced repeated model reconciliation during app usage.


Pivii:
@coderabbitai review

Pivii:
@coderabbitai review

Issue #439, opened 2026-08-27
Normal polish breaks the Natural contract three ways: no ASR repair, register rewritten, content deleted


Pivii:
> *This was generated by AI during triage.*

## Summary

Six French dictations were measured end to end on device (iPhone16,2, iOS 26.6, build 1.8.0 (27), `parakeet-tdt-0.6b-v3` + Normal polish, Apple FM). All six returned `outcome: success` and none was rejected by the guardrail. Reading the pairs against [ADR 0003](../blob/develop/docs/adr/0003-natural-polish-contract.md) shows the Natural contract is broken in three directions at once, and they are one calibration problem, not three:

- **Rule 8 is never applied.** Six incoherent ASR segments went through untouched, every one recoverable from the surrounding words.
- **The Preserve list is violated.** Spoken register was rewritten into written register.
- **The Forbidden list is violated.** Content was deleted, and one word was substituted for another.

The model is under-repairing exactly where the contract demands repair, and over-editing exactly where it forbids it. Raw pairs live in #437, which measured the formatting question on the same run.

---

## A. Rule 8 is not applied

> **8.** Repair ASR hallucinations when a segment is clearly incoherent (pseudo-words, off-language fragment that does not fit) by reconstructing the speaker's intent in the target language using surrounding context.

| Raw (Parakeet) | Intended | Context available in the same sentence |
|---|---|---|
| `suivre une visio depuis la salle à tante` | `la salle d'attente` | `dentiste`, `rendez-vous`, `11h` two clauses earlier |
| `il faut que je répète le comptable` | `je rappelle` | the speaker says `qu'il faut que je rappelle` 15 words later |
| `si je les zappais` | `si je l'ai zappé` | `ça fait une semaine et demie qu'il attend` |
| `ce qui est un peu le cas honnête` | `honnêtement` | adverbial position, `honnête` is ungrammatical there |
| `Il prend ses morceaux` | `ces morceaux` | `le système le découpe en petits morceaux` one sentence earlier |
| `on soumet la partie Apple Store` | `App Store` | same paragraph contains `StoreKit`, `sandbox`, `on soumet` |

`la salle à tante` is the clearest case: it is not a French phrase, it is a pseudo-phrase, and it is precisely the "pseudo-word … reconstruct from surrounding context" case rule 8 was written for.

Rule 8 did fire once, on the one failure that was impossible to miss — an off-language fragment:

```
[...]

Issue #437, opened 2026-08-27
Lift the <<NL>> ban at discourse boundaries so long dictations get paragraphs, measured against the paired Typeless set


Pivii:
## Context

Pierre compared long French dictations in Dictus and Typeless and observed a clear formatting gap:

- Dictus currently transcribes, then applies a light faithful polish for spelling and small corrections.
- On longer dictations, provisionally around 20 seconds and above, the result often remains one large wall of text.
- Typeless appears to restructure the same kind of speech naturally with paragraph breaks, line breaks, and spacing.
- This observed Typeless behaviour is **not a Notes mode**. It does not necessarily create bullets, extract key points, or summarise.
- Typeless requires an internet connection, while Dictus must keep any equivalent capability fully on-device.

The exact Typeless rules are not known yet. They must be inferred from paired real-world examples rather than assumed from a small number of tests.

### A. Natural long-form formatting

This issue is about an automatic, faithful presentation pass for longer dictations:

- preserve the complete meaning and order of the speech;
- create paragraph breaks when the discourse naturally changes subject or section;
- add line breaks and spacing where they improve readability;
- do not turn everything into bullets by default;
- do not summarise, extract only key points, or discard details;
- do not require the user to arm a Smart Mode if this can safely remain part of Normal polish.

This is the behaviour currently observed in Typeless and the primary scope of this benchmark.

### B. The existing `Notes` Smart Mode in #79

This is a different, intentionally transformative mode:

- the user speaks several ideas;
- the mode turns those ideas into a sequence of concise bullet points;
- it may remove verbal filler and synthesise each idea;
- its purpose is to produce an actual note or idea list, not merely improve paragraph layout.

The current name `Notes` may be too broad because users could interpret it as ordinary long-form formatting. Naming should be revisited separately in #79. Possible directions include a name that communicates bullets, ideas, or an outline, but no replacement name is decided yet.

## Initial product hypothesis

Natural long-form formatting probably belongs to Normal polish because it should change presentation without changing substance. The existing bullet-oriented mode remains a Smart Mode because it intentionally transforms and condenses the speech.

This is a hypothesis to validate, not a settled implementation decision.

Do not decide that 20 seconds is a hard trigger. It is an initial observation to compare against text length, pauses, sentence count, discourse changes, and content shape.

## Evidence collection

Pierre will provide paired examples from Typeless and Dictus:

- record the spoken source in Voice Memos so the input can be retained as a fixture;
- capture the raw Dictus transcription;
[...]

Issue #385, opened 2026-08-23
Polish silently drops trailing words from a transcription


Pivii:
The polish layer sometimes drops words from the end of a transcription. The output reads well, so nothing signals the loss — the user only notices if they remember what they said.

## Observed

Device, iPhone 15 Pro Max, iOS 26.6, 2026-08-23 19:51:53Z, build 1.8.0 (27) on `24ad7f0`. Parakeet `parakeet-tdt-0.6b-v3`, French, `explicit(fr)`, natural mode, Apple FM.

Raw, 147 characters:

> Chez Décagement, nous avons conçu et testé ces baskets ultra flexibles qui permettent à votre enfant de courir, sauter et grandir **et leur campagne.**

Polished, 130 characters:

> Chez Décagement, nous avons conçu et testé ces baskets ultra flexibles qui permettent à votre enfant de courir, sauter et grandir.

`engineMs=2645`, `outcome=success`, no guardrail violation. The trailing clause is gone and nothing in the event says so.

## Why this is worth an issue rather than a shrug

The dropped fragment here was almost certainly a transcription artefact, and removing it arguably improved the text. That is exactly what makes it worth looking at: the engine made an editorial judgement about content, and the contract in ADR 0003 is about register and punctuation, not about deciding which of the user's words were meant.

The failure mode this points at is silent. A polish that drops a real trailing clause — a qualifier, a negation, a name — produces fluent output that the user has no reason to re-read. `PolishGuardrail` exists to catch the engine going off the rails; it did not fire here, so whatever it measures does not cover this.

## What is not known

- **How often.** One observation. Nothing has been counted.
- **Whether it is length-related, position-related, or artefact-related.** All three are plausible: the fragment was at the end, it was ungrammatical, and it followed a sentence that already read as complete.
- **Whether the guardrail could catch it.** `PolishGuardrail` and `PolishPostpass` both already compare raw and polished; a 12% length drop with no punctuation change may already be measurable there.

The debug ring stores `raw` and `polished` side by side for every event, so the first cheap step is to look for the pattern in existing exports rather than to reproduce it: count events where `polishedCharCount` is materially below `rawCharCount` and read what went missing.

## Out of scope of #361

The polish pipeline is untouched by #361 — that issue moved *where* the call runs, not what it does. This is pre-existing behaviour, surfaced while testing that PR.

Related: ADR 0003 (natural polish contract), #141 (the pipeline), `PolishGuardrail`, `PolishPostpass`.

[...]

Issue #270, opened 2026-07-30
Guard the Foundation Models context ceiling: estimate before calling, warn while recording


Pivii:
## Context

Captured during the #79 Smart Modes design session (2026-07-30).

`AppleFoundationModelsPolishEngine` documents the constraint in its own comments: the Foundation Models session has a **4096-token ceiling covering instructions, input and output together**. It is the reason the engine throws away its session after every call rather than reusing it.

Nothing in the pipeline checks the input against that ceiling before calling the model.

## Today: a silent, mild failure

A very long dictation makes the engine throw `exceededContextWindowSize`. `PolishPipeline.transform` catches it as `.engineFailed`, and `resolvedOutput` returns the deterministic floor. The user gets their text with verbal punctuation applied but no polish, and no explanation. Mild, rare, and easy to miss — which is why it has never been reported.

## With Smart Modes: a loud failure

The same overflow means the mode did not run at all. The user asked for a translation and gets their original language back, after having spoken for several minutes. #79 resolves that by failing closed with an explicit message — but failing *closed* after eight minutes of speech is still a bad outcome. The user should have known before opening their mouth, or at least while speaking.

## Proposal

**Estimate before calling.** Compute a cheap token estimate of instructions + input + expected output, and when it exceeds the budget, refuse explicitly rather than letting the engine throw.

**Warn during recording.** While the LLM stage is enabled — free polish or a Smart Mode — surface a visual cue in the recording overlay as the dictation approaches the ceiling. Roughly a thousand words of output, so on the order of seven to eight minutes of speech; the exact threshold should come from measurement, not arithmetic. The overlay already shows an elapsed timer, so there is a natural place for it.

The cue must not read as an error. Approaching the limit is normal; the user simply needs to know that past a point, the model stops being part of the deal.

## Acceptance criteria

- [ ] Input size is estimated before the engine call; oversized input is refused with an explicit reason rather than throwing
- [ ] The recording overlay signals when a dictation is approaching the ceiling
- [ ] The signal is informative, not alarming, and does not appear on normal-length dictations
- [ ] Thresholds are derived from measurement on device and documented in code

## Out of scope

- Chunking long input and stitching the results (map-reduce for summarisation) — a separate, larger piece of work
- Removing the ceiling by switching backends (see #268)

## Related

- #79 — Smart Modes
- #267 — processing-stage animation (same overlay surface)

Pivii:
> *This was generated by AI during triage.*

[...]

PR #441, opened 2026-08-27
Ban deleting dictated content, preserve the spoken forms, and measure why rule 8 cannot be widened

Pivii:
refs #439 · refs #437 (fixtures)

## What this was for

Six French dictations measured on device broke the Natural contract (ADR 0003) three
ways at once: rule 8 never repairing an incoherent ASR segment, the Preserve list
violated by the one repair that did fire, and dictated content deleted. #439 argues they
are one calibration problem — what the `(.natural, fr)` prompt authorises and forbids —
and asks for one measured round against one fixture set, thresholds declared before the
first model call.

The round was run. **Two of the three defects turned out not to be prompt problems, and
the measurement says so with evidence.** What ships is the third one: three prompt lines,
measured, at a cost of 191 characters of maximum dictation length.

## To test by hand

Mac harness measurements throughout. **Nothing was confirmed on a physical iPhone**, and
item 1 is the one only a device can settle.

1. **Device pass.** Install this branch, keyboard French, transcription language
   **Auto-detect**, Normal polish on. Dictate the six #437 scripts (`raw` field of
   `DictusCore/Sources/polish-harness/fixtures/longform-fr.json`). Expect outputs
   comparable to #437's measurement comment: still one block with no line breaks, no
   salutation or sign-off invented, no guardrail rejection. The one thing this PR changes
   that you can see: **`l'abonnement machin` should come back as `machin`, not `machine`.**
2. **Read the six probe outputs and agree with the verdict.**
   `cd DictusCore && swift run polish-harness show ../docs/research/439-natural-contract/harness/probe-isolated.json --runs 5`.
   Each probe is one short sentence carrying one of the six ASR errors plus the context
   that makes the intended word obvious. Expect **0/5 repaired** on five of them. This is
   the measurement behind not widening rule 8, and it is worth one human read rather than
   trusting a table.
3. **No long-dictation probe needed any more.** The first version of this PR cost 580
   characters of ceiling and needed one; the cut brings it to 191 (4 160 → 3 969, against
   #270's 3 500 floor), which is inside the noise of a probabilistic overflow and not
   observable by dictating. `PolishContextBudgetTests` is the check, and it is green.
4. Reproduce anything with
   `cd DictusCore && swift run polish-harness show Sources/polish-harness/fixtures/longform-fr.json --runs 5`
   (add `--lang auto` for the path the device runs actually took) then
   `python3 docs/research/439-natural-contract/harness/score.py <capture>`.
   Needs macOS 26+ with Apple Intelligence; not run by CI.

## What changed

**The prompts — three lines each, and nothing else.** `PolishNaturalPromptFR`: placeholder
words (`machin`, `truc`, `bidule`) and the spoken forms (`ça`, `ça va` + infinitive) added
to PRESERVE; an explicit ban on deleting meaningful words added to FORBIDDEN, naming rules
6 and 7 as the only licence to remove a word and rule 8 as the only licence to change one.
`PolishAutoPrompt`: the same two PRESERVE items plus number and time formats, and the same
deletion ban worded without a rule 8 to point at, since that prompt carries none.

**Rule 8 is deliberately unchanged**, and a test now pins that decision rather than the
text. `PolishAutoPrompt` still carries no ASR-repair rule at all, with a doc comment
recording that this is now measured rather than an oversight.

**The six fixtures, first commit, for both issues.** `fixtures/longform-fr.json` — the six
device dictations, `raw` verbatim from the debug-ring export and byte-identical to #437's
"Full pairs" comment. `<<NL>>` leak check kept in every case, as #437's scope fence
requires.

**`--lang` on the harness.** Which prompt a dictation reaches is a device *setting*, not a
property of the text: all six measured runs were on `autoDetect` and so took
`PolishAutoPrompt`, while `lang: "fr"` runs them under `PolishNaturalPromptFR`. Both carry
the contract, so both are measured from one file instead of two copies of six transcripts.

**ADR 0003 gains an amendment** recording what ships and what was measured and rejected —
including that rule 8's reach is narrower than its wording suggests, with the probes as
the standing test for any future engine.

## Acceptance criteria, with evidence

Bars in `docs/research/439-natural-contract/bars.md`, committed before the first model
call. Numbers in `findings.md` (§Addendum for what ships), raw captures in `raw/`.
5 runs × 6 fixtures per arm per path — 60 outputs per version.

| Bar | Status | Evidence, `develop` → shipped |
|---|---|---|
| 1 · repairs ≥4/6 per run | ❌ **fails, unchanged** | 0/5 runs clear it on both paths, before and after. Five of six never repair, on any run, under any prompt tried |
| 2 · register preserved, 0 | ⚠️ **one item won, one did not** | `machin`→`machine` **7/10 → 3/10**; hour formats 2/10 → **0/10**; digits spelled out 2/10 → **0/10**; `cela` 6/10 → 8/10, unmoved |
| 3 · content deleted, 0 | ❌ **fails** | `en calcul` 9/10 → 9/10. Measured to be a segmentation defect, not a missing prohibition |
| 4 · content invented, 0 | ✅ holds | **0 in 60 outputs**, both paths — no salutation, sign-off, bracket placeholder or fact |
| 5 · ratio + guardrail | ✅ holds | **60/60 `success`, 0 guardrail rejections**, every ratio inside [0.92, 1.15] |
| 6 · scope fence (#437) | ✅ holds | **0 line breaks, 0 `<<NL>>` leaks in 60 outputs** |
| ceiling (#270) | ✅ holds | 4 160 → **3 969** chars of transcript, floor 3 500 never moved, `PolishContextBudgetTests` green on both bounds |
| no regression elsewhere | ✅ holds | `seed.json` **14/14 ×3**; `auto.json` 10,9,9 — identical to `develop`'s own 10,9,9 |
| build · tests · lint | ✅ | `DictusCore`/`DictusApp`/`DictusKeyboard` BUILD SUCCEEDED; `swift test` **1286 passed, 0 failures**; `swiftlint --strict` **0 violations** |

One watch item, reported as such: fixture 5's trailing `ça m'échappe mais ça me reviendra`
— #385's exact signature — was dropped **2 times in 60 outputs**, against 0 in 10 runs on
`develop`. Two occurrences is not a signal at this resolution, but it never happened on
`develop` and it has now appeared in three separate arms.

## What contradicts the issue

Both are measured, both stand, and both are being handled at the issue level rather than
here.

**1 — #439's premise on rule 8 is falsified.** The issue reasons *"Rule 8 did fire once …
So the capability is present."* It is not. Each broken segment was handed to the model
**alone, in one sentence, with the context that makes the intended reading obvious**:
**0/5 on five of the six**. Under a 1.5 KB prompt whose stated primary job was finding
misheard words, method spelled out, nothing competing: **still 0/5**. Apple FM detects an
off-language fragment (5/5) and does not detect a homophone that reads as fluent French.
Closing bar 1 needs a different mechanism — a lexicon or n-gram improbability check before
the model, a second pass, or a larger engine.

**2 — the deletion is a segmentation defect, and it belongs to #437.** `en calcul`
survives **6/20** in the unpunctuated run-on, flat from 311 to 838 characters, and **5/5**
when the same 311 characters carry their sentence boundaries. The words dropped are the
tail of the clause the model has to cut in two, dropped while it decides where the cut
goes. No clause in a fidelity prompt reaches it.

`harness/probe-isolated.json` and `harness/probe-length-ladder.json` are committed as the
standing evidence for both — one command against any future engine.

## The trade, decided

The first version of this PR widened rule 8 and cost **580 characters** of maximum
dictation length for one Preserve win. Decided by the maintainer: take the short version.
The product's central complaint — fixture 6 is his own unscripted voice saying it — is
that *long* dictations come back as a wall of text, and spending long-dictation headroom
on a marginal fidelity win, on the free tier, while #437 is about to make long dictations
better, is backwards.

Cost as shipped: **191 characters**, about 31 words, for the one Preserve item that moves
a number. Recorded here as settled, not open.

## Summary by CodeRabbit

* **New Features**
  * Added French long-form speech-polishing fixtures and evaluation scenarios.
  * Added language-routing controls, including automatic detection and French overrides.
  * Added route-specific evaluation expectations.

* **Improvements**
  * Preserves dictated wording, placeholders, spoken forms, numbers, times, and formats more reliably.
  * Prevents deletion of meaningful words except when explicitly permitted.
  * Clarified French polishing behavior and automatic-mode rules.

* **Documentation**
  * Expanded harness usage guidance and documented measured results and contract updates.

* **Bug Fixes**
  * Removed a build warning caused by an unhandled documentation file.


Pivii:
@coderabbitai review

Issue #423, opened 2026-08-25
An armed Smart Mode still reads as selected in the fan after entitlement is lost, and the dictation silently runs Normal


Pivii:
## Symptom

Found by the maintainer on device, 2026-08-25, while validating PR #411 (`rev e62892c@HEAD`, iPhone 15 Pro Max, iOS 26.6). Not caused by that PR — the code is on `develop`.

With `translate.en` armed, turn Smart Modes off in Settings, then dictate:

- The fan still renders `translate.en` as the selected mode. It is greyed, correctly, but it still reads as *the mode you are in*.
- The dictation runs **Normal**. French in, French out.
- **Nothing says so.** No toolbar message, no fan state change, no difference the user can see until they read the output and notice it was not translated.

In his words: *"visuellement ça me montrait que j'étais sur le mode anglais"* while the dictation was Normal.

## Evidence

```
14:54:08  smartModeFanOpened entries=3 reason=notSubscribed
14:54:14  WARNING [transcription] <APP> smartModeSkipped mode=translate.en reason=notSubscribed disarmed=false
14:54:29  WARNING [transcription] <APP> smartModeSkipped mode=translate.en reason=notSubscribed disarmed=false
```

The mode stays armed, the pipeline skips it, and the run completes as an ordinary success.

## What is NOT the bug

**Keeping the armed value is correct and deliberate.** `SmartModeStore.resolveArmedMode()` documents it:

> It does not disarm, because `.notSubscribed` is recoverable: resubscribe and the mode they chose is still there.

Clearing it would silently lose a setting over a condition that can lift. That reasoning holds and should not be reopened. The same doc block explains why returning the mode anyway is also wrong — it would fail closed on every dictation and insert nothing, indefinitely, for a choice made weeks ago.

So the state layer is behaving as designed. **The defect is entirely in what the user is shown.**

## What is the bug

Two things, and they are separable:

**1. The fan renders an unrunnable armed mode as the current selection.** Greying the row communicates "you cannot pick this", not "the thing you already picked will not happen". Those are different messages and only the second is true.

**2. The fallback is silent.** #79 specifies that an unavailable Smart Mode fails closed with an explicit message. This path fails *open* and says nothing: text is inserted, `outcome` is a success, and the only trace is a `WARNING` in a debug log the user never reads.

The second is the sharper one. A user who dictates a paragraph expecting a translation gets their own language back and has to notice on their own.

## Reachability

[...]

Issue #404, opened 2026-08-24
The non-subscriber's fan is a single Dictus Pro row (decided in #392, not built)


Pivii:
## Why this exists

#392 is closed: entitlement is now enforced at arming and at resolution (PR #395). What that issue's comment thread *also* settled — what a non-subscriber actually **sees** — is not built, and would be lost in a closed issue.

## The decision, from #392's thread

For a user without Pro, the long-press fan shows **one row: Dictus Pro**. Not three greyed mode rows with a reason line. Pierre, 2026-08-24:

> qu'on n'ait plus trois modes, mais une seule pilule Dictus Pro. Je pense que c'est la meilleure expérience.

Releasing on it leaves the keyboard for the app's paywall.

Two objections it answers by construction:

- **It is not a button.** `SmartModeFanView` is `allowsHitTesting(false)`: the finger went down on the mic pill and one gesture tracks it to the release, so a tappable control inside would take the drag from the recogniser. As the only *row*, it is selected by that same drag and committed by `commit()`, which opens the app instead of arming.
- **It costs no row budget.** There is no competition for `maximumEntries` when there is one entry. It gets the whole area, far above the 44 pt floor.

The honesty argument for showing disabled mode rows — let them see what the feature is before being told they cannot have it — is answered one screen later. The paywall describes the modes with room to do it; three unusable rows and a reason line is a worse advertisement than one control that leads somewhere.

## The rule that makes it safe

Also from that thread, and now the prerequisite rather than an independent half:

> il faut qu'on ait une règle qui, si on n'est plus en Dictus Pro, on est forcément en mode normal

With no mode rows there is no Normal row, and Normal is how a user *clears* a sticky mode. A subscriber who armed → EN and then lapsed would have an armed mode and no surface to disarm it. That is only safe because `resolveArmedMode()` already refuses to honour a mode without entitlement — shipped in #395. `.notSubscribed` stays **recoverable**, so it never disarms: they resubscribe and the mode they chose is still there.

Note the shape of this failure is exactly #402's, which shipped as a real bug on a different door: unpinning everything stranded the user in the armed mode. #403 fixed that by opening the fan on Normal alone. **Whatever this issue builds must not reintroduce it** — the Pro row replaces the *mode* rows, and the interaction with #403's armed-mode-alone fan needs a decision of its own.

## Open, and explicitly not answered here

**Should the discovery hint point a non-subscriber at the gesture at all?**

[...]

PR #443, opened 2026-08-27
Stop showing an armed Smart Mode that will not run, and give a non-subscriber the same fan greyed

Pivii:
refs #423, refs #404, refs #79

## What this was for

- **#423.** With a mode armed, turning Smart Modes off in Settings left the keyboard still showing that mode as *the mode you are in*. The dictation silently ran Normal — French in, French out — and nothing told the user. Keeping the armed value is correct and deliberate and is unchanged; what was wrong is that the UI read as selected and that the fallback was silent. **Device-validated 2026-08-29 on build `daad730`.**
- **#404.** The non-subscriber's fan. Decided in #392, built here as a single Dictus Pro row, then **re-decided on device on 2026-08-29** ([the amended spec](https://github.com/getdictus/dictus-ios/issues/404#issuecomment-5461472125)): the fan keeps its shape for everyone — `Normal`, the default-pinned modes greyed with a `PRO` tag, and a `Dictus Pro` row as the way out. Plus the finding from the same session: releasing on that row landed on the app's home screen instead of the paywall.
- **The rename**, decided on 2026-08-27 and recorded on #79: the bullet mode is `Liste` / `List`. Display name only — the identifier does not change. **Device-validated in the same session** (`smartModeArmed mode=notes` while the UI reads `Liste`).

## To test by hand

Items 1–7 are #423 and the rename, both already device-validated on 2026-08-29; they are kept as a regression list. Items 8–14 are the reworked #404 and are the ones that need you.

1. **Arm `→ EN`, then turn Smart Modes off in Settings → Pro Features.** Toolbar reads `→ EN · inactif` in grey; the mic pill has no badge.
2. **Long-press the mic.** `Normal ✓`, `→ EN INACTIF`, greyed mode rows, and **"Les modes sont désactivés. Réactive-les dans Dictus."**
3. **Dictate.** Text inserted, and one grey line: **"Les modes sont désactivés. → EN a été dicté en Normal."**
4. **Dictate again.** No message.
5. **Turn Smart Modes back on and dictate.** The mode runs, no re-arming, toolbar back to accent blue.
6. **Turn Apple Intelligence off with a mode armed and dictate.** **"Apple Intelligence est désactivé. → EN a été dicté en Normal."**
7. **A device that had `Notes` armed before this build** still has the bullet mode armed after it, under its new name, in the same pinned position.
8. **#404 needs `PremiumFlags.paywallVisible = true`** and no Pro entitlement. With the flag flipped, long-press the mic: expect **four equal rows** — `Normal ✓`, `Liste PRO`, `→ EN PRO`, `✨ Dictus Pro ›` — and **no reason strip**.
9. **Release on `Normal`.** It must arm nothing, clear any armed mode, and **start a recording**, exactly as it does for a subscriber. This is the function the one-row shape removed.
10. **Release on `Liste` or `→ EN`.** Refused, with "Les modes font partie de Dictus Pro." in the toolbar. Nothing armed.
11. **Release on `Dictus Pro`.** DictusApp opens **on the paywall**, as a full-screen cover. Nothing armed or disarmed.
12. **Also check the hamburger panel's Dictus Pro pill** — it sends the same intent and now lands on the paywall too.
13. **As a Pro subscriber who switched Smart Modes off**, long-press: the **ordinary** fan with `INACTIF`, never `PRO`, and never the Dictus Pro row.
14. **Put `paywallVisible` back to `false`** and confirm the fan returns to today's greyed rows and reason line, with no Dictus Pro row anywhere.

**On an iPhone SE the four-row geometry is reasoned about, not seen** — the arithmetic (46.7 pt against the 44 pt minimum) is asserted in `SmartModeFanLayoutTests`, and there is no SE here to look at. If one is ever available, item 8 is the check.

## What changed since the device test

**The fan keeps its shape for everyone.** `SmartModeFanLayout.entries(offersProUpgrade:)` now returns `Normal` + the default pins + `Dictus Pro` instead of `[.pro]`. Four rows is the ceiling `maximumPinnedModes` already measured; the Pro row takes a slot, so two mode rows remain, and they are the **seed** rather than the stored list — a non-subscriber cannot reach the mode list, so the fan promises exactly what they would get on subscribing. The no-dead-space rule is untouched: rows still divide the area evenly and completely, which never required *one* row.

**Two things stopped being the same value.** `unavailableReason` drives the strip and therefore the geometry, and it is nil in this fan because the Pro row is the message. Whether a mode may be armed is now `modesAreArmable`. Inferring one from the other — which is what the code did — would have drawn these rows enabled and let a release arm them.

**Three row tags, one precedence, in DictusCore.** `SmartModeFanLayout.tag` resolves the check (what will run), `INACTIF` (armed and it will not, #423) and `PRO` (behind the subscription). `PRO` beats `INACTIF` when both would apply, which is #404's own rule — `INACTIF` is for a subscriber who switched Smart Modes off, and they already pay. The `SmartModeEntitlement` split from the first round is what makes that distinction expressible at all, and it is upheld.

**The Pro row lands on the paywall.** Nothing routed on `dictus://open` at all, so both Pro entries brought the app forward on whatever screen it was showing. `KeyboardOpenURL` gives the two processes one vocabulary — the keyboard builds the URL, the app parses it with the same type, and the round trip is a test instead of two strings that drifted. `MainTabView` answers `.pro` with `paywallCover`, gated on `PremiumFlags.paywallVisible` on the app side too. The hamburger panel's pill (#241) is fixed by the same route.

### #404, against the amended criteria

|---|---|---|
| `Normal`, the default modes greyed with `PRO`, and a `Dictus Pro` row — not a single row | ✅ simulator | Screenshot: four rows, `Normal ✓`, `Liste PRO`, `→ EN PRO`, `✨ Dictus Pro ›`, no reason strip. Log: `smartModeFanOpened entries=4 reason=notSubscribed`. |
| `Normal` is releasable and starts a recording | ✅ simulator | `smartModeArmed mode=normal` then `startRecording`, and the stored mode cleared. |
| Row count never exceeds four; same layout a subscriber sees | ✅ unit + simulator | `testTheUpgradeFanNeverExceedsTheRowCeiling` asserts the SE row clears 44 pt; the capture shows four equal rows. |
| A switched-off subscriber still gets `INACTIF`, never `PRO` | ⚠️ unit only | `testProBeatsInactiveOnTheSameRow` and `testAnArmedModeThatWillNotRunIsTaggedInactive`. Reaching it live needs `proActive = true`, which `SubscriptionManager` resets without a StoreKit entitlement. Manual step 13. |
| Releasing on the Dictus Pro row opens DictusApp **on the paywall** | ✅ simulator, end to end | One run: `smartModeFanProSelected` → `openDictusApp intent=pro` → `keyboardProIntent source=keyboard` → screenshot of the paywall cover over "◀ Réglages". |
| Releasing on it arms and disarms nothing | ✅ code + simulator | `commit()` switches over the case and returns before any store write; the run above left the armed value untouched. |
| Still gated behind `PremiumFlags.paywallVisible` | ✅ | With the shipping `false`, the simulator showed today's greyed rows and reason line and no Pro row. The flag is untouched in this branch — `git diff` does not list `PremiumFlags.swift`. |

### #423 and the rename

Both were device-validated by the maintainer on 2026-08-29, including the `.switchedOff` copy this PR previously flagged as unverified. Unchanged in this round.

## Verification

- `cd DictusCore && swift test` — **1317 tests, 1 skipped, 0 failures**.
- `swiftlint lint --strict` — clean.
- `xcodebuild build` for all three targets — **BUILD SUCCEEDED**, on the shipping `paywallVisible = false`.
- Rebased on `origin/develop` (`6e61626`) with no conflicts.
- Simulator (iPhone 17, iOS 26.5, headless): the four-row fan, the `Normal` release, and the keyboard-to-paywall chain, all with `paywallVisible` flipped **locally only** and reverted before committing.

## One correction worth recording

While verifying the paywall I read `axe describe-ui` and concluded five times that the cover was not presenting. **It was.** `describe-ui` does not traverse into a `fullScreenCover`, so it kept returning the Home screen underneath. Three changes made on that false premise — a deferred `DispatchQueue.main.async` presentation, and moving the cover from the `ZStack` to the `TabView` with a comment citing a measurement that never happened — have been reverted. A screenshot settles this and the accessibility tree does not; `docs/agents/simulator.md` does not yet say so.

## Not verified

- The `PRO`-versus-`INACTIF` distinction live, for the reason in the table.
- The four-row geometry on an SE-sized screen.
- Anything on a physical device — the maintainer's list above.

## Summary by CodeRabbit

* **New Features**
  * Smart Mode now explains when an unavailable mode is skipped and a fallback is used.
  * Added localized Smart Mode names, including the “List” label.
  * Added Pro upgrade rows and status indicators in the Smart Mode menu.
  * Added direct navigation from the keyboard to Pro and Settings screens.
  * Added a paywall display when selecting Pro features.
* **Bug Fixes**
  * Prevented stale Smart Mode skip notices from reappearing.
  * Improved Smart Mode availability handling when features are switched off.
  * Clarified toolbar indicators when an armed mode will not run.


Pivii:
## CodeRabbit triage

**Fixed — `DictationHandoff.swift:76`.** The fallback `"Smart Mode"` now goes through `String(localized:)`. The key already existed in the app catalog with its French value (`Mode intelligent`), shared with `ProFeature.displayName`, so only the call site changed.

**Fixed — `SmartModeFanView.swift:152` and `:224`.** `PRO` and `Dictus Pro` are localized `Text` values with catalog entries in both languages. Both are identical in French today; they are in the catalog so a future locale can change them, which a `verbatim` literal makes impossible.

**Rejected — `SmartModeFanLayout.swift:31`.** The claim is that the Pro row renders as `Normal` because `name(_:)` maps every nil `smartMode` to `"Normal"`. It does not reach `name(_:)`: `SmartModeFanView.body` routes `.pro` to `proRow(isHighlighted:)`, which renders its own label and never calls it. The device capture on `e56ad4c` and the simulator screenshot both show the row reading `Dictus Pro`. The finding reads `name(_:)` in isolation. The localisation fix above touches that same label and deliberately keeps the `.pro` branch rendering it directly, so the fall-through it describes still does not exist.

**These are string changes only — no behaviour changed**, so the device validation of the four-row fan, the `Normal` release, the `PRO` tags, the paywall landing and the `INACTIF` treatment stands. `swift test` 1317 passing, `swiftlint lint --strict` clean, all three targets build. `Localizable.xcstrings` carries the two new keys and no build-regeneration noise.

**Follow-up, not done here:** `ToolbarView.proEntry` uses `Text(verbatim: "Dictus Pro")` for the hamburger panel's pill and its accessibility label. Same defect, pre-existing and outside this diff.

Issue #449, opened 2026-08-28
Onboarding model download stalls in background and restarts the 445 MB file from byte 0


Pivii:
## Source

User report and exported device log, 2026-08-28.

- Version/build: `1.8.0 (27)`
- Commit/rev: `a896107@develop`
- iOS/device: `iOS 26.6`, `iPhone16,2` (iPhone 15 Pro Max)
- Model: `parakeet-tdt-0.6b-v3`

## Bug

The onboarding model download is a foreground-only `URLSession`. If the user leaves Dictus, iOS suspends the app and the current large-file transfer stops. When the app becomes active again, the 30 second stall detector fires, the retry starts the current file again from byte 0, and enough foreground/background cycles eventually fail with `NSURLErrorNetworkConnectionLost`.

This makes onboarding look as if the model download resets every time the user checks another app. It also forces the user to keep Dictus open for several minutes.

### Attempt 1

- `10:44:01` download starts, catalogue announces 800 MB
- `10:44:04` server total is 483 MB
- `10:44:26` reaches 20%, 98 MB
- app is backgrounded during the transfer
- on return, `Encoder.mlmodelc/weights/weight.bin` reports two 30 second stalls
- `10:46:20` fails with `La connexion réseau a été perdue.`

### Attempt 2

- `10:46:26` restarts at 0%
- reaches 60%, 289 MB
- another background/foreground cycle produces two stalls
- `10:49:18` fails with the same connection-lost error

### Attempt 3

- `10:49:20` restarts at 0%
- user keeps Dictus active
- `10:53:19` reaches 483/483 MB
- Parakeet compiles successfully in 17.0 seconds

Measured from the exported log:

| Attempt | Last progress | Duration | Result |
| --- | ---: | ---: | --- |
| 1 | 98 MB / 20% | 139 s | failed |
| 2 | 289 MB / 60% | 172 s | failed |
| 3 | 483 MB / 100% | 239 s | succeeded |

The successful foreground transfer averaged about 16.3 Mbit/s. No HTTP 429 or 503 appears in the log.

## Confirmed code mechanism in build 27

`DictusApp/Models/ModelManager.swift` explicitly documents the download as foreground-only: background `URLSession` was deferred and the implementation assumes the user keeps the app open.

[...]

PR #468, opened 2026-09-01
Background model download with durable byte-range resume (refs #449)

Pivii:
Refs #449.

## What this was for

A first-run user starts the Parakeet download, checks another app while half a
gigabyte arrives, and comes back to a progress bar at 0%. The transfer was a
foreground-only `URLSession`, so iOS suspended it with the process; the 30 s stall
detector met the user on their return and the retry restarted the current file at
byte zero. On Parakeet that file is `Encoder.mlmodelc/weights/weight.bin`, 445 MB of
a 483 MB payload, so the whole download effectively restarted. The exported log from
2026-08-28 has three attempts — 98 MB, 289 MB, then the full 483 MB — and the only
one that finished was the one where the phone was left alone for four minutes.

The ask: a user can leave Dictus while a model downloads, and the transfer either
keeps going under iOS background rules or resumes from the last received byte.
Returning shows durable progress, and reopening onboarding cannot start a second
download of the same model.

## To test by hand

Everything below needs a physical device: the simulator does not suspend apps the
way a device does, and `simctl terminate` does not cancel background tasks the way a
user force-quit does. Install this branch, then in Dictus Settings > Debug log,
"Effacer" before each run so the log is easy to read.

1. **Background for 60 s.** Fresh install (delete Dictus first). Run onboarding to
   "Modèle vocal", tap "Installer le modèle", wait until the bar passes 10%, then go
   to the home screen and leave Dictus for 60 seconds. Come back.
   *Expect:* the bar is at or above where you left it, never 0%. The MB counter has
   moved forward, not backward.
2. **Three background cycles, one transfer.** Repeat step 1 three times during the
   same download, then let it finish.
   *Expect:* the model finishes and the card says it is ready. In the exported log,
   `modelDownloadChunk` lines for `Encoder.mlmodelc/weights/weight.bin` cover chunks
   0 to 13 with no index appearing twice, and there is exactly one
   `modelDownloadStarted` line for the model.
3. **Force quit mid-file.** Start the download again on a fresh install. When it
   passes 30%, swipe up and force-quit Dictus from the app switcher. Wait 10 s.
   Relaunch it.
   *Expect:* the screen comes back showing roughly the percentage you left at, not
   0%. The log carries `modelDownloadSessionRestored`, then
   `modelDownloadResumed … offset=<N>MB source=relaunch` where N is close to where
   you quit, and the download continues to 100%.
   *Corrected by the device test of 2026-09-01:* the background tasks SURVIVE a user
   force-quit on device. The log said `modelDownloadSessionRestored tasks=2 models=1`,
   not `tasks=0` as this paragraph originally predicted. The resume therefore does not
   depend on the bytes on disk alone — the system hands the live transfers back — and
   the durable manifest is what covers the case where it does not.
4. **Airplane mode mid-file.** Start a download, let it pass 20%, turn on airplane
   mode for 30 s, turn it off.
   *Expect:* the download resumes rather than restarting. If it could not resume, the
   log says why, in a `modelDownloadRangeRejected` line naming the rule that refused
   it. A `modelDownloadIntegrityFailed` line would mean a resume appended bytes that
   did not belong together — that one should never appear, and if it does the issue
   needs reopening.
5. **Cellular.** Repeat step 1 once on cellular rather than Wi-Fi, then export the
   debug log and attach it to the issue. The acceptance criterion asks for Wi-Fi and
   cellular with an exported log.
6. **Delete partial download.** Start a download, let it pass 20%, background the app
   and force-quit it. Relaunch, let it resume, then from the model card's overflow
   menu choose "Supprimer le téléchargement partiel".
   *Expect:* the card returns to "Available" and the next download starts from 0%,
   not from the partial. (Check storage in iOS Settings if you want to confirm the
   space came back.)
7. **A Whisper model, not just Parakeet.** In Settings > Modèles, download Whisper
   Small and background the app once during it.
   *Expect:* the same resume behaviour. The service is shared by both engines.

## What changed

**DictusCore (pure, unit tested)**

- `HTTPRangeResume` — parses `Content-Range` and decides whether a response may be
  appended at the offset we hold. `206` with an exact start appends; `200` restarts
  cleanly; a `206` starting elsewhere or a resource that changed size is refused.
- `ModelDownloadManifest` — the durable record: repository, pinned revision, per-file
  size, SHA-256, bytes appended, ETag. Plus the chunk arithmetic and the
  `taskDescription` codec that lets a relaunched process recognise its own tasks.
- `ModelFileIntegrity` — streaming SHA-256 and size check before publication.
- Five new log events for the transfer.

**DictusApp**

- `BackgroundModelDownloadService` — one `URLSessionConfiguration.background` for the
  whole app. Files move as 32 MB aligned ranges, two in flight; each chunk is moved
  out of the system temp location as it lands, which is what makes progress durable.
  Appends truncate to the manifest offset first, so a crash mid-append cannot double
  a chunk. Completed files are still published by one atomic move.
- `ModelRepoDownloader` — keeps the repository knowledge and the completeness
  tripwire, and pins every URL to a commit SHA. It no longer moves bytes.
- `AppDelegate` — recreates the session at launch and holds the completion handler
  from `handleEventsForBackgroundURLSession`.
- `ModelManager` — seeds state and progress from the manifest on disk, re-enters
  `downloadModel` for anything unfinished (once per process, never from a background
  launch), and keeps both `ModelManager` instances in agreement about what is
  preparing.

## Deviation from the issue's implementation direction

Point 4 asks for URLSession resume data first, with HTTP Range as the fallback. This
implements Range as the primary mechanism and does not use resume data at all.

Resume data cannot carry acceptance criterion 3. iOS cancels a background session's
tasks when the user force-quits, does not relaunch the app, and hands resume data
only to a running process through `didCompleteWithError` — a killed process receives
nothing. The maintainer's follow-up comment on the issue specifies the Range path in
full (canonical `/resolve/<commit>/<path>`, `Range` + `If-Range`, require 206 with an
exact `Content-Range` start, restart cleanly on 200, validate size and SHA-256) and
never mentions resume data, so that is what is built. The two are deliberately not
mixed: resume data replays a request with a `Range` header of its own, and its
interaction with ours is not something worth debugging in the field.

If you would rather have resume data as a fast path inside a chunk, say so and it can
be added — but it cannot replace the range path without losing criterion 3.

## Acceptance criteria

| --- | --- | --- |
| Background 60 s, progress does not reset | Verified on device | 4 minutes across other apps, no reset; `modelDownloadCompleted` at 12:28:20 and Parakeet loaded `loading → ready`, warm inference 112 ms. |
| Three cycles, one transfer, no duplicate tasks | Partly | One run per model is enforced by the service and by `ModelDownloadTaskTag`; the duplicate-chunk case was found and fixed on the simulator. The backgrounding half is step 2 above. |
| Force-quit, relaunch, resume without repaying | Verified on device | Force-quit at ~67%, relaunch: `modelDownloadSessionRestored tasks=2 models=1`, `resumedAfterRelaunch percent=67`, download completed. The two background tasks survived the process, which contradicts what this PR first assumed. |
| Connection loss retries from durable progress or records why | Not verified | The decision logic is unit tested against the live repository's real headers; the field behaviour is step 4 above. |
| Completed files atomically published, incomplete ones fail verification | Verified | One `moveItem` per file, unchanged; the Phase 5 tripwire is unchanged; `ModelFileIntegrity` added in front of it. Published `weight.bin` measured at exactly 445187200 bytes. |
| Same service for Parakeet and WhisperKit | Verified by construction | `ModelRepoDownloader.Configuration` is unchanged and both engines route through the one service. Parakeet exercised end to end; Whisper is step 7 above. |
| On-device Wi-Fi and cellular with an exported log | Not verified — device only | Step 5 above. |

## Verification run here

- `xcodebuild … -scheme DictusApp` and `-scheme DictusKeyboard` — both BUILD SUCCEEDED.
- `cd DictusCore && swift test` — 1486 tests, 0 failures.
- `swiftlint lint --strict` — 0 violations.
- End to end on a headless iPhone 17 Pro simulator (iOS 26.5): onboarding download of
  Parakeet v3, 483 MB, chunked 14 ways across the encoder weights, all `206` with
  validated ETags, killed at 30%, relaunched, resumed at the durable offset,
  published, compiled in 3.3 s, selected, Continue offered.

---

## Device test round 1 (2026-09-01, on `29df47f`) — two defects found and fixed

Fixed in `66a05c5`:

- **A valid 206 was refused, and "erreur 206" reached the user.** The background
  `URLSession` resumes its own tasks: iOS suspended the app 2 996 625 bytes into an
  aligned 32 MB chunk, the system re-issued the request from there, stitched the legs
  and delivered the complete chunk with a `Content-Range` describing only the second
  leg. Reading that header as the body's start refused a good chunk — in exactly the
  case this PR exists for — and cost ~37 MB per suspension. A start inside the
  requested span is now accepted as a client-side resume, and the assembled chunk is
  validated by its LENGTH instead of by the header. A refused range is also no longer
  reported as an HTTP error: 206 is a success, and the case is retryable.
- **`modelDownloadStarted` logged three times for one download**, because a resume
  re-enters `downloadModel`. It now logs once per transfer, which is what makes
  acceptance criterion 2 checkable by its own method. No duplicated work was ever
  happening: the 14 chunks each appeared exactly once.

Also confirmed green on device: long backgrounding without reset, force-quit and
automatic resume, zero `modelDownloadIntegrityFailed`, every chunk `206` with a
validated ETag from `us.aws.cdn.hf.co`, warm inference at 112 ms.


Pivii:
@coderabbitai review

Issue #438, opened 2026-08-27
The Parakeet download tripwire accepts an empty .mlmodelc directory as a completed file


Pivii:
## What is wrong

`ModelRepoDownloader.Configuration.parakeet()` declares its final-verification paths as directory names:

```swift
requiredPaths: { _ in ModelNames.ASR.requiredModels + [ModelNames.ASR.vocabularyFile] }
```

`ModelNames.ASR.requiredModels` are the `.mlmodelc` bundle names. Phase 4 of `download` checks each with `FileManager.fileExists`, which answers `true` for a directory. And this downloader creates a file's parent directory *before* it fetches the file:

```swift
try FileManager.default.createDirectory(
    at: destination.deletingLastPathComponent(),
    withIntermediateDirectories: true
)
```

So `Encoder.mlmodelc/` exists on disk from the moment the first byte of its first file is requested. A Parakeet download interrupted inside `Encoder.mlmodelc/weights/weight.bin` — which is about 92% of the payload, and therefore where an interruption almost always lands — clears the tripwire with an empty bundle, and the downloader hands an incomplete cache to `AsrModels.load`.

The tripwire exists precisely to "fail loudly here instead of silently handing an incomplete cache to the engine", per its own comment. On this path it does not.

## Why it is filed separately

PR #435 fixed the identical hole on the WhisperKit path for #433: `WhisperModelRepository.requiredDownloadPaths` now names the leaf files inside each bundle, and reads the bundles off the variant folder so it covers whatever the variant shipped. Parakeet was explicitly excluded from #433 by decision, because FluidAudio caches one directory per `AsrModelVersion` shared by every model of that version, which makes "are its files complete" a different question there. The tripwire hole is real on both paths, but only the WhisperKit half belonged in that PR.

Refs #433. See also PR #435 for the shape of the fix on the other engine.

## What a fix has to establish

- The leaf entries a compiled Parakeet bundle must hold. Do not copy WhisperKit's `coremldata.bin` / `model.mil` / `weights/weight.bin` without checking: FluidAudio's bundles come from a different repository and have not been inspected for this.
- Whether `AsrModels.modelsExist`, which this tripwire says it mirrors, has the same weakness. If it does, the mirror is faithful and the question becomes whether to diverge from it deliberately.
- Whether the vocabulary file is a real file (it is, so it is fine as declared).

## Acceptance criteria

[...]

PR #479, opened 2026-09-02
Refuse a Parakeet bundle directory as a completed download (refs #438)

Pivii:
refs #438

## What this was for

The Parakeet download had a final check whose job is to refuse an incomplete cache before it reaches the engine, and it could not do that job. It listed what it required as the four `.mlmodelc` **bundle names**, and it tested them with `FileManager.fileExists`, which says yes to a directory. Those directories exist long before they hold a model. So a Parakeet download that stopped inside `Encoder.mlmodelc/weights/weight.bin` — 445 MB of the 482 MB payload, which is where an interruption almost always lands — passed the check and was handed to FluidAudio as a finished download.

What happens next is the part worth fixing: FluidAudio's own cache check has the exact same blind spot (`fileExists` on the same four directory names). It finds them, decides nothing needs downloading, throws while loading the model, and its catch block **deletes the whole model directory and downloads it again** — 482 MB, on the dictation hot path, with no progress bar and no way to cancel. That is precisely what #252 built a guard to prevent, and the guard was not strict enough to see it either.

## To test by hand

Everything below needs the physical iPhone: Parakeet's compile is Apple Silicon / ANE work, and interrupting a 482 MB transfer is not something a headless run can produce faithfully. Nothing here changes any UI, so there is nothing to look at; what you are checking is that the app refuses a broken download and still accepts a good one.

1. Install this branch on the iPhone, open Dictus, go to **Modèles**, and make sure **Parakeet** is already downloaded and working from before this branch. Do a short dictation with it.
   → Expected: dictation works exactly as it did. This is the one that matters: the completeness check got stricter, and a cache that was fine yesterday has to still read as fine today. If Dictus instead says the model is not installed, this PR is wrong and should not be merged.
2. Delete Parakeet from **Modèles**, then start downloading it again and let it run to 100%.
   → Expected: the download finishes and the model prepares as before. The stricter check must not fire on a healthy download.
3. Do a dictation with the freshly downloaded Parakeet.
   → Expected: normal transcription.
4. Delete Parakeet again, start the download, and **turn on airplane mode when the bar is somewhere between 10% and 90%** (that stretch is the encoder weights).
   → Expected: the download stops with an error and offers to retry. What must never happen is the model landing in **Téléchargés** as if it were finished.
5. Turn airplane mode back off and retry the download from the model card.
   → Expected: it resumes rather than restarting from zero, and finishes.
6. Repeat step 4, but instead of airplane mode, **force quit Dictus** (swipe up in the app switcher) mid-download. Reopen Dictus and go to **Modèles**.
   → Expected: Parakeet is NOT under Téléchargés, and its card offers to download or resume. Then tap the microphone in the keyboard and try to dictate with whatever model is active.
   → Expected: no long silent freeze while something re-downloads half a gigabyte in the background. This is the failure this PR closes off.
7. Export the debug log (Réglages → Debug) after steps 4 and 6.
   → Expected: no `modelDownloadCompleted` line for `parakeet-tdt-0.6b-v3` from either interrupted attempt.

If you want to see the tripwire itself fire rather than trust it, there is no user-facing way to do it: it now sits behind the transfer's own accounting, which usually fails first. Its job is the case where the transfer *believes* it succeeded — see "One thing this does not claim" below.

## What changed

**`DictusCore/Sources/DictusCore/ParakeetModelRepository.swift`** — one declaration of what a complete Parakeet bundle holds, and a repo-relative path list built from it.

`requiredBundleEntries` is `coremldata.bin`, `model.mil`, `weights/weight.bin`. Established from the repository, not copied from the WhisperKit fix, as the issue asked. The tree of `FluidInference/parakeet-tdt-0.6b-v3-coreml` at revision `7dd20fe` was read on 2026-09-02: all four bundles FluidAudio requires — `Preprocessor`, `Encoder`, `Decoder`, `JointDecision` — carry those three, plus `metadata.json` and `analytics/coremldata.bin`. The last two are deliberately left out. `Encoder_v2` and `ParakeetDecoder` in the same repository ship without `metadata.json`, which makes it the least stable entry of the five, and requiring it catches nothing the weights do not already catch. (This is the same reasoning that kept `model.mlmodel` out of the WhisperKit list, reached independently rather than inherited.)

`requiredDownloadPaths(requiredModelBundles:vocabularyFileName:)` turns that into the leaf paths a finished download owes. It is a fixed list rather than the disk enumeration the WhisperKit side does, and the difference is deliberate: there the set of bundles is a property of the variant and only knowable from the files, while here it is `ModelNames.ASR.requiredModels`, a compile-time constant the caller hands over. The FluidAudio cache is also one directory per `AsrModelVersion` shared by every model of that version, so enumerating it would demand completeness of bundles this downloader never fetched.

`isCompiledModelBundle` now requires every entry instead of the marker alone.

**`DictusApp/Models/ModelRepoDownloader.swift`** — `Configuration.parakeet()` takes its required paths from that function, and Phase 5 refuses a path that resolves to a directory whatever the list names. Both engines have now shipped a required-path list that named a directory (#433 for WhisperKit, this one for Parakeet), and a check that cannot tell a file from a directory will accept the third.

## The three questions the issue asked to establish

1. **The leaf entries a compiled Parakeet bundle must hold** — read off the repository tree, listed above with the reason two entries are excluded.
2. **Does `AsrModels.modelsExist` have the same weakness** — yes. FluidAudio 0.12.4, `AsrModels.swift:366-381`: `fileExists` on the four bundle directory names plus the vocabulary. The mirror was faithful and worthless, so this diverges from it deliberately, and the divergence is written down at the call site.
3. **Is the vocabulary a real file** — yes. `ModelNames.ASR.vocabularyFile` is `parakeet_vocab.json`, 151 122 bytes at the repository root, and it is the file `AsrModels.loadVocabulary` actually reads (`Names.vocabulary(for:)` returns it for every version). Fine as declared, unchanged.

## Beyond the acceptance criteria, declared

The three criteria name the tripwire only. Hardening `isCompiledModelBundle`, which is the load-path guard from #252, goes one step past them. It is in this PR because it is the same defect one layer down, its failure mode is the worse of the two (the 482 MB wipe-and-re-download described at the top), and leaving it would have left two disagreeing definitions of "complete" in the same file with the weaker one believed — the mistake PR #435's review round was spent on. It is one constant and one `allSatisfy`, and it is removable on its own if you want only the narrow fix.

Its risk is the mirror image: a stricter load guard could report a genuinely complete cache as absent and send someone with 482 MB on disk back to the model manager. Evidence against it is criterion 4 below.

## One thing this does not claim

PR #468 changed the mechanism the issue describes. The downloader no longer creates a file's parent directory before requesting its bytes; `BackgroundModelDownloadService.publish` creates it when a **complete** file is moved into place. The defect survived anyway — the repository is listed breadth first, so every bundle's small top-level files are published before any bundle's `weights/`, and a stopped transfer leaves four bundle directories holding their markers and no model — but the bundle is no longer *empty* as the issue says, it is hollow.

The other consequence of #468 is that an interrupted transfer now fails inside the transfer's own accounting, before Phase 5 is reached. So this PR does not fix a crash a user is hitting today on the interrupt path; it repairs the backstop for the case the accounting cannot see — a repository listing that comes back short of a file, or a cache that changed under a manifest — where the tripwire is the only thing standing between an incomplete cache and FluidAudio's wipe-and-re-download. That is what the tripwire's own comment says it is for.

## Acceptance criteria and evidence

| --- | --- | --- | --- |
| 1 | A download interrupted inside a bundle's weights fails the tripwire instead of passing it | **Verified (unit)** | `testADownloadStoppedInsideTheEncoderWeightsFailsTheTripwire` builds that exact on-disk shape, asserts all four bundle directories exist (which is why the old list passed), and asserts the required-path check stops on `Encoder.mlmodelc/weights/weight.bin`. `testARequiredPathThatIsADirectoryFailsTheTripwire` covers Phase 5's new refusal of a directory |
| 2 | A completed download still passes, verified against the real repository file list | **Verified** | All 13 required paths checked against the live repository listing at revision `7dd20fe` (2026-09-02): 13/13 present. Then checked against a real complete Parakeet cache on the Mac (`~/Library/Application Support/FluidAudio/Models/parakeet-tdt-0.6b-v3-coreml`, downloaded 2026-08-25 and previously loaded): 13/13 present as regular files, byte sizes identical to the repository. Plus `testACompletedDownloadSatisfiesEveryRequiredPath` |
| 3 | The required entries come from the repository tree or FluidAudio's own code, not assumption | **Verified** | Tree API readings above, pinned in `testRequiredDownloadPathsCoverEveryBundleAndTheVocabulary`. Bundle set and vocabulary name read from FluidAudio's `ModelNames.ASR` at the call site rather than restated |
| 4 | The stricter load guard does not refuse a healthy cache | **Partly verified** | The real Mac cache in criterion 2 satisfies every entry. Not verified on the iPhone — step 1 of the manual list |
| 5 | A real interrupted Parakeet download behaves | **Not verified** | Needs the device. Steps 4-7 of the manual list |

Commands run in the worktree:

- `cd DictusCore && swift test` — **1553 tests, 1 skipped, 0 failures**. `ParakeetModelRepositoryTests` goes from 13 tests to 20
- `swiftlint lint --strict` — **0 violations, 0 serious in 241 files**
- `xcodebuild -project Dictus.xcodeproj -scheme DictusApp -destination 'generic/platform=iOS Simulator' -derivedDataPath build/DerivedData build` — **BUILD SUCCEEDED**, producing `DictusApp.app`, `DictusKeyboard.appex`, `DictusWidgets.appex` and `DictusCore.o`
- FluidAudio resolved at `0.12.4` (revision `9830ce83`), which is the checkout every line quoted above was read from

Not merged: the 1.8.1 TestFlight cut is pending and nothing may land before it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model download verification to confirm that all required model files are present.
  * Incomplete downloads, missing weight files, and directories in place of required files are now correctly detected.
  * Model availability checks now provide more reliable results for partially downloaded or corrupted model bundles.

### DIFF
--- a/DictusApp/Models/ModelRepoDownloader.swift
+++ b/DictusApp/Models/ModelRepoDownloader.swift
@@ -82,14 +82,38 @@ final class ModelRepoDownloader {
         /// declared its download verified at the one moment it most likely was not.
         let requiredPaths: @Sendable (URL) -> [String]
 
-        /// Parakeet v3 repo. Matches FluidAudio's `Repo.parakeet.remotePath`,
-        /// its file selection, and mirrors `AsrModels.modelsExist` verification.
+        /// Parakeet v3 repo. Matches FluidAudio's `Repo.parakeet.remotePath` and its
+        /// file selection.
+        ///
+        /// Required paths moved to `ParakeetModelRepository` for issue #438, and they
+        /// deliberately stop mirroring `AsrModels.modelsExist`. That function checks
+        /// `fileExists` on the four bundle DIRECTORY names, which this tripwire used to
+        /// copy — and a directory is not a file. The bundles are created here as soon as
+        /// their first small file is published, and because `listRequiredFiles` walks the
+        /// repository breadth first, every bundle's `coremldata.bin` and `model.mil` are
+        /// on disk before any bundle's `weights/`. A transfer stopped inside
+        /// `Encoder.mlmodelc/weights/weight.bin` — 445 MB of the 482 MB payload, so where
+        /// an interruption almost always lands — therefore left four bundle directories
+        /// standing, none of them holding a model, and the tripwire waved them through to
+        /// `AsrModels.load`. FluidAudio's own cache check has the same blind spot, so the
+        /// mirror was faithful and worthless; the leaf list is the deliberate divergence.
+        ///
+        /// The bundle set stays FluidAudio's `ModelNames.ASR.requiredModels` rather than a
+        /// reading of the cache directory (which is what the WhisperKit configuration
+        /// below does): that cache holds one directory per `AsrModelVersion`, shared by
+        /// every model of the version, so what is on disk is not the same question as
+        /// what this download owed.
         static func parakeet() -> Configuration {
             Configuration(
                 repoPath: Repo.parakeet.remotePath,
                 directoryPatterns: ModelNames.ASR.requiredModels.map { "\($0)/" }.sorted(),
                 includesRootMetadata: true,
-                requiredPaths: { _ in ModelNames.ASR.requiredModels + [ModelNames.ASR.vocabularyFile] }
+                requiredPaths: { _ in
+                    ParakeetModelRepository.requiredDownloadPaths(
+                        requiredModelBundles: ModelNames.ASR.requiredModels,
+                        vocabularyFileName: ModelNames.ASR.vocabularyFile
+                    )
+                }
             )
         }
 
@@ -250,14 +274,21 @@ final class ModelRepoDownloader {
             onProgress: onProgress
         )
 
-        // Phase 5: parity tripwire — mirror each engine's own existence check
-        // (AsrModels.modelsExist for Parakeet, the CoreML bundle set for WhisperKit)
-        // so we fail loudly here (with a localized error) instead of silently
-        // handing an incomplete cache to the engine. If the repo layout ever
-        // drifts, this is the signal.
+        // Phase 5: the tripwire — every leaf file a finished download of this engine
+        // owes has to be on disk, so we fail loudly here (with a localized error)
+        // instead of silently handing an incomplete cache to the engine. If the repo
+        // layout ever drifts, this is the signal.
+        //
+        // A directory is not a file, and both engines learned that the hard way
+        // (#433, #438): a required path that names a `.mlmodelc` bundle is satisfied
+        // by the empty shell an interrupted download leaves. Both lists name leaf
+        // files now, and this check refuses a directory whatever they name, so the
+        // hole cannot be reopened from the list alone.
         for requiredPath in configuration.requiredPaths(cacheDir) {
             let path = cacheDir.appendingPathComponent(requiredPath)
-            guard FileManager.default.fileExists(atPath: path.path) else {
+            var isDirectory: ObjCBool = false
+            let exists = FileManager.default.fileExists(atPath: path.path, isDirectory: &isDirectory)
+            guard exists, !isDirectory.boolValue else {
                 throw DownloadError.missingRequiredFile(requiredPath)
             }
         }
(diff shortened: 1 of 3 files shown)


Pivii:
@coderabbitai review

Issue #467, opened 2026-09-01
After a keyboard dictation, the app opens on a stale result overlay showing the raw


Pivii:
## Summary

The "Last transcription" card on the app's home screen shows the **raw** STT output, not the polished text the keyboard actually typed. The card is a copy-to-clipboard surface, so the user copies a version of their dictation that exists nowhere else and that they never chose.

Reported by the maintainer, reproducible on demand, and easiest to see on a mixed-language dictation: the document received clean French, the home card kept the franglais raw.

**Read the next section before planning a fix.** The obvious cause — "the keyboard never sends the polished text back to the App Group" — is measurably false. That path exists, and it ran on every dictation in the capture below.

## What the evidence already rules out

Captured on device, App 1.8.1 (28), `rev 69c5404@HEAD` (PR #463's branch), iPhone 15 Pro Max, iOS 26.6.1, 2026-09-01.

The keyboard writes the typed text to `SharedKeys.lastPolishedTranscription` in `KeyboardPolishCoordinator.finish(text:insert:)`, posts `polishDidFinish`, and DictusApp's `DictationHandoff.polishHandoffFinished()` reads it and assigns `lastResult = typed`. `HomeView` renders `coordinator.lastResult` from `DictationCoordinator.shared`, a single shared instance reached through `@EnvironmentObject`.

That handler **ran on all 8 keyboard dictations** in the session, from the app process, with the polished character count:

```
09:15:42 <APP> polishHandoff step=handedOff outcome=raw      chars=176
09:15:42 <KBD> polishHandoff step=claimed   outcome=pending  chars=176
09:15:45 <APP> polishHandoff step=finished  outcome=inserted chars=194
```

`chars=194` is the polished text (193 in the debug export, plus the one character `DictationTail` adds). `outcome=inserted` is emitted only on the branch where `typed` is non-nil, which is the same branch that assigns `lastResult`. Every one of the 8 dictations logged the same shape, each with its polished length.

So at the moment the log was written, `DictationCoordinator.shared.lastResult` held the polished text — and the card still showed the raw.

The three writes to `lastResult`, for reference:

| site | value | when |
|---|---|---|
| `DictationCoordinator.startDictation` | `nil` | every dictation start |
| `DictationHandoff.handOffToKeyboard` | **raw** | when the keyboard claims the dictation |
| `DictationHandoff.polishHandoffFinished` | **polished** | when the keyboard reports what it typed |

[...]

PR #491, opened 2026-09-05
Land the polished text on the home card: return to idle, and stop deleting a refused result (refs #467, #495)

Pivii:
Two issues, one PR, because they share a surface — the home card — and one device test. **#467** is why the card is in front of the user at all; **#495** is why the text on it is the right one.

## What this was for

**#467 — the app opened on a stale overlay.** After a keyboard dictation, opening DictusApp landed on the recording overlay showing the **raw** transcription, while the host document had the polished text. Confirmed on device 2026-09-04 (`rev c08136f`, Translate → English armed, spoken French): the document got the English, the screen the app opened on showed the French. The home card was never the defect — it shows the polished text and always did. It was **covered** by a recording overlay that a completed keyboard hand-off never took down.

**#495 — a refused insertion threw away the text it had produced.** When the keyboard refuses to insert (the user left the field, switched app, dismissed the keyboard), the polished text was already generated and then deliberately deleted. The app was told "nothing was typed", fell back to the raw, and the card showed French where the user armed Translate → English. Confirmed on device 2026-09-05, on this branch at `f4d9a3d`: `outcome: success`, `engineMs: 2769`, 181 characters of English in the polish debug — and `polishHandoff step=finished outcome=not-inserted chars=0` in the log at the same second. The Smart Mode did not fail. It succeeded and the result was discarded.

Together: the document received nothing on that path, so the home card is the user's **only** copy of that dictation — and #467 is what makes them land on it.

## To test by hand

Nothing below was verified at runtime. Both fixes need a real keyboard dictation with polish, which the simulator cannot produce headlessly. Phase 4 covered the build, the tests and the lint only.

**#467's overlay fix was validated on device at `f4d9a3d`; the three commits added since then un-validate this PR, so steps 3, 8 and 9 re-cover it.**

Arm **Translate → English** in the keyboard's mic fan (long-press the mic) and keep it armed throughout, so raw and polished are unmistakably different.

1. In Notes, dictate a sentence **in French** with the Dictus keyboard. Let it insert. → The document receives the English.
2. Switch to DictusApp. → You land on the **home dashboard**, not on the recording screen (centred card, "Touchez le texte pour copier", X top-left, mic below). **(#467)**
3. Read the "Dernière transcription" card. → The **English**, same as the document. Tap it → "Copié !", and pasting gives the English. **(#467)**
4. Now the refusal. In Notes, start a dictation in French and **leave the app while it is transcribing** (swipe home, or switch app). → Nothing is typed anywhere. **(#495)**
5. Open DictusApp. → Home dashboard, and the card holds the **English** — the text the refused insertion produced. Before this branch it held the French. **(#495, criterion 1)**
6. Tap the card. → Pasting yields the English. **(criterion 1)**
7. Check the log for that dictation. → `polishInsertionRefused` with its reason (`off-screen`, `no-identifier` or `different-document`), and `polishHandoff step=finished outcome=refused-with-text chars=<n>`. **`outcome=inserted` must not appear.** **(criteria 2 and 3)**
8. A refusal that produced nothing: dictate something very short or let a mode fail, with the keyboard away from the field. → The card holds the **raw**, and the log says `outcome=not-inserted chars=0`. **(criterion 4)**
9. A normal keyboard dictation that inserts, end to end. → Document gets the English, card gets the English, log says `outcome=inserted`. **(criterion 5)**
10. Dictate **from inside DictusApp** ("Nouvelle dictée"). → The recording screen stays up when it finishes and shows its result; the X dismisses it. This path must be unchanged. **(criterion 5)**
11. Force-quit and relaunch. → No transcription card at all.

### #467 — three commits (`4d1f089`, `ead5e22`, `f4d9a3d`)

1. **`PolishHandoffConclusion` (DictusCore).** Which status a concluded hand-off may take back. Only `.ready` — a `.failed` is a sentence the user still has to read, and the four live statuses belong to a dictation that has replaced this one.
2. **`polishHandoffFinished` returns the app to `.idle`.** `MainTabView` mounts `RecordingView` on `status != .idle`, and nothing ever wrote `.idle` on the path where the keyboard reports in — only `cancelDictation`, the overlay's own X, and the polish watchdog. The watchdog's own `.ready` check moves to the same rule, so the two endings of one hand-off leave the same thing behind.
3. **`RecordingView` tracks `coordinator.lastResult`** instead of snapshotting it once at `.ready`. The belt.

The cause, in three moves: `.ready` is published on the keyboard path while `lastResult` still deliberately holds the raw (so the `@State` copy is taken at the raw by construction); the polished text arrives later with no status change to carry it, so it reaches no observer; and the overlay was still mounted. The device log settles the third — `step=finished` at 14:01:50, then `from=ready to=clearing-for-new` at 14:01:52, with no `→ idle` between them.

### #495 — three commits (`a42644e`, `62b0471`, `727ac36`)

1. **`PolishedTextChannel` (DictusCore).** The pair — what the keyboard produced, and whether the document received it — behind one type, the way `PendingDictationChannel` and `DictationErrorChannel` own their keys. A type rather than a bare second key because the brief's own requirement is that the flag is *never left behind*: a stale `true` claims an insertion that did not happen, and two keys touched by four call sites in two targets is the shape where one site updates one key. `record` writes both, `clear` clears both, and `read` answers `.nothing` without a text whatever the flag says — so a flag can never speak on its own.

2. **`finish(text:insert:)` becomes `finish(reporting:inserting:)`.** The two decisions — what the app is told, what the document receives — were riding one flag, and that flag is what deleted the text. They are the same string on every path that types and differ only at the refusal. The refusal branch now passes `outcome.text`, which it already had in hand:

   ```swift
   PersistentLog.log(.polishInsertionRefused(reason: reason, ageMs: pending.ageMs))
   finish(reporting: outcome.text, inserting: nil)
   ```

   **No `DictationTail`** on that path: nothing was typed, so there is nothing to separate, and a trailing `". "` would follow the text into wherever the user pastes it.

3. **The four consumers.** `lastResult` and the history `updateText` take the produced text whether or not it was typed; the log outcome is three-valued; the Live Activity follows `lastResult`. Both clearing sites (`DictationHandoff` on hand-off, the launch sweep) go through the channel.

Only one of the four `finish(…nil, nil)` sites ever had text — the refusal. The other three (recovery with an armed mode and no generation, a generation that returned nothing, the stage watchdog) genuinely have none, and the raw stays the right outcome there. Verified by reading each.

### What deliberately did not change

- **`mayInsert` and the refusal itself.** Untouched, with its three reasons and its `polishInsertionRefused` line. Typing into a document the user did not choose is what this product refuses everywhere (#391).
- **The keyboard's toolbar message.** The refusal branch still returns before the `announce` block.
- **`outcome=inserted`.** Produced only by `.inserted`, which is written only by a `finish` that goes on to call `insertDictation`. A test asserts that neither other ending can carry that label, and `refused-with-text` deliberately contains neither `not-inserted` nor `inserted` as a substring, so the greps already used on these captures keep returning exactly the lines they used to.
- **The inserted path and the in-app path.** Byte for byte, apart from the `finish` signature.

## The history reversal, stated plainly

#495's brief reverses the rule that stood at `DictationHandoff.swift:346-349`: a refusal used to leave the **raw** in a subscriber's history record, as "the honest record". It now gets the **polished** text, like the card. The old reasoning was written when the raw was the only text the dictation had; it is not, and the history is the same kind of recovery surface as the card — the place a user goes for text their document never received. The comment has been rewritten to say what it now does and why.

**This path is unexercised in the shipping build and could not be tested.** `TranscriptionHistoryStore.append` returns nil without the Pro entitlement, so `polishHandoffHistoryID` is nil and `updateText` is never reached. The change is correct by reading and unverifiable by running, on device or otherwise.

## Known asymmetry, reported not fixed

On the inserted path `lastResult` receives the **tailed** text, so the home card carries a trailing separator; the new refused path writes the **untailed** text. The inconsistency predates #495, is invisible at the card's three-line cap, and changing the inserted path would alter behaviour validated on `f4d9a3d`. Named here as the issue asks; not fixed.

## Acceptance criteria

**#467**

| # | criterion | status |
|---|---|---|
| 1 | home card shows what the document received | **manual** — step 3 |
| 2 | cause named with its evidence | **met** — above |
| 3 | a refused insertion still leaves the raw | **superseded by #495** — it now leaves the *polished* text when one exists, and the raw when none does (step 8). The change is deliberate and is #495's whole point. |
| 4 | copying the card yields the polished text | **manual** — steps 3 and 6 |
| 5 | history path | see the reversal above |
| 6 | device, mixed-language | **manual** |

**#495**

| # | criterion | status |
|---|---|---|
| 1 | card shows the English, copying yields the English | **manual** — steps 5 and 6 |
| 2 | document still receives nothing, `polishInsertionRefused` still logged | **met by construction** (`mayInsert` untouched), visible at step 7 |
| 3 | three endings distinguished, `inserted` never lies | **met + tested** in DictusCore; visible at steps 7, 8, 9 |
| 4 | a refusal that produced nothing still leaves the raw | **met by construction** (`.nothing`), step 8 |
| 5 | inserted path and in-app path unchanged | **met by construction**, steps 9 and 10 |
| 6 | the PR says what happens to a subscriber's history, and that it could not be exercised | **met** — above |
| 7 | device, mixed-language | **manual** |

## Verification run

```
xcodebuild build -scheme DictusApp -destination 'platform=iOS Simulator,id=1720B34A…'
  ** BUILD SUCCEEDED **           (builds the embedded keyboard, widgets and DictusCore)

cd DictusCore && swift test
  Executed 1625 tests, with 1 test skipped and 0 failures (0 unexpected) in 14.921s

swiftlint lint --strict
  Done linting! Found 0 violations, 0 serious in 249 files.
```

Seventeen new tests: five on `PolishHandoffConclusion` (including a sweep over `DictationStatus.allCases`), twelve on `PolishedTextChannel` — the round trip for all three endings, the clearing rule, both halves of "a flag can never speak on its own", and the three log labels including the grep-disjointness.

Not unit-testable here, and not faked: the `RecordingView` and `MainTabView` behaviour, and the `finish` / `polishHandoffFinished` call-site wiring. `swift test` covers DictusCore only and there is no host-app test target, so assertions there would restate the code rather than exercise it.

`Localizable.xcstrings` was not regenerated and nothing else was left dirty.

refs #467, refs #495

## Summary by CodeRabbit

* **Bug Fixes**
  * Polished transcription results now remain visible even when insertion into the active document is declined.
  * Recording results update reliably when polished text arrives, preventing stale result cards.
  * Dictation handoff cleanup now keeps transcription text and insertion status synchronized.
  * Dictation status returns to idle more consistently after keyboard handoff completion.
  * Empty or incomplete handoff results no longer appear as successful insertions.
  * Dictation history now records polished text even when document insertion is refused.

### DIFF
--- a/DictusApp/DictationCoordinator.swift
+++ b/DictusApp/DictationCoordinator.swift
@@ -321,7 +321,10 @@ class DictationCoordinator: ObservableObject {
             defaults.removeObject(forKey: SharedKeys.lastTranscriptionTimestamp)
             defaults.removeObject(forKey: SharedKeys.lastTranscriptionPolicy)
             defaults.removeObject(forKey: SharedKeys.lastTranscriptionDuration)
-            defaults.removeObject(forKey: SharedKeys.lastPolishedTranscription)
+            // Through the channel since #495: the polished text and the flag saying
+            // whether it was typed are one answer, and a sweep that took only half of
+            // it would leave the flag to describe the next dictation.
+            PolishedTextChannel.clear(in: defaults)
             defaults.synchronize()
         }
         if let pending = PendingDictationChannel.current, pending.isExpired() {
--- a/DictusApp/DictationHandoff.swift
+++ b/DictusApp/DictationHandoff.swift
@@ -247,9 +247,12 @@ extension DictationCoordinator {
         defaults.set(token, forKey: SharedKeys.handoffToken)
         defaults.set(Date().timeIntervalSince1970, forKey: SharedKeys.lastTranscriptionTimestamp)
         defaults.set(DictationStatus.ready.rawValue, forKey: SharedKeys.dictationStatus)
-        // The previous dictation's polished text, if the keyboard left one behind.
-        // Cleared here so `polishDidFinish` cannot be answered with a stale string.
-        defaults.removeObject(forKey: SharedKeys.lastPolishedTranscription)
+        // The previous dictation's answer, if the keyboard left one behind. Cleared
+        // here so `polishDidFinish` cannot be answered with a stale string — and since
+        // #495 the "was it typed" flag goes with it, which is why this is one call and
+        // not two `removeObject`s: a flag outliving its text would claim an insertion
+        // for whatever landed next.
+        PolishedTextChannel.clear(in: defaults)
         defaults.synchronize()
 
         DarwinNotificationCenter.post(DarwinNotificationName.statusChanged)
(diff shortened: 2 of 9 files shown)


Pivii:
@coderabbitai review

Three commits were pushed after your first pass (`f4d9a3d` → `727ac36`) implementing #495 on the same branch: a new `PolishedTextChannel` in DictusCore, a signature change on `KeyboardPolishCoordinator.finish`, and the four consumers in `DictationHandoff.polishHandoffFinished`. Please review the branch as it now stands.

Pivii:
@coderabbitai full review

The incremental review requested at 12:50Z started an analysis chain at 12:51Z and has not produced a verdict in 33 minutes. Requesting a full review of `727ac36` instead.

Issue #531, opened 2026-09-08
Recover the last transcription from the expanded Dynamic Island, with a copy button


Pivii:
## Why

When an insertion goes wrong — the keyboard drops the text, the wrong app is focused, the field rejects it — the transcript is not lost, but it is out of reach. Today the Dynamic Island shows it for exactly **1 second** in the `.ready` phase, then auto-returns to standby (`DictusApp/LiveActivityManager.swift:657`). The user sees the result flash by and cannot get it back without opening the app.

History (`HistoryView`) covers the archive case and is Pro-gated. This is the opposite: a zero-navigation escape hatch for the dictation that just happened.

## The shape

In standby, the expanded Dynamic Island (long press) gains the last transcript plus a copy button, in the currently empty bottom region:

```
[logo] Dictus              [power] [mic]
       On
─────────────────────────────────────────
"Alors ouais, en fait, moi ce que…"  [Copy]
```

The bottom region stays empty until a first dictation has happened — that is today's screen unchanged.

## Decisions

1. **The long press cannot itself copy.** A Live Activity receives no gestures; the long press belongs to iOS and only expands the view. The interaction is long press → tap `Copy`. Two gestures, and there is no way to make it one.
2. **No expiry window.** The transcript stays available until the next dictation replaces it. A timed window creates two states the user cannot tell apart before expanding — sometimes the text is there, sometimes not, for no reason they can see.
3. **Nothing is removed.** `power` and `mic` keep their place in `expandedTrailing`; the transcript lands in `expandedBottom`, which is `EmptyView()` in standby today (`DictusWidgets/DictusLiveActivity.swift:254`).
4. **Never on the lock screen.** `lockScreenView` stays on `On`. The lock screen banner cannot be expanded at all, so the two surfaces are already separate functions in the same file — this costs nothing to enforce, and skipping it would leave the last dictation readable on a locked phone.
5. **Copy confirmation** is a state flip on the button (checkmark), not an animation the Live Activity cannot run.

## Probe first

Writing to `UIPasteboard` from a `LiveActivityIntent` while the app is backgrounded is **unverified**. Every other piece depends on it and there is no acceptable fallback — opening the app to copy defeats the purpose. Measure this on device before building anything else.

## Work

1. Probe the background `UIPasteboard` write from a `LiveActivityIntent`.
2. Add a last-transcript field to `DictusLiveActivityAttributes.ContentState` that survives the return to standby (today standby is rebuilt with an empty state).
[...]

Issue #42, opened 2026-03-17
Dynamic Island: intermittent desync and disappearance (Live Activity persistence)


Pivii:
## Context
Dictus uses a Live Activity / Dynamic Island to reflect the mic state (stand-by / recording / transcription finished).

## Problem
After chaining multiple recordings, the Dynamic Island becomes **unstable**:
- either it becomes **out of sync** with the actual mic state,
- or it **disappears** entirely while the system still seems to consider Dictus “active”.

This is **intermittent** (not 100% reproducible), but tends to happen after several cycles.

### A) State desynchronization
- I am **recording**, but Dynamic Island still shows **stand-by**.
- I **finished** a recording and even **started a new one**, but Dynamic Island remains stuck on a previous **“transcription finished / confirm”** state.

### B) Dynamic Island disappears
- I am no longer recording, but the Dynamic Island **is gone**.
- Only the **orange mic indicator** remains (like before Dynamic Island was implemented), suggesting Dictus is still “active”.
- Temporary “fix”: open Dictus again, then leave the app → Dynamic Island comes back in **stand-by**.

## Approx. steps to reproduce
1. Open Dictus
2. Repeat several cycles: start/stop recording, wait for transcription, start again, etc.
3. After some number of cycles, one of the symptoms above appears

## Expected behavior
- Dynamic Island should reliably reflect the **real state** (stand-by / recording / processing).
- It should not disappear unexpectedly: either the Live Activity is active and consistent, or it is properly ended.

## Notes / diagnosis ideas
- Live Activity **persistence** issue (not updated, not ended, or partially recreated).
- **State machine / race condition**: concurrent updates (start/stop/transcription callbacks) arriving out of order.

Things to inspect:
- where the Live Activity is created / ended
- where state updates are sent
- behavior when recordings are chained quickly
- fallback path where the “orange mic” remains while the Live Activity is lost

## Extra info to add (for better debugging)
- iOS version + iPhone model (Dynamic Island)
- does it happen more often when backgrounding / lockscreen / stand-by?
- logs (if possible) around Live Activity create/update/end

PR #539, opened 2026-09-10
Recover the last transcription from the expanded Dynamic Island (refs #531)

Pivii:
refs #531

## What this was for

When an insertion goes wrong — the keyboard drops the text, the wrong app is focused, the field
rejects it — the transcript is not lost, but it is out of reach. The Dynamic Island shows it for
exactly one second in `.ready` and then returns to standby with an empty state; History covers the
archive case and is Pro-gated behind a navigation. This is the opposite: a zero-navigation escape
hatch for the dictation that just happened. Long press the island, read the transcript, tap Copy,
get the **full** text back.

## To test by hand

The simulator cannot produce speech, so every criterion that starts with a real dictation is here.
Install the branch on the phone, and make sure Dictus has been in the background at least once so
the standby pill exists.

1. **Before any dictation, on a fresh install** — long press the Dynamic Island. Expect: Dictus /
   On, the power and mic buttons on the right, and **nothing** underneath them. (Covered on the
   simulator, listed here because it is one long press away.)
2. **Dictate normally, let the text land in a field**, background the app, long press the island.
   Expect: the first ~100 characters of the dictation on one or two lines, with a blue copy button
   at the right of the row. A dictation longer than that ends with `…`.
3. **Tap the copy button, then paste into Notes.** Expect: the **whole** dictation, not the
   truncated preview. This is the criterion the whole feature turns on.
4. **Watch the button as you tap it.** Expect: it swaps to a green checkmark for about two seconds
   and goes back to the copy glyph. *Not verified anywhere* — the pasteboard write is measured,
   the glyph swap was never caught on camera on the simulator. If it never flips, say so: the
   copy still works and the fix is a number.
5. **Lock the phone.** Expect: the banner reads `Dictus / On` with power and mic, and **no
   transcript**. (Verified on the simulator, but this is the privacy one — worth one look on the
   real device.)
6. **Dictate a second time**, then long press again. Expect: the new dictation, not the old one.
7. **Dictate something that produces nothing** (arm a Smart Mode that fails, or cancel during
   polish), then long press. Expect: the **previous** transcript is still there. A dictation that
   produced nothing has not replaced anything.
8. **Let the phone sit for several minutes with Dictus suspended**, then long press. Expect: the
   transcript is still there. The probe only backgrounded the app for seconds; a simulator does not
   reclaim processes the way a device does.
9. **Kill Dictus from the app switcher.** Expect: the island disappears entirely — no pill, no long
   press, no button. That is the feature's boundary, not a bug.

## What changed

| | |
|---|---|
| `DictusCore/Sources/DictusCore/LastTranscriptRecall.swift` | New. The durable store: full text in, preview derived. |
| `DictusCore/Sources/DictusCore/SharedKeys.swift` | `lastTranscriptRecall`, documented as *not* a hand-off. |
| `DictusCore/Sources/DictusCore/DictusLiveActivityAttributes.swift` | `lastTranscriptPreview` and `lastTranscriptCopied` on `ContentState`. |
| `DictusApp/LiveActivityManager.swift` | Records on `endWithResult`; one `standbyContent()` feeds every visible standby. |
| `DictusWidgets/CopyLastTranscriptIntent.swift` | New. The `LiveActivityIntent`, in **both** targets. |
| `DictusWidgets/DictusLiveActivity.swift` | The row in `expandedBottom` for `.standby`. |
| `Dictus.xcodeproj/project.pbxproj` | Dual target membership for the new intent. |
| `DictusCore/Tests/DictusCoreTests/LastTranscriptRecallTests.swift` | 11 tests on the store and the preview. |

Three decisions worth reading before the diff:

**The store is not one of the existing keys.** `lastTranscription` and `lastPolishedTranscription`
are hand-off channels — claimed by the keyboard, cleared by DictusApp on the next hand-off — so
their absence is an ordinary state and says nothing about whether a dictation happened. The new key
is written and claimed by nobody: replaced by the next dictation and by nothing else, which is
decision 2 in storage form. It is also independent of `TranscriptionHistoryStore`, which does not
record a non-subscriber's dictation at all.

**Both new `ContentState` fields are Optional, the `Bool` included.** ActivityKit persists this
state and decodes it with whatever binary renders the activity next, so an activity started before
an app update is decoded by code that has the fields. Swift's synthesised decoder calls
`decodeIfPresent` for an Optional and `decode` for a non-Optional — a plain `Bool` would throw on
the missing key. This is reasoned, not measured: `DictusLiveActivityAttributes` is `#if os(iOS)`
and needs ActivityKit, and the repo's suite runs on macOS, so there is no way to unit-test the
decode.

**The intent got its own file rather than joining `StopStandbyIntent.swift`.** The probe put it
there to inherit target membership for free; shipped code follows the repo's one-file-one-
responsibility rule instead, and the membership is added by hand in the project file. A missing
membership produces **no build error** — only a button that silently does nothing — so it is
verified by both targets' object files, not by a green build:

```
Dictus.build/Debug-iphonesimulator/DictusApp.build/Objects-normal/arm64/CopyLastTranscriptIntent.o
Dictus.build/Debug-iphonesimulator/DictusWidgets.build/Objects-normal/arm64/CopyLastTranscriptIntent.o
```

## One thing found while validating, and fixed

`startStandbyActivity` adopts an orphaned activity — one that outlived the app process — with
whatever content it was left holding. So the transcript row was present on a pill the app had
created and absent on a pill it had merely re-adopted: same phone, same dictation, two different
screens, for a reason the user cannot see. That is exactly the state decision 2 rules out.
Measured before the fix (fresh app process behind a surviving pill): power and mic, empty bottom
region, transcript sitting in the App Group the whole time. After: the row is there. Only
`.standby` is refreshed — any other adopted phase is a dictation in flight, and stamping standby
over it is #42.

## Two deliberate deviations from the issue

**The copy button carries an icon, not the word `Copy` the mock draws.** `DictusWidgets` has no
string catalogue — every string in that file is hardcoded English — so a `Copy` label would be the
one new string in the project no French user can read. The two buttons it sits under are icons for
the same reason. One line to change if you want the word.

**`.ready` now uses the same preview derivation as the standby row.** It was a bare
`String(prefix(100))`; it is now the shared function, which collapses newlines and appends `…` when
it cuts. Two truncation rules for the same string in the same view would drift.

## Acceptance criteria

| Criterion (issue "Device validation") | Status | Evidence |
|---|---|---|
| Bottom region empty before any dictation, power and mic unchanged | **Verified** | Simulator, store empty: expanded island shows Dictus / On / power / mic and nothing else |
| Transcript present in the expanded island | **Verified** | Simulator, store seeded with 192 characters: the row renders, cut at 100 with `…` |
| Copy yields the **full** text, not the preview | **Verified** | Clipboard seeded `SENTINEL_BEFORE_COPY_531`; after the tap, 192 characters ending in `…a cote de la plaque.` |
| Lock screen reads `On`, no transcript | **Verified** | Simulator locked with a transcript stored: banner shows Dictus / On / power / mic |
| Copy confirmation flips to a checkmark | **NOT verified** | The `perform()` that wrote the pasteboard also calls it, but the swap was never caught in a screenshot. Manual step 4. |
| A new dictation replaces the transcript | **NOT verified** | Needs a real dictation. Manual step 6. |
| A real dictation feeds the store at all | **NOT verified** | No simulator produces speech. Manual steps 2-3. |
| Long suspension | **NOT verified** | The probe's own note. Manual step 8. |

`cd DictusCore && swift test` — 1802 tests, 1 skipped, 0 failures.
`swiftlint lint --strict` — 0 violations in 265 files.
`xcodebuild build -scheme DictusApp` for the iPhone 17 Pro simulator — BUILD SUCCEEDED, all three
targets plus the widget extension.

https://claude.ai/code/session_01C78XD4CAdZZ58WMTHNYrAH

### DIFF
--- a/Dictus.xcodeproj/project.pbxproj
+++ b/Dictus.xcodeproj/project.pbxproj
@@ -115,12 +115,14 @@
 		AA00P205 /* PolishDebugView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA10P205 /* PolishDebugView.swift */; };
 		AA00P207 /* PolishDebugExporter.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA10P207 /* PolishDebugExporter.swift */; };
 		AA0000F6 /* StopStandbyIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100005 /* StopStandbyIntent.swift */; };
+		AA000531 /* CopyLastTranscriptIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100006 /* CopyLastTranscriptIntent.swift */; };
 		AA0000G0 /* KeyTapSignposter.swift in Sources */ = {isa = PBXBuildFile; fileRef = AA1000G0 /* KeyTapSignposter.swift */; };
 		AA000P10 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = AA100P10 /* PrivacyInfo.xcprivacy */; };
 		AA000P20 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = AA100P20 /* PrivacyInfo.xcprivacy */; };
 		BB000001 /* DictusWidgetBundle.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100001 /* DictusWidgetBundle.swift */; };
 		BB000002 /* DictusLiveActivity.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100002 /* DictusLiveActivity.swift */; };
 		BB000003 /* StopStandbyIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100005 /* StopStandbyIntent.swift */; };
+		BB000004 /* CopyLastTranscriptIntent.swift in Sources */ = {isa = PBXBuildFile; fileRef = BB100006 /* CopyLastTranscriptIntent.swift */; };
 		BB000010 /* DictusCore in Frameworks */ = {isa = PBXBuildFile; productRef = BB900010 /* DictusCore */; };
 		BB000020 /* DictusWidgets.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = BB200001 /* DictusWidgets.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		EE000001 /* dictus_trie.cpp in Sources */ = {isa = PBXBuildFile; fileRef = EE100001 /* dictus_trie.cpp */; };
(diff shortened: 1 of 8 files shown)


Pivii:
Closing: the feature rests on a write iOS does not allow, measured on device.

`UIPasteboard.general.string` is **silently refused when the writing process is backgrounded**. No error, no exception, no effect. Instrumented probe on iPhone 15 Pro Max / iOS 26.6.1, build `d47b6f4`:

```
component=copyIntent action=pasteboardWrite
host=app  bundle=com.pivi.dictus  wrote=36  readBack=-1  hasStrings=false
```

`host=app` rules out the one hypothesis that had a fix — `perform()` does run in the host app process, not the widget extension. The text was present (36 chars), the assignment executed, the read-back is nil.

**The simulator does not enforce this**, which is why the probe on branch `probe/531-live-activity-pasteboard` passed and this one does not. That is the durable lesson: a `UIPasteboard` write measured on a simulator says nothing about a device.

The branch and both probe branches are deleted. #531 is closed `not_planned` with the same measurement.

Issue #536, opened 2026-09-10
The polish glossary has no measurable effect: Apple FM ignores our term list


Pivii:
## What we measured

The polish glossary appends one sentence to every prompt (`PolishGlossary.swift`):

```
Spell these terms exactly as written: Dictus, WhisperKit, Parakeet v3,
FluidAudio, GitHub, TestFlight, iOS, App Store, Argmax, Apple Intelligence.
```

Since #80, the user's own canonical terms join that list (`activePromptBlock`). Two device measurements say the model ignores it.

| Date | Engine output (`raw`) | Glossary holds | Polished output |
| --- | --- | --- | --- |
| 2026-09-08 | `dictus` | `Dictus` | **`dictés`** — actively corrupted |
| 2026-09-10 | `Parakit V3` | `Parakeet v3` | **`Parakit V3`** — left alone |

Both on iPhone 15 Pro Max, iOS 26.6.1, `parakeet-tdt-0.6b-v3`, French, Apple FM, `natural` mode.

The control in the same session: `whisperflow` is **not** in the glossary, and the model corrected it to `WhisperFlow` on its own. So Apple FM fixes what its own priors already know and does not act on our list.

## The likely cause: we ask the wrong question

"Spell these terms exactly as written" only helps when the term is **already correct** in the transcript. Our actual need is the opposite — the transcript holds `Parakit V3` and we want `Parakeet v3`. That is a **repair**, not a preservation.

`n0an/VivaDicta`, already the reference for #80's matching rules, ships the same two layers (a deterministic `ReplacementsService` plus an LLM glossary) and frames its glossary as repair:

> When **similar phonetic occurrences** are detected between words in the `<TRANSCRIPT>` text and terms in `<CUSTOM_VOCABULARY>` or `<CLIPBOARD_CONTEXT>` — prioritize the spelling from these context sources over the `<TRANSCRIPT>` text.

Three further differences worth testing, in decreasing order of expected effect:

1. **Framing** — phonetic repair vs. spelling preservation. The one above.
2. **Structure** — the vocabulary arrives in a tagged `<CUSTOM_VOCABULARY>` block rather than in a sentence buried among other instructions.
3. **Length** — they maintain a second, ~120-token prompt for small on-device models, with the comment that small models follow a short prompt more reliably than a long one.

[...]

Issue #80, opened 2026-04-08
Custom vocabulary — a deterministic replacement pass over the transcript


Pivii:
> **Rewritten 2026-09-07 after a grilling session.** Thirteen decisions are settled below and should not be re-litigated. The original body proposed biasing WhisperKit through an `initialPrompt` parameter; that parameter does not exist in the version we ship, and the engine most users run is not WhisperKit. What the feature *is* changed as a result.

## The promise, and what exists

`ProFeature.swift:39` / `:48` — **"Teach Dictus your technical terms"** / *"Apprenez à Dictus vos termes techniques"*. It is the third Pro feature; `smartMode` and `history` are built, this one is not.

`SharedKeys.vocabularyEnabled` is seeded to `true` at first launch by `ProStatusManager.seedFeatureTogglesIfNeeded` and **read by nobody**. Two occurrences in the whole repo: its definition and its mapping in `ProFeature`. It is a switch wired to nothing.

Nothing here is reachable by a user until #279 flips `PremiumFlags.paywallVisible`.

## Why this is not engine biasing

Two facts, both measured, that the original body did not have.

**WhisperKit has no `initialPrompt`.** In 0.16.0 the public `DecodingOptions` exposes `promptTokens: [Int]?` and `prefixTokens: [Int]?` — token IDs you must encode yourself, capped at `(448/2)-1 = 111` tokens by `TextDecoder.swift:338-341`, silently truncated with `suffix()` beyond that. Passing them also disables the prefill cache: `TextDecoder.swift:352-355` requires `options?.promptTokens == nil` for `usePrefillCache`, and we pass `usePrefillCache: true` at both production sites.

**Whisper is not the default engine.** `ModelInfo.swift:588` recommends `parakeet-tdt-0.6b-v3` on any iPhone with ≥6 GB. Most users never go through WhisperKit at all.

**And Parakeet's purpose-built mechanism is English-only.** FluidAudio 0.12.4 ships `ASR/CustomVocabulary/` with `configureVocabularyBoosting(vocabulary:ctcModels:config:)`, which rescores the TDT's words against a **second CTC encoder**. Measured from the HuggingFace API:

| Repo | declared languages |
| --- | --- |
| `FluidInference/parakeet-tdt-0.6b-v3-coreml` — what we ship | 25, incl. `fr` |
| `FluidInference/parakeet-ctc-110m-coreml` — the judge | **`["en"]`** |
[...]

PR #540, opened 2026-09-10
Cut the polish glossary and require a variant (refs #536)

Pivii:
## What this was for

The vocabulary feature promised two things and only one of them worked.

A vocabulary entry is a correct spelling plus the shapes the engine writes instead. Each variant triggers a deterministic replacement on the transcript — that half is validated on device and untouched here. The other half, #80 decision 7, put the canonical term into the polish prompt, which is what let a term with **no** variant claim to protect its own spelling, and what made the paywall sentence "Teach Dictus your technical terms" true for someone who knows the spelling they want but not yet the mangling.

That half was measured on 2026-09-10 and it does nothing. `Dictus` sat in the prompt while Apple FM wrote `dictés`; `Parakeet v3` sat in it while `Parakit V3` came back untouched. In the same session `whisperflow`, in nobody's list, was corrected from the model's own priors. VivaDicta ships the same two layers, asks for phonetic repair from a tagged `<CUSTOM_VOCABULARY>` block — every axis a rewrite would have changed — and corrected `Claude Code` in 0 of 5 dictations on the same iPhone, with its own log showing Apple FM ran every time.

So this PR takes outcome B in full: the term list comes out of all eleven prompts rather than being reworded, the paywall sentence is **kept**, and the state that made it false stops being reachable — the add sheet now requires at least one variant. No stored data is touched: an entry saved before today with no variants still loads, still lists, and says on its own row what it needs.

## To test by hand

Everything below needs a human eye or a physical iPhone. The simulator checks that are already done are listed under Verification.

1. **A real dictation still polishes.** On device, Normal polish on, dictate two or three ordinary French sentences. Expect: polished output as before, punctuation and capitalisation intact. The prompts each lost two lines; nothing else moved. This is the only check that the shorter prompts did not change polish behaviour, and no simulator can run Apple FM.
2. **A Smart Mode still runs.** Arm `List`, dictate a rambling paragraph, validate. Expect: bullets, same language. Then arm a Translate mode and dictate again. Expect: a translation. Both prompts lost the same two lines.
3. **The vocabulary replacement still corrects.** Add a term with a variant, dictate the variant. Expect: the canonical spelling in the document. This is the half the feature now rests on entirely.
4. **The paywall cards, both languages.** French then English. Expect: Vocabulaire / Vocabulary reading "Apprenez à Dictus vos termes techniques" / "Teach Dictus your technical terms", unchanged. Captured on the simulator already; worth one look on the real screen since it is the sentence this issue was about.
5. **The orange row, on a real screen.** Settings → Mes termes with an old variant-less entry present. Expect: an orange second line, "Sans effet pour l'instant. Ajoutez ce que Dictus écrit à la place." Judge whether orange is right there, or whether it should be plain secondary grey. That is a taste call and it is yours.

## What changed

**Three commits.**

1. **The glossary is cut.** `PolishGlossary` deleted with `terms`, `promptBlock`, `activePromptBlock` and `block(for:)`. The eleven prompt builders lose the `glossary:` parameter and the two-line block they interpolated it into: `PolishAutoPrompt`, `PolishNaturalPrompt{FR,EN,ES,DE}`, `PolishRepairPrompt{FR,EN,ES,DE}`, `SmartModeNotesPrompt`, `SmartModeTranslatePrompt`. `AppleFoundationModelsPolishEngine` and `SmartModeCatalogue` lose their call sites, and the catalogue's "known limit" doc comment — a term added mid-process reaching Smart Modes only at the next launch — goes with the mechanism it described, which also settles the staleness CodeRabbit raised on #525. `CustomVocabulary.glossaryTerms()` goes, and `activeEntries()` with it since that was its only caller.
2. **The second field becomes required.** `VocabularyEntry.hasEffect` carries the rule; `VocabularyEditorView.canSave` is one call to it. It is deliberately **not** folded into `isValid`, which `VocabularyStore.read` filters on at every load — merging them would delete every variant-less entry written before today, silently, on the next launch. The list marks such an entry in orange with the action that fixes it. Copy updated in `Localizable.xcstrings` with French.
3. **Docs.** An update section on ADR 0002 retiring the decision and the three-mechanism consequence it created; `CONTEXT.md` keeps `PolishGlossary` as a definition a reader will meet in old issues and says what replaced it; the language-onboarding guide stops telling the next author to add a slot that no longer exists; the #80 corpus note stops crediting the glossary.

`VocabularyReplacer` and `CustomVocabulary.corrected` are untouched.

## Acceptance criteria

| --- | --- | --- |
| `PolishGlossary` and `glossaryTerms()` gone, no prompt carries a term list | Done | `grep -rn --include='*.swift' "PolishGlossary\|glossaryTerms" .` returns nothing. New `PolishPromptInventoryTests` walks all fourteen prompt strings and asserts none contains "Spell these terms exactly as written", "Domain vocabulary" or "CUSTOM_VOCABULARY". |
| `swift test` green | Done | 1791 tests, 1 skipped, 0 failures. |
| `swiftlint --strict` at 0 | Done | 0 violations in 262 files. |
| All three targets build | Done | `DictusApp`, `DictusKeyboard`, `DictusCore` all `** BUILD SUCCEEDED **` on iOS Simulator, after `./scripts/patch-fluidaudio-swift5.sh build/DerivedData`. |
| The sheet refuses an entry with no variant | Done | Simulator, both languages: Save greyed out with "Add at least one. A term on its own changes nothing Dictus writes." / "Ajoutez-en au moins une. Un terme seul ne change rien à ce que Dictus écrit." Save turns blue as soon as a variant is typed. |
| An existing variant-less entry still opens and still lists | Done | A `Systeko` entry with `"variants":[]` seeded into the App Group file loads, lists with the orange notice, and opens in the editor keeping its term. Unit-covered too: `testAnEntryStoredWithoutVariantsStillLoadsAndStillLists`, `testAnEntryWithNoVariantsIsStillValid`. |
| The paywall card unchanged in both languages | Done | `ProFeature.swift` and `PaywallView.swift` are not in the diff. Rendered and read on the simulator: "Apprenez à Dictus vos termes techniques" and "Teach Dictus your technical terms". |
| #80 decisions 6 and 7 amended | Done | Comment posted on #80. |

The simulator checks needed a temporary local flip of `PremiumFlags.paywallVisible` to reach a screen #279 still hides. It was reverted before pushing and is not in the diff; the three target builds above were re-run on the reverted source.

refs #536

https://claude.ai/code/session_01TeYZTMSBk6b2Gcc8XFubpx

## Summary by CodeRabbit

* **Changes**
  * Removed the built-in glossary term list from polishing prompts.
  * Custom vocabulary continues to work through replacement rules.
  * Smart, natural, repair, and automatic polishing no longer use glossary terms.

* **Bug Fixes**
  * Vocabulary entries now require at least one replacement variant to be saved.
  * Entries without effective variants are clearly marked and cannot be saved as new entries.
  * Existing variant-less entries remain available and load without changes.

* **Documentation**
  * Updated vocabulary and polishing guidance to reflect the current behavior.


Pivii:
## Device-validated on 2026-09-10, `dc04542`

iPhone 15 Pro Max, iOS 26.6.1, `parakeet-tdt-0.6b-v3`, French, Apple FM, polish enabled. Log header reads `rev dc04542@HEAD`.

**The build carried one uncommitted change**: `PremiumFlags.paywallVisible` flipped to `true`. The vocabulary screen sits inside the Pro section, which that flag hides, so no part of the entry-point half is reachable otherwise. Nothing else differed from this branch.

### Polish still runs, and the shortened prompts did not degrade it

Three dictations, `outcome: success` on all three, no guardrail rejection, no engine failure. Latencies 933 / 1468 / 3503 ms. Each prompt lost two lines and nothing downstream noticed.

### The replacement pass works in both processes

| Writer | Engine wrote | After the pass |
| --- | --- | --- |
| `<APP>` | `Tictus`, `Cloud Code` | `Dictus`, `Claude Code` |
| `<KBD>` | `cloud code` | `Claude Code` |

`vocabularyApplied enabled=yes entries=2 replacements=2 / 0 / 1` across the three.

**And polish preserved those corrections.** The pass runs upstream of Apple FM, so the open question was whether the model would undo it once the glossary stopped asking it not to. It did not: `Dictus` and `Claude Code` survive into `polished` in both captures. That is the one regression this PR could plausibly have caused, and it did not happen.

### Smart Modes

Reported working by the maintainer on device, checked by hand after the log above was exported. Not covered by a capture — no `List` or Translate event appears in the exported log, so this line rests on direct observation rather than on evidence in this thread.

### One consequence worth recording

In the keyboard capture, lowercase `dictus` stayed lowercase. That is not a regression — it is precisely the gap the glossary claimed to close and was measured in #536 not to close. The remedy is now explicit and deterministic: add `dictus` as a variant. Matching is case-insensitive, so one variant covers every capitalisation.

The only ERROR in the log is `subscriptionError loadProducts empty result`, which is #215 and appears only because the paywall flag was flipped for this build.

Issue #542, opened 2026-09-11
A cold-start dictation is silently lost while the model compiles, and the readiness gate cannot see it coming


Pivii:
## What happens

On 2026-09-11, on a device whose Core ML cache had just been discarded by a dev install, a cold-start dictation from the keyboard with **Turbo** selected produced **no text at all**, and said nothing about why.

```
09:32:19  modelLoadStateChanged from=ready to=loading   reason=init-preload   ← Turbo starts compiling
09:32:20  dictation starts, user is returned to Notes, recording runs
09:32:26  statusChanged recording → transcribing
09:32:56  statusChanged transcribing → idle             ← 30 s later, silently. No text.
09:35:59  modelLoadStateChanged loading → ready         ← 3 min 40 after the load began
09:35:59  transcriptionStarted  model=…turbo_632MB
09:35:59  transcriptionFailed   error=Swift.CancellationError
```

The user spoke, the waveform ran, the Dynamic Island showed "transcribing", the status went back to idle as if it had finished — and nothing was inserted. `Dictus → History` had no entry either. The transcription finally ran almost four minutes later against a model that had just finished compiling, and was cancelled.

The `CancellationError` half is the known #144 (slow Turbo compile + concurrent dispatch). What this issue is about is **why nothing stopped the dictation from starting in the first place**, and why nothing told the user.

## Root cause: the keyboard's readiness gate cannot see a cold load coming

The keyboard already has a gate. It works, and the same log shows it firing correctly twice:

```
09:27:16  <KBD> dictationDeferred reason=keyboard opened Dictus for model preparation
09:27:17  <APP> urlComponents query=source=keyboard,intent=prepare
09:27:17  <APP> dictationDeferred reason=keyboard prepare-only URL
```

The keyboard reads the model load state from the App Group and, when it reads `loading`, sends `dictus://dictate?intent=prepare` instead of starting a dictation — which opens the preparation screen and starts nothing.

**But at 09:32:19 the App Group said `ready`.** It said `ready` because the model *file* is downloaded. The compile had not started yet: it starts when the app launches, which is the same instant the keyboard is handing off. The keyboard asked the right question and got an answer that was true and useless.

So the gate is reliable exactly when the app is already running and already loading, and blind in the case that costs the most — a cold app with a cold Core ML cache.

**`ready` in the App Group means "the model file is on disk", not "the model can transcribe now".** Those two diverge for minutes after any install, and a dev install discards the Core ML cache every time (see the existing note on that).

[...]

Issue #484, opened 2026-09-03
Tapping New recording while a model is loading does nothing at all, where the keyboard shows the preparation screen


Pivii:
## Source

Device validation of PR #477 (issue #458), 2026-09-03. `1.8.1 (29)`, `rev dd7e837@HEAD`, iPhone16,2, iOS 26.6.1, Turbo selected and compiling for the first time.

Found by the maintainer while trying to reproduce #458. It is not a #477 regression: the guard below is on `develop` and #477 does not touch `DictationCoordinator`.

## Bug

Tap **New recording** on the Home screen while a model load is in flight, and **nothing happens at all**. No overlay, no message, no disabled state, no sound. The button looks alive and answers nothing.

Twelve taps in eleven seconds, every one refused in silence:

```
12:44:11  statusChanged from=idle to=clearing-for-new source=startDictation-reset
12:44:11  dictationStarted fromURL=false appState=0 engineRunning=false
12:44:11  dictationDeferred reason=model load in flight (state=loading)
…                                                    (×12, 12:44:11 → 12:44:22)
```

`statusChanged … to=clearing-for-new` is a log label, not a state: the published `DictationStatus` never leaves `.idle`, so no view has anything to react to.

## Why

`DictationCoordinator.swift:599` refuses the start and returns:

```swift
if currentLoadState == ModelLoadState.loading.rawValue && !fromURL {
    PersistentLog.log(.dictationDeferred(reason: "model load in flight (state=loading)"))
    // Keep the keyboard request pending. KeyboardState re-reads the
    // shared state before its fallback URL and opens the prepare-only
    // flow instead of showing a refusal message.
    return
}
```

The comment says what is missing. The `!fromURL` condition and the note both describe **the keyboard's** answer to this situation — #262 routes a cold-start keyboard tap to the model preparation screen instead of refusing it, and that path works: tapping the keyboard mic while Turbo compiles opens Dictus on the preparation screen with its progress. The in-app button is the same refusal with no such route, and it is the one path where the user is already inside the app and could be shown the truth for free.

## What the maintainer asks for

Match the keyboard. A tap on **New recording** while the model is loading should take the user to the model preparation screen — the same screen `dictate?intent=prepare` raises — so the wait is visible and named, rather than the button silently swallowing the tap.

Two things this must not undo:

[...]

Issue #432, opened 2026-08-27
The preparation screen never says that a first Turbo compile takes about three and a half minutes


Pivii:
## What happens

Selecting Turbo, or launching the app with Turbo active after an install, puts the user on the preparation screen for around three and a half minutes. The screen says the model is preparing. It does not say how long that takes, and nothing distinguishes Turbo from a model that is ready in thirty seconds.

The maintainer sat through it on 2026-08-27 and asked for the wait to be named. His words: he could not tell whether it was normal or whether something was stuck. That is the correct reaction to the current screen, because nothing on it answers that question.

## The measurements

Four cold readings of a full Turbo compile on the same iPhone 15 Pro Max, iOS 26.6, across three days:

| date | reading | source |
| --- | --- | --- |
| 2026-08-25 | 236 s | benchmark for #408 |
| 2026-08-26 | 212 s and still running when interrupted | device log |
| 2026-08-26 | 202 s | device log, `init-preload` to `init-preload-success` |
| 2026-08-27 | 210 s | device log, same pair |

So: about three and a half minutes for a first Turbo compile, and a couple of seconds for every load after it, once the Core ML bundle cache exists.

Medium's cold compile is 32 s. That is the gap the copy has to carry.

Two things a user cannot know from the screen today, and both matter more than the number:

- The long wait is **once per model per install**. Every later load of the same model is seconds.
- **Installing a build discards the cache**, so a tester who installs a new build pays it again. That is a development concern rather than a shipping one, but it is why the wait keeps reappearing during testing.

## What this asks for

Copy on the preparation screen that names the expected wait when the model being prepared is one of the slow ones, and says the wait is once rather than every time. Not a progress bar — Core ML gives no progress for a compile, which is exactly why the screen has none today.

Deliberately not specified here: the exact wording, whether the number is per-model from `ModelInfo` or a coarse "a few minutes", and whether anything changes for the fast models. Those are for whoever picks this up.

## Why it is worth doing

#428 cut an escape button from the preparation screen after four review passes found fourteen defects in it, all of them caused by letting the user act mid-compile. This issue is the same problem approached from the honest end: rather than letting the user leave a wait they do not understand, tell them what they are waiting for. It carries none of that risk, because it changes text and nothing else.

Related: #428 for the screen itself and why it has no exit; #408 for the Turbo variant these timings belong to.

PR #544, opened 2026-09-11
Refuse a cold-start dictation the model cannot transcribe yet (refs #542)

Pivii:
## What this was for

A cold-start dictation from the keyboard was being lost in silence. On 11 September, with Turbo selected on a device whose Core ML cache had just been discarded, the keyboard read `ready` from the App Group and started a dictation. The user spoke, the waveform ran, the Dynamic Island said "transcribing", the status went back to `idle` thirty seconds later, and nothing was inserted. `History` had no entry either. The transcription finally ran 3 min 40 after the fact, against a model that had only just finished compiling, and was cancelled.

The gate was not broken. It was blind. `ready` in the App Group means "the model file is on disk", not "the model can transcribe now", and those two diverge for minutes after any install. The compile starts when the app launches, which is the same instant the keyboard is handing off — so the keyboard asked the right question and got an answer that was true and useless. Capture `b884a489` holds both sides of that race in one file: the same gate was wrong at 12:13 and got away with it because the load it could not see took one second, and right at 12:38 only because somebody had started that load 49 seconds earlier.

#23 did not cause this. It removed the only thing that made it legible: before auto-return, a dictation started during a long load left the user inside DictusApp, looking at the preparation screen that explains itself.

Two things had to change. The gate needed a question it could actually answer, and a dictation parked behind somebody else's Core ML compile had to stop being cancelled by a watchdog meant for stages that will never hand over.

## To test by hand

Four scenarios, and three of them cost a full uninstall. Nothing below could be verified by an agent: they all need a real install cycle on a physical device.

1. **Uninstall Dictus. Install this branch. Open the app once, complete or resume onboarding with Parakeet selected, and let the model finish preparing. Leave the app.** Expected: the preparation screen shows a sentence beginning "After each update…" and never "It happens only once".
2. **Uninstall again, install again, and this time do not open DictusApp beyond what the install requires. Go to Notes, bring up the Dictus keyboard, tap the microphone.** Expected: Dictus opens on the preparation screen, **no recording starts**, the waveform never runs, and you are not sent back to Notes. Nothing is captured, so nothing can be lost. The screen tells you to return to your app and tap again.
3. **Wait for that preparation to finish, return to Notes, tap the microphone again.** Expected: a normal dictation. Recording starts, and when you stop, you are returned to Notes automatically and the text is inserted. This is #23 behaviour, unchanged.
4. **Repeat 2 and 3 with Turbo 632MB selected instead of Parakeet.** Expected: identical, with the wait named in minutes on the preparation screen. This is the configuration that lost the words in the original capture.
5. **Without reinstalling anything: force-quit DictusApp from the app switcher, then dictate from the keyboard in Notes.** Expected: the dictation **starts normally**, waits a few seconds, and the text arrives. This is the ordinary cold start and it must not have moved. If this one shows a preparation screen, the warmth record is not being written and the change is wrong.
6. **Same as 5, but check the log afterwards.** Expected: a `WarmInference … action=completed` line carrying `warmthRecorded=true`, and no `watchdogReset source=appTranscription`. Scenario 5 is the path that parks on the init lock, which is what the first commit is about.
7. **Settings → Models, re-download a model you already have, then dictate from the keyboard before opening the app again.** Expected: one preparation screen, then normal dictation. A fresh download starts from a fresh cache, so its warmth record is cleared.

Two lines in the debug log tell the two refusals apart, and it is worth grepping for them: `dictationDeferred` now carries `(loading)` when the app is working on the model right now, and `(cold-cache)` when nothing is working on it yet.

## What changed

**Commit 1 — the watchdog.** `isWaitingForNeuralEngine`, the flag that stops the 30 s transcription watchdog from cancelling a dictation parked behind somebody else's compile, was raised in one place only: `acquireNeuralEngine`'s queue wait. A dictation started while the app process was dead never reaches that wait — it parks earlier, in `awaitInFlightEngineInit`, on the launch preload's `initTask`. So the deferral covered the rarer of the two paths, and the common one survived on margin: two measured survivals in capture 127, 7 s parked and 3 s parked, against a 30 s budget, both on a warm cache. The deferral is now raised in both waits, under the same strict rule — only while the work belongs to another caller, never around a caller's own compile, because that trades a recoverable failure for an unbounded hang.

**Commit 2 — the gate.** New `ModelWarmth` in DictusCore records which models have completed a warm inference in *this* installation, keyed by (bundle container UUID, iOS version, model identifier). The container gets a fresh UUID at every install and the keyboard extension lives inside it, so both processes read the same identity with no hand-off; the iOS version covers an OS update, which discards compiled artefacts without touching the container. App Group storage survives a reinstall, which is precisely why the old signals lied. The record is written at the existing `WarmInference action=completed` point, which is the first moment anything has proof the model produces a result rather than merely being resident.

`RecordTapRouting.decide` gains that one input and the keyboard calls it instead of keeping its own copy. The order of its questions is unchanged and a load already in flight still presents exactly as it did under #484.

## The ten decisions

| # | Decision | Status | Evidence |
|---|---|---|---|
| 1 | Two commits, watchdog first | Done | `e60c717` then `b791a45` |
| 2 | Cold model routes to preparation, never records | Done | `RecordTapRouting` returns `.presentPreparation`; the keyboard sends `intent=prepare`, which returns from `handleIncomingURL` before any auto-return code. Device scenarios 2 and 4 above |
| 3 | Watchdog deferral on the init-lock wait too | Done | `awaitInFlightEngineInit` raises `isWaitingForNeuralEngine` inside its wait loop, lowered on all three exits including the early `return true`. Not unit-testable: both flags are DictationCoordinator state, in DictusApp, which `swift test` does not cover. Device scenario 5 |
| 4 | Keyed by (container, iOS version, model) | Done | `ModelWarmth.installIdentity`, 6 unit tests including the app/extension agreement and the simulator's two-UUID path |
| 5 | No threshold, no duration table, never `firstPreparationSeconds` | Done | The `.idle`/`.ready` arm reads only the record. `firstPreparationSeconds` appears nowhere in the gate path |
| 6 | Dead process with a warm cache goes through untouched | Done | `testAWarmModelIsUntouchedOnEveryLoadState` over every status and both non-loading states. Device scenario 5 is the one that proves it |
| 7 | One rule, in `RecordTapRouting` | Done | The keyboard's `isModelLoading()` is gone; its three call sites go through `preparationDeferral()`, which delegates |
| 8 | "It happens only once" replaced | Done | Three strings, English and French, now say "After each update…". The duration bucket is untouched: decision 8 forbids a first-install/update *variant*, not #432's per-model figure |
| 9 | Warmth cleared when files are re-downloaded | Done | `ModelManager.downloadModel` clears at the start of the transfer, one site for both engines, so an interrupted download cannot leave a claim standing over half-written files |
| 10 | `CancellationError` and post-preparation auto-return out of scope | Honoured | Neither touched |

## One thing added that the spec does not name

Decision 2 says a cold model routes to the preparation screen, and the screen "already exists and already says the right things". That is true for the keyboard path, where the app is launching and the launch preload is what the screen watches. It is not true in one state: the app already alive with `modelLoadState` at `.idle`, which in a live process means the last load failed or was abandoned. There, a refusal would raise a screen with nothing preparing behind it; the screen would reach its ready state, dismiss, and send the user back to a tap that refuses again.

So the in-app refusal kicks `preloadActiveModel()` in that exact state and no other. `.loading` is #484 untouched, and `.ready` cannot be reached cold in a live process because the load that writes `.ready` is the same one that records the warmth.

## What I am not comfortable with

**A load that fails outright now shows a preparation screen instead of an error.** If the kicked load throws, `modelLoadState` returns to `.idle`, and `ModelLoadingOverlay` maps a downloaded model plus `.idle` to its ready state — so the user sees a checkmark over a model that did not load, and the next tap starts the cycle again. That mapping predates this change (`loadGaveUp` only covers the launch deadline, `init-preload-deadline`), and widening it means deciding what the screen says about a genuine load failure, which is a copy question this issue did not answer. It loses no words, which is this issue's bar, but it is a real edge and it is worth a follow-up.

**Decision 8's "one string" is read as one wording, not one string literal.** The three per-model duration variants from #432 are kept, and each lost "It happens only once". Reading it as a single literal would have thrown away the measured durations, which #432 exists for. If that reading is wrong, it is three lines to change.

## Verification

| Check | Result |
|---|---|
| `xcodebuild build` — DictusApp, DictusKeyboard, DictusCore, iOS Simulator | exit 0, 0 errors each |
| `cd DictusCore && swift test` | 1839 tests, 1 skipped, 0 failures |
| `swiftlint lint --strict` | 0 violations in 265 files |
| `Localizable.xcstrings` | 9 lines changed, all three deliberate strings; no build noise committed |

Everything else is on the manual list above. No agent can install this on a device, and all four of the spec's validation scenarios need one.

https://claude.ai/code/session_01PZuFwSqDRX29HZSqyrj5i9

## Summary by CodeRabbit

* **Bug Fixes**
  * Recording now routes to model preparation when the model’s processing cache is cold, including from the keyboard.
  * Model preparation now starts automatically when needed instead of showing an inactive preparation screen.
  * Preparation status remains accurate while the model is loading or initializing.

* **User Experience**
  * Updated preparation messaging explains that the initial wait can recur after app or operating-system updates, while later waits typically take only a few seconds.


Pivii:
## Device validation — the cold-cache gate fires

Validated on iPhone 15 Pro Max / iOS 26.6.1, `rev b791a45@fix/542-cold-cache-gate`, Parakeet selected, on a container installed and **never launched** before the tap.

```
14:57:42  keyboardMicTapped
14:57:42  dictationDeferred reason=keyboard opened Dictus for model preparation (cold-cache)
14:57:42  appLaunched                       <- the app was dead until the tap
14:57:42  launchURLCaptured intent=prepare
14:57:42  dictationDeferred reason=keyboard prepare-only URL
14:57:42  modelLoadStateChanged from=ready to=loading reason=init-preload
```

No `dictationStarted`, no `statusChanged … recording`, no waveform. This is the exact state that lost the user's words at 09:32:19 in capture 127, and it now refuses instead of recording into a compile.

Also observed across the session's captures:

- `(loading)` still fires on its own path, 6 times, so the pre-existing rule is intact (decision 7's ordering).
- `WarmInference … warmthRecorded=true`, 6 times, on both engines (decision 4).
- A warm-cache cold start let through and completed: transcription in 124 ms, text inserted (decision 6).
- Zero `watchdogReset` in every capture.
- The preparation screen's new wording confirmed by eye (decision 8).

### What is NOT proven on device, deliberately

- **Turbo 632MB.** Same code path as Parakeet, only slower to stage. Skipped by the maintainer's decision.
- **Decision 9** (a re-download clears the record). Near-impossible to stage by hand: re-downloading happens inside the app, which then preloads and re-warms the model, closing the window before a keyboard tap can read it. Covered instead by `testClearingOneModelLeavesTheOthersWarm` and by the single call site at the first line of `ModelManager.downloadModel`, before any transfer.
- **Decision 3** (the watchdog deferral on the init-lock wait). Its flags are `DictationCoordinator` state in DictusApp, which `swift test` does not reach. No watchdog fired in any capture, which is consistent with the fix but is not a proof of it.

### A note for a later reader

`ModelWarmth.isWarm` returns `true` when the install identity cannot be read (`testAnUnknownIdentityReadsAsWarmRatherThanLockingTheUserOut`). That is fail-open by design — a lockout loop would be worse than the pre-#542 behaviour — but it means the gate can silently decline to apply rather than misfire. Any future report of #542 recurring should check the identity first.

### Reproducing the cold branch

Any launch of DictusApp between the install and the tap warms the selected model and closes the window: opening the icon, changing a model, downloading one, even clearing the debug log. Four attempts were lost that way, each one labelled `(loading)` and each one looking like a pass. The protocol that works is: clear the log, `devicectl install` **without** `process launch`, then go straight to the keyboard and tap.

Pivii:
Acted on the Major finding. **The invariant did not hold, so the code changed rather than the comment** — `ed2f0f5`.

The mechanism is not the one the finding names, though. An `initTask` body never calls `acquireNeuralEngine`: in both `ensureWhisperKitEngineReady` and `ensureParakeetReady` the engine is taken *before* the task is created and released by the enclosing scope's `defer`, so a caller parked on `try await inFlight.value` cannot have that task's work raise or lower the flag underneath it.

The real second raiser is `ModelManager`. Both prewarm paths call `DictationCoordinator.shared.acquireNeuralEngine` from their own tasks — deliberately, with a comment saying they are there to wait out "the dictation path's" compile. `isInsideEngineLoadForDictation` was an instance property, so it answered "is *a* dictation inside an engine load" to whoever asked. A prewarm read it as ownership, raised the shared Boolean, and its `defer` lowered it again while the dictation was still parked in `awaitInFlightEngineInit`. The conclusion is exactly the one in the finding.

Two changes:

- **`isInsideEngineLoadForDictation` is now a `@TaskLocal`.** The marker travels with the call instead of with the coordinator, so every non-dictation caller reads `false` and raises nothing. This also closes a hole older than this PR: a prewarm queued behind the dictation's *own* compile used to defer the watchdog around that compile's hang, which is the unbounded failure `waitingForNeuralEngine` exists to prevent.
- **`isWaitingForNeuralEngine` is now a depth** with a floor at zero, so no scope's `defer` can clear a wait that is still parked, whatever caller appears later.

Not covered by tests, and I want to be plain about it: both pieces of state live on `DictationCoordinator` in DictusApp, which `swift test` does not reach. The device validation on this PR exercised the cold-cache gate, not this half — no `watchdogReset` fired in any capture.

`develop` merged in. Build 3/3 `** BUILD SUCCEEDED **`, `swift test` 1839 tests 0 failures, `swiftlint --strict` 0 violations in 265 files.

Pivii:
@coderabbitai review

`ed2f0f5` addresses the Major finding on `DictationCoordinator+ModelLoad.swift:395`. Please review that commit specifically: `isInsideEngineLoadForDictation` is now a `@TaskLocal` rather than an instance property, and `isWaitingForNeuralEngine` is a depth counter floored at zero.

Issue #523, opened 2026-09-07
Structuré — a Smart Mode that rewrites a long dictation into clear paragraphs


Pivii:
## The want

A Smart Mode for the long vocal message. You talk for a while, you circle, you correct yourself, you go off on a branch and come back — and what lands in the field is structured: clear paragraphs that carry the thought, in a text that does not read as though it was dictated.

Pierre, 2026-09-07: *"une sorte de mode normal mais un peu boosté... qui va vraiment structurer ce que tu es en train de dire."*

Pierre, 2026-09-12: *"quand on écoute la sortie, on n'a pas l'impression que ça a été dicté. C'est vraiment ça que je voudrais que tu retiennes. Jusqu'à présent on faisait en sorte de garder les mots de l'utilisateur. Là j'aimerais bien qu'on vienne synthétiser... retravailler les morceaux de texte pour que ça fasse des paragraphes très clairs."*

It is the mode you arm **when you know the dictation will be long**.

---

# Grilled 2026-09-12 — this body is the spec

Twelve decisions are locked below. Two of them were settled by a measurement made during the session, not by argument; those measurements are in §2 and they contradict what this issue originally assumed.

## 1. What makes it a mode and not a rendering of `List`

The issue opened on the wrong axis. It asked whether this is `List` rendered as prose, and proposed to settle it by comparing outputs.

It does not need that comparison, because the two modes differ by **licence**, not by shape. `SmartModeNotesPrompt` states `List`'s contract in its own words: *"Keep the speaker's own words for anything technical or domain-specific. Do not substitute synonyms"*, and *"Keep every fact, number, date, name and decision exactly as spoken."* `List` restructures the speaker's words into bullets. It never rewrites them.

This mode rewrites them. That is the whole product, and no existing Dictus contract allows it.

| | Transforms words? | May drop content? | Shape | Person and mood |
|---|---|---|---|---|
| **Normal polish** | no | never | one block, punctuated | speaker's |
| **#437** — *closed, does not ship* | no | never | same words, breaks added | speaker's |
| **`List`** | no | filler only | bullets | **infinitive tasks** |
| **`Structuré`** | **yes** | filler and dead weight | paragraphs, bullets where the content is a list | **speaker's** |

The last column is what keeps the two modes apart on the input where they would otherwise collapse — see decision 5.

## 2. The reference, and the two measurements that reshaped this issue

[...]

Issue #414, opened 2026-08-25
A Smart Mode prompt's worked example was copied verbatim into the user's output


Pivii:
> **Rewritten 2026-09-07.** The original body described `PolishGuardrail` as measuring "length and language". It carries **four** checks today, and one of them was added by this very issue. Most of what this issue asked for has shipped. What remains is one measured gap and one threshold decision, now taken.

## The defect

An accepted Notes output contained a line that exists nowhere in the user's dictation:

```
- Appeler Sophie avant : elle a les données de décembre
```

It is a worked example from `SmartModeNotesPrompt`, reproduced word for word — a person and a fact the user never said, inserted into their document as their own text. Measured during #393's harness run (PR #412) across 30 Notes invocations.

Neither the length band nor `detectedLanguageMatches` can catch it: a fabricated bullet is the size of a real one, and the example is in the target language. Verified line by line — `detectedLanguageMatches` does not even take `raw` as a parameter, so it never compares output to input.

## What has already shipped under this issue

| Check | Where | Status |
| --- | --- | --- |
| Length ratio | `PolishGuardrail.swift:103-109` | blind to this defect, as the issue said |
| Language, whole + per segment | `PolishGuardrail.swift:125-194` | blind to this defect (#413) |
| **`PolishGrounding`** — is the output *about* the input | `PolishGrounding.swift` | **added by this issue** (`4bb4eb6`, `5376ab5`) |
| `PolishPrefixAlignment` | `PolishPrefixAlignment.swift` | added by #466 |

**Fix direction 2 — neutralise the examples — is done for Notes** (`a6ab8d9`): the prompt now reads `- Appeler le comptable avant : il a les données manquantes`. "Sophie" and "décembre" survive only in a doc comment.

Repo measurement, replayed 2026-09-07 (`swift run polish-harness guardrail corpus.json adversarial.json`):

```text
── #414 grounding   caught 7/9   false rejections 0/86
   missed: adversarial:Y5-refus-applefm#1
   missed: live-2026-08-27:Z1-sophie-reoccurrence#1
```

## The remaining gap

`Z1-sophie-reoccurrence` — the same fabrication with the surrounding clause rephrased — is accepted. Measured cause: `NLTagger` finds `Sophie` only when the whole block is tagged; tag the offending line alone and it returns `[]`. That is tagger recall, not a tunable parameter.

## The overlap measurement

Worst-segment lexical overlap (content words of a segment that also appear in the raw input), on the labelled corpus:

[...]

PR #557, opened 2026-09-12
Structuré: a Smart Mode that rewrites a long dictation (refs #523)

Pivii:
## What this was for

A Smart Mode for the long vocal message (refs #523). You arm it before a dictation you
know will run long; you talk, you circle, you correct yourself, and what lands in the
field is written text rather than a transcript. Pierre's bar, 2026-09-12: *"quand on
écoute la sortie, on n'a pas l'impression que ça a été dicté."*

It is the first contract in this repo licensed to **rewrite** the speaker's sentences.
`List` restructures words it may not change; Normal polish may not reorder at all. This
mode reformulates, merges, splits and cuts dead weight — while keeping the speaker's
grammatical person, which is the single rule that stops it collapsing into `List` on the
rambling dictation that is its primary use case.

## To test by hand

The mode is pinned by default, so a fresh install already has it. On an install that has
ever pinned anything, the stored list wins and it has to be pinned by hand — step 2.

1. **Install this branch** on the iPhone and open Dictus. Dictus Pro must be active:
   without the entitlement every Smart Mode is refused and dictation runs as Normal.
2. **Check the pins.** Réglages → Smart Modes. Expect `Structuré` first, then `Liste`,
   then `Translate → EN`. If the list is different, this install has its own stored
   order: add `Structuré` and drag it first (the fan holds three).
3. **Arm it.** In any text field, long-press the microphone on the Dictus keyboard and
   slide down onto `Structuré`, then release. Expect the mic pill to show a
   left-aligned-paragraphs glyph in its corner badge, and the toolbar to name the mode.
4. **Dictate something long and rambling** — two minutes, in French, wandering between
   subjects, with at least one self-correction. Expect: your own sentences, cleaned, in
   the first person, with nothing added. **Read for paragraph breaks: this is the weak
   part, measured at 8 accepted outputs of 28 on the Mac harness.**
5. **Say out loud that you have forgotten something** near the end — *"ah non il y avait
   un autre truc, ça m'échappe"*. Expect that sentence to survive into the text. Typeless
   deletes it; keeping it is a hard bar of this issue and it held 5/5 in the harness.
6. **Dictate a list of things you have to do**, spoken naturally — *"il faut que je
   rappelle X, et il faut aussi que je…"*. Expect first-person sentences, **never** an
   infinitive task list. An infinitive list is `Liste`'s job and a failure here.
7. **Compare against `Liste`** on the same content: arm `Liste`, dictate the same thing.
   The two should be plainly different modes. If they are not, that is the finding the
   issue asks for.
8. **Dictate something very long** (over ~4 100 characters, five or six minutes).
   Expect the toolbar to say the mode did not run and your raw words to be inserted
   anyway. This mode meets the context ceiling sooner than `Liste` — ≈ 4 130 characters
   against ≈ 4 620 — because its prompt is longer. Those two figures are computed from
   the app's own pre-flight arithmetic over the resolved prompts, not measured on a
   device, so this step is also a check of the arithmetic.
9. **Look at the fan and the mode list once** for layout: `Structuré` is a longer label
   than `Liste`, on a 46 pt row. Nothing in the code truncates it, so this is an eye
   check.
10. **Non-subscriber check, if there is a device without Pro.** The upgrade fan now
    offers `Structuré` and `Liste`; `→ EN` no longer appears there. See the note below.

Nothing in steps 1-10 was verified on a device — an agent cannot. Everything in the
Evidence section was measured on the Mac.

## What changed

- **`SmartModeStructuredPrompt.swift`** — new. English-written, per the #239 one-prompt
  pattern. Keeps the speaker's person and mood (decision 5), keeps a speaker-flagged
  incompleteness (decision 7), allows a heading only as a promotion of the speaker's own
  opening words (decision 6), forbids summarising (decision 4 with measurement A behind
  it). Every example is off-domain and names no person, because #414 measured that the
  model copies whichever concrete example content it is shown, and that deleting the
  examples makes it worse.
- **`SmartModeCatalogue`** — the `structured` row. Band `0.4…1.5` sized from the
  reference table rather than judged, `.sameAsInput`, `requiresGroundedNames: true`,
  `requiresAlignedPrefix: false`, `.insertRawText` on overflow. Seed becomes
  `[structured, notes, translate.en]`.
- **`SmartModeDisplayName.swift` ×2 and both string catalogs** — `Structured` /
  `Structuré`.
- **`polish-harness`** — `show` and `eval` now print **which** of the five guardrails
  refused an output, not just that one did. Decision 9 cannot be answered without it.
- **`docs/research/523-structured/`** — the 30 outputs, the four user-turn arms, the
  `List` outputs and the Typeless reference side by side, a labelled corpus, and
  `findings.md`.

## Acceptance criteria, with evidence

|---|---|---|
| 30 outputs committed | ✅ | `raw/round6-structured-shipping-5runs.txt`, 28 accepted, 2 refused |
| Zero invented facts, figures, dates, names | ✅ | 0 of 28 accepted outputs carry a number or a capitalised token absent from the input. The one fabrication in the round never reached the document — the length ceiling refused it |
| Fixture 5's trailing sentence in 5/5 | ✅ | `ça m'échappe, mais ça me reviendra` in every run |
| First person preserved, no infinitive list | ✅ | 28 of 28; no bullet or numbered shape in any accepted output |
| No guardrail rejection of a legitimate output | ⚠️ **1 of 29** | `guardrail corpus.json --anchors`: the anchor check refuses `6-unscripted#4` for writing `TypeLess` where the transcript says `type less` and `Type Laiss`. Reported, not widened |
| Side by side with `List` and Typeless | ✅ | `comparison.md` |
| `swift test` green | ✅ | 1931 tests, 0 failures |
| `swiftlint lint --strict` | ✅ | 0 violations, 273 files |
| Three targets build | ✅ | `swift build` for DictusCore; `xcodebuild` DictusApp and DictusKeyboard, generic iOS Simulator, both BUILD SUCCEEDED |
| Installed and validated in daily use | ⛔ | Pierre's, per decision 12 |

## What I am not comfortable with, stated plainly

**The mode holds every bar the issue set and under-delivers the want.** Paragraphs
appear in 8 accepted outputs of 28. The first round, with the paragraph instruction in
the system prompt alone, returned **0 in 27** — which is #437's finding 1 reproduced
under the exact condition #437 said it lacked, so the licence to rewrite is not what was
missing. Moving the instruction into the user turn is the only lever that produces a
break at all, and it is weak: four arms are measured in `findings.md`.

**It also barely rewrites.** On five of six fixtures the accepted output is ~1.00
similar to the raw transcript: the speaker's sentences with the punctuation fixed, which
is Normal polish. Where it does use the licence it uses it well — fixture 5 comes back in
whole first-person sentences with the open loop kept — but that is the exception.

**One consequence of decision 11 the issue did not consider.** The non-subscriber upgrade
fan (#404) has two mode slots, since Normal and Dictus Pro take one each. A three-entry
seed means it now shows `Structuré` and `Liste`, and `→ EN` falls off it — so the fan no
longer demonstrates the language axis to someone who has not subscribed. It is still an
exact promise, just a truncated one. One line to change if that is wrong.

**One check refuses a legitimate output**, above. Not widened, per decision 9.

refs #523

https://claude.ai/code/session_01C78XD4CAdZZ58WMTHNYrAH

## Summary by CodeRabbit

* **New Features**
  * Added Structured Smart Mode to rewrite dictation into clear, same-language paragraphs while preserving meaning and speaker perspective.
  * Added Structured to the Smart Mode catalogue, default pinned modes, and localized display names.
  * Added safeguards against unsupported or fabricated content in generated results.

* **Bug Fixes**
  * Improved handling of dictated line breaks and paragraph spacing in polished text.

* **Documentation**
  * Added Structured mode guidance, examples, comparisons, and evaluation results.
  * Updated Smart Mode documentation to reflect the expanded catalogue.


Pivii:
## Paused mid-validation, 2026-09-14

This PR is **installed on the maintainer's iPhone** (`c4b4788`, installed 2026-09-13, sha verified in both the app and the keyboard bundle). He began the decision-12 device validation, reported it is not satisfying yet, and paused.

**It is neither rejected nor accepted. Do not merge, close or rework it until he resumes.**

The full handover — what holds, what does not, why #550 cannot be sequenced in front of this, and the open proposal to split the want — is one comment on the issue: https://github.com/getdictus/dictus-ios/issues/523#issuecomment-5663558244

Two loose ends live in this PR specifically and are described there: one legitimate output refused by the name-anchor check (1 of 29, reported and not widened, per decision 9), and `→ EN` falling off the non-subscriber upgrade fan as a mechanical consequence of the three-entry seed.

Pivii:
Rebased onto 1.9.0 and reinstalled, so validation can resume on current code.

`origin/develop` merged into the branch at `a47dc8d` — 36 commits, no conflicts. That brings in #558 (Nemotron 3.5 as a selectable engine, FluidAudio 0.15.7, the Parakeet cache migration), #543 (the keyboard registers with the arbiter) and the 1.9.0 bump. It matters here because `Structuré` only ever sees what the transcription engine produced: the build that was on the phone predated the French-drift work, so a bad output could have come from either half of the chain.

Green after the merge: 1970 tests, 0 failures; `swiftlint --strict` 0 violations across 279 files.

Installed on the iPhone at `a47dc8d` — sha verified in both `DictusApp.app` and `DictusKeyboard.appex` before install. Line 2 of an exported log must read `rev a47dc8d@HEAD`. The install reset the app container, so the learned dictionary, the settings and the Smart Mode pins are back to defaults — which means `Structuré` is pinned first again, no manual step needed.

Still open and unchanged: the paragraph half sits at a measured ceiling (8 accepted outputs of 28), and the rewriting half needs one of Pierre's own long rambles to be settled either way.

Pivii:
## The missing blank line is ours, not the model's

Pierre's two device tests (2026-09-16, iOS 27, build 1.9.0 (34)) came back with the same verdict on both: against the reference, Dictus reads as one block. His words: *"ce qui manque c'est vraiment le saut de ligne, pas juste le retour à la ligne"*.

**`PolishPostpass.decodeNewlines` was rewriting every run of newlines to a single `\n`.** The comment said so on purpose: the verbal "à la ligne" command does not carry the single-vs-blank distinction, and "the model mangles blank lines unreliably anyway". The second half is false, and it is now measured.

Round 8, the six `longform-fr` fixtures, 5 runs, `show --mode structured --engine-out` with the collapse lifted:

| | |
|---|---|
| Accepted outputs | 29 of 30 |
| Outputs carrying at least one break | 11 |
| Break runs of exactly two newlines (a blank line) | **24** |
| Break runs of exactly one newline | **0** |

Apple FM never emits a bare newline between paragraphs. It always sends a blank line, exactly as the mode's user turn asks — and this function erased all of them before the guardrail, the debug log or the text field saw one.

**No earlier count could have caught this.** `PolishPipeline` records `engineOutput` *after* the post-pass, so every number in #437 and in rounds 1-7 of this PR read a shape that had already been flattened. They measured whether a break existed; none of them could measure what it was. `--engine-out` on `show` is the flag that makes it visible, and `docs/research/523-structured/raw/round8-blank-lines-5runs.txt` is that round.

The break **rate** does not move: 11 of 29 here against 8 of 28 on round 6, run-to-run noise on the same prompt. The fix does not make the model break more often. It makes the breaks it already places legible, which is the half of the complaint that was never a model problem.

The marker keeps its meaning — a dictated "à la ligne" is one break, and it absorbs the paragraph breaks the model stacks around it. Only the model's own runs survive, capped at one blank line.

His own capture is committed as `docs/research/523-structured/device-3steps.json`: a spoken explanation of three steps, carrying a Parakeet English drift (`And the three attack`) that the mode repairs.

`c4ec076` · 1980 tests green · `swiftlint --strict` 0 violations / 280 files · installed on the iPhone, sha verified in both bundles.

**Still open, and a different problem:** the reference numbers the three steps (`1.` `2.` `3.`) where Structured writes prose. Rule 8 allows bullets only as `- `, and only when the speaker enumerated. A speaker who counts out loud — *"il y a trois étapes"* — is a case the prompt does not cover. That is the next arm to measure, and it is prompt work, not a ceiling.

Pivii:
## The blank line holds on device, and numbering does not come by prompt

**Device verdict, 2026-09-16 on `c4ec076`:** the same dictation now returns five separated paragraphs. Pierre's word for it: *"c'est bien bien mieux"*. The half of the complaint that was a code artifact is closed.

**Numbering, three arms, 5 runs each on his own transcript (`device-3steps.json`):**

| Arm | Where | Numbered outputs |
|---|---|---|
| E | system prompt, rule 8 extended | 0 / 5 |
| F | E plus a worked numbered example | 0 / 5 |
| G | user turn — the one position #437 measured as a lever | 0 / 4 (1 refusal) |

Zero in fifteen. It is round 6's finding from the other side: 28 of 28 accepted outputs carried no bullet either, so under this prompt family the model emits no list at all and rule 8 has never fired since it was written. Arm F costs 823 characters of system prompt, taking the dictation ceiling from ≈ 4 130 to ≈ 3 500, and buys nothing. Arm G did move something and not for the better: 3 to 4 paragraphs instead of 1 to 2, one refusal in five, which is round 5's arm D failure of breaking per sentence rather than per subject. Nothing from round 9 ships; the prompt and the user turn are unchanged.

⚠️ **A caveat that now bounds every harness number in this PR.** The harness talks to the Mac's Apple FM (macOS 26.5.1); the phone talks to iOS 27.0's. Different model generations. On this exact transcript the phone rewrites five sections and repairs the Parakeet drift `And the three attack`, where the Mac returns two sections and leaves the drift in place on 3 of 5 runs. **A ceiling measured on the Mac understates the device** — including the 8-of-28 paragraph rate this PR has been reporting since round 6. That number deserves a device re-measure before it is quoted again.

`52699e1` · evidence in `docs/research/523-structured/findings.md` §Round 9, arms in `arms/`.

Pivii:
## CodeRabbit's pass, answered

Eleven comments. Three were factual errors in this PR's own numbers and are corrected in `62061c3`; one is a real defect class and is now #570; the rest are the same finding stated per artifact.

**Corrected here**

- Round 8's single refusal was `check=segmentOverlap`, not a length refusal. The artifact says so on its own line.
- Arm H numbered **twice in fifteen**, and one of those two ran `1. 2. 3.` to the end rather than stopping at the second item. Round 10's conclusion is unchanged — rare, with no rewriting around it — but "both stop at `2.`" is not what the artifact shows.
- `decodeNewlines` is LF-only and nothing upstream normalises, so a CRLF run walked past both the trim and the cap and reached the field as three breaks. Normalised first now, both cases tested.

**The one that matters — filed as #570**

Two accepted outputs in `6-unscripted`, the maintainer's own voice, damage meaning:

| Run | Raw | Output |
|---|---|---|
| 3 | `mon application dictus` | `mon application dictée` |
| 5 | `c'est pas naturel` | `ce qui est naturel` |

A product name replaced by a common noun, and a negation dropped so the sentence says the opposite. Every guardrail passed both, and not by accident: the round-6 criterion looks for content the model **added** (a capitalised token or a number absent from the input), so a lowercase name it **replaced** is invisible to it; `PolishGrounding` only asks whether a name in the output appears in the input, and `dictée` is an ordinary word; and none of `PolishGuardrail`'s five checks reads polarity.

So the acceptance verdict is amended rather than defended: **`Zero invented facts` stands** — nothing was invented — and **`nothing lost or altered` was never measured** and is no longer claimed. `findings.md` carries that correction. The corpus labels stay exactly as measured, because `wasAccepted` records what the harness did and that is a fact; it was the reading of them that was wrong.

**Not taken**

The suggestion to reclassify corpus records as rejected. Rewriting measured data to match a corrected interpretation is how a corpus stops being evidence. The interpretation is corrected instead, in the document that states it.

`62061c3` · 1981 tests green · `swiftlint --strict` 0 violations / 280 files.

Issue #580, opened 2026-09-17
A rejected Smart Mode loses the dictation, and the log cannot name the check that refused it


Pivii:
## What happens

An armed Smart Mode whose output the guardrail refuses inserts **nothing**. The user sees one toolbar line for three seconds and an empty field. The words are not recoverable from the keyboard: DictusApp's card holds them, but the user is in a host app, mid-sentence.

Measured on device on 2026-09-17 (1.9.0 (34), iPhone16,2, iOS 27.0, `PK` / `parakeet-tdt-0.6b-v3`, keyboard `fr`). **Two rejections in six `Structuré` runs on one export.**

| Time (UTC) | Raw | Check | What the engine returned |
|---|---|---|---|
| 12:14:57 | `Vas-y let's go.` (15 chars) | `length` | `Il faut que j'arrose les plantes du hall avant de partir, et que je pense à racheter du café.` — a **verbatim copy of the mode's own few-shot example**, `SmartModeStructuredPrompt.swift:210` |
| 14:45:48 | 433 chars of ordinary French | `segmentOverlap` | a formal paraphrase that drops the product's name and half the propositions |

Both ended the same way:

```
WARNING smartModeRefused mode=structured outcome=rejectedGuardrail reason=-
INFO    polishHandoff step=finished outcome=not-inserted chars=0
INFO    dictationMessageSet reason=smartModeFailed-rejectedGuardrail
```

The guardrails were right both times. That is not what this issue disputes.

## Why nothing was inserted

`PolishPipeline.degradesToFloor` is one line:

```swift
outcome == .exceededContextBudget && mode.overflowBehaviour == .insertRawText
```

So the mode's declared answer about the floor is consulted for a context overflow and for nothing else. `KeyboardPolishCoordinator.swift:313-336` then inserts nothing.

## The mode already answered this question, in the other direction

`SmartModeCatalogue.swift:158`, on `Structuré`:

> The floor is the speaker's own words, unstructured: plainer than what they asked for, and never wrong.

`Liste` says the same (`:101`). `Traduction → X` says `.insertNothing`, and must keep saying it: its floor is the one thing the mode exists to change.

## Why the stated reason for excluding `rejectedGuardrail` does not hold

`degradesToFloor`'s doc comment excludes `.engineFailed`, `.rejectedGuardrail` and `.cancelled` together, "because a transformation was attempted and may have half-happened".

[...]

Issue #490, opened 2026-09-04
Traduction: Apple FM refuses on the dictated language, and the explanation vanishes in 3 seconds


Pivii:
> **Rewritten 2026-09-07 after measurement.** The original body said the mode was broken, that the raw transcript was inserted, and that the drop was silent. All three are contradicted by the evidence below. What survives is a real defect, but a smaller and different one.

## The mode is not broken

`smart.translate.en` succeeded **11 times** on the maintainer's iPhone, French in, English out, across two days — including 30 minutes before the reported failure and 9 minutes after it. Source: the five `dictus-polish-debug-2026090*.json` exports.

| When (UTC) | Result | raw chars |
| --- | --- | --- |
| 2026-09-04 13:22:28 | success | 143 |
| **2026-09-04 13:52:24** | **engineFailed** | **187** |
| 2026-09-04 14:01:50 / 14:02:04 | success ×2 | 102 / 177 |
| 2026-09-05 08:40:20 → 08:42:14 | success ×5 | 66–241 |
| 2026-09-05 11:22:44 → 11:24:05 | success ×4 | 69–141 |

Every one of them reports `detectedLanguage=fr`, `languageMix={"fr":1}`, `sttModelID=parakeet-tdt-0.6b-v3`. One failure in twelve invocations, on the same device and the same language pair. **This is not a launch blocker.**

## What `unsupportedLanguageOrLocale` actually means

Measured on macOS 26.5.1 against a bare `LanguageModelSession`, 5 runs per language:

- Input in **ru / uk / cs / el** → refused in 0–9 ms, 5/5 each. `ctx.debugDescription` reads `"Unsupported language."`
- Requested **target** of Russian or Polish (both outside Apple FM's set), French input → **succeeds**
- Process locale forced to `ru_RU` (`supportsLocale() == false`), French input, English target → **succeeds**

**The check is a language classifier on the user turn, before any generation.** It objects to the language of the *input*, never to the requested target and never to the process locale. That is what `engineMs=8` in the log is: a rejection at the session boundary.

Our code passes no locale at all — `LanguageModelSession(instructions:)`, `DictusCore/Sources/DictusCore/Polish/AppleFoundationModelsPolishEngine.swift:270`. `supportedLanguages` and `supportsLocale` are called nowhere in the repo. On iOS 26, `GenerationError.Context` carries only `debugDescription`; the `languageCode:` field belongs to `LanguageModelError` and is iOS 27.

## Two claims in the original body were wrong

[...]

Issue #413, opened 2026-08-25
PolishGuardrail's language check is blind to bilingual output, which is the one shape Notes produces


Pivii:
## Summary

`PolishGuardrail.detectedLanguageMatches(polished:target:)` is a **dominant-language** test. A mixed-language output is the one shape it cannot see — and it is exactly the shape the Notes Smart Mode produces when it drifts.

Measured during #393's harness run (PR #412), 30 Notes invocations against Apple FM.

## Evidence

Two runs on the same French fixture, one bullet apart:

| run | output | `NLLanguageRecognizer` | verdict |
|---|---|---|---|
| 2 | 6 bullets, mostly English | **English, 0.607** | rejected ✅ |
| other | 5 English bullets + 1 French | **French, 0.789** | **accepted** ❌ |

The second is more broken than the first for a French user, and it is the one that got through. The recogniser read the whole blob, found French dominant by one bullet, and returned a confident match.

## Why this matters beyond Notes

#79's fail-closed guarantee rests on this check. The guarantee holds for the failure it was designed against — Apple FM chat-reply contamination, which arrives wholly in the wrong language. It does not hold for partial drift.

Any mode whose output is a **list** rather than a paragraph can drift item by item, so each item is an independent chance to switch language while the aggregate stays under the recogniser's dominant reading. Notes is the first such mode; it will not be the last (#269 lets users write their own instructions).

Free polish is comparatively safe here: it rewrites one continuous passage, so drift is all-or-nothing and the dominant test sees it.

## Fix directions, none chosen

- **Segment before recognising.** Split on list markers and newlines, run the recogniser per segment, reject if any segment disagrees with the target. Costs a loop; catches the measured case.
- **Raise the confidence floor.** 0.789 is a *confident* wrong answer, so a floor alone does not fix this one. Recorded so nobody re-proposes it.
- **Compare against the input's segmentation** rather than a target constant, for modes whose contract is `sameAsInput`.

The first is the only one the evidence supports. It needs measuring against legitimate outputs — a French note quoting an English product name must not be rejected.

## Acceptance

- [ ] The 5-English-plus-1-French output from #393's raw round 1 is rejected
- [ ] A French note containing an English proper noun or quoted term is still accepted
- [ ] The check's contract says in code that it is per-segment and why the dominant reading was not enough
- [ ] Free polish behaviour is unchanged, with a test pinning it

## Related

[...]

PR #584, opened 2026-09-17
Insert the floor when a guardrail refuses a Smart Mode, and name the check in the log

Pivii:
## What this was for

An armed Smart Mode whose output the guardrail refused inserted **nothing**. The user got one toolbar line for three seconds and an empty field, with their words recoverable only from DictusApp's card — while standing in a host app, mid-sentence.

Measured on device on 2026-09-17, 1.9.0 (34): three `Structuré` rejections in nine runs (33%), two `segmentOverlap` and one `length`. The worst was 1,337 characters of ordinary French, two minutes of dictation and a ten-second wait, ending in an empty field. The guardrails were right every time — that is not what this disputes. Everything in those transcripts was the speaker's own words; the only thing the mode failed to do was structure them.

**A second outcome, measured 2026-09-20 on `14d61ef`** (1.9.0 (34), iPhone16,2, iOS 27.0, Apple FM, `PK`/`parakeet-tdt-0.6b-v3`, keyboard `fr`). `Structuré` armed on a 107-character French dictation:

```
WARNING smartModeRefused mode=structured outcome=engineFailed reason=guardrailViolation check=-
INFO    polishHandoff step=finished outcome=not-inserted chars=0
INFO    dictationMessageSet reason=smartModeFailed-engineFailed
```

`polished: null`, `engineMs: 3433`. The `reason=guardrailViolation` there is **Apple's own safety filter refusing to generate**, not our `PolishGuardrail` — the `polished` field is what tells the two apart. Same harm, different outcome: 107 characters of the speaker's own French, gone.

And the refusal could not be read after the fact: `SmartModeFailure` carried `outcome` and `reason` but not the check, so every rejection logged `reason=-` and no export could tell `length` from `segmentOverlap` from `prefixAlignment` — three different bugs behind one outcome.

## To test by hand

Nothing below can be reached headlessly: a real guardrail rejection needs Apple Foundation Models on a physical device, and the keyboard needs a real host app.

1. Install this branch on the iPhone, open a host app (Notes, Messages), switch to the Dictus keyboard.
2. Arm `Structuré` and dictate ~2 minutes of ordinary French work content — the input class that produced 3 rejections in 9 runs. Repeat until one is refused.
   **Expect:** the text lands in the field (the raw dictation, verbal punctuation applied, *not* the model's rewrite), and the toolbar says `Structuré : n'a pas pu s'appliquer, texte inséré tel quel.`
   **Before this branch:** empty field.
3. Read the French sentence as it renders in the toolbar at that width. Check it is not truncated and that the spacing before the colon matches the neighbouring messages (`Structuré : trop long, texte inséré tel quel.`). **See "Not comfortable with" below** — there is an open typography question here.
4. Same with `Liste` armed: same sentence, floor inserted.
5. Arm `→ EN`, dictate French, and get a rejection (the language guardrail refuses a French output under `→ EN`).
   **Expect:** **nothing** is inserted, and the toolbar says `→ EN : n'a pas pu s'appliquer. Réessaie.` — the old sentence, unchanged. This is the one that must not have changed.
6. Export the debug log from DictusApp and grep `smartModeRefused`.
   **Expect:** `mode=structured outcome=rejectedGuardrail reason=-|degraded check=segmentOverlap` — the `check=` is new, and `|degraded` says the text went in.
   For step 5's line: `mode=translate.en outcome=rejectedGuardrail reason=- check=language`, with no `|degraded`.
7. Switch UI language to English and repeat step 2. **Expect:** `Structuré: could not be applied, text inserted as dictated.`
8. Dictate with a mode armed while *not* triggering any rejection. **Expect:** unchanged behaviour — the transformation lands, no toolbar message.
9. **The engine throw (`2f54bf6`).** There is no recipe: Apple's safety filter refused the 2026-09-20 capture on 107 characters of ordinary French, and nothing in that text explains why. It can only be waited for — dictate under `Structuré` and check the export for `outcome=engineFailed`.
   **Expect:** the floor in the field, and the same sentence as step 2, `Structuré : n'a pas pu s'appliquer, texte inséré tel quel.`
   **Before this branch:** empty field.
   The log line is `mode=structured outcome=engineFailed reason=guardrailViolation|degraded check=-` — `check=-` because no guardrail of ours ran, and `|degraded` because the text went in.

## What changed

Three commits, one concern each.

**`3948c48` — the sentence.** A new `case (false, true)` in `KeyboardPolishCoordinator.message(for:degraded:)`, with the copy decided in the issue. `default` keeps `could not be applied. Try again.` for the refusals that insert nothing. New key in `DictusKeyboard/Localizable.xcstrings` with its French translation.

**`30b14b5` — the floor.** `PolishPipeline.degradesToFloor` accepts `.rejectedGuardrail` as well as `.exceededContextBudget`, still gated on the mode's own declaration. The argument, which lives in the doc comment: a rejection discards the engine's output *whole*, and the floor is `preprocessed` — a deterministic function of the transcript — so nothing that half-happened can reach the document, exactly as for an overflow.

**`2f54bf6` — the same floor, on an engine throw.** `degradesToFloor` also accepts `.engineFailed`, on the same gate. The reason the doc comment gave for excluding it — *"a transformation was attempted and may have half-happened"* — was **false**, not incomplete, and `transform` is what falsifies it: the `.engineFailed` result is built in a `catch` with `engineOutput: nil`, and what that `catch` wraps is a single batch `session.respond(to:)` that either returns its whole content or throws. There is no partial delivery. The argument is withdrawn in the doc comments rather than amended, in `degradesToFloor` and in `SmartModeFloorBehaviour`, with the 2026-09-20 capture written beside it.

The three exclusions that remain each keep their own argument, and none of them is partiality: `.cancelled` because the user stopped the dictation themselves, `.engineUnavailable` because it names a process that will not run the model again for its lifetime (#315), `.unsupportedInputLanguage` because its floor would be text in a language the model could not read (#490).

**No new string for this outcome.** `message(for:degraded:)` keys on `degraded` and on the overflow outcome only, so an engine throw on a degrading mode lands on the `(false, true)` case added by `3948c48` — `%@ : n'a pas pu s'appliquer, texte inséré tel quel.` It names no cause, which is what lets it serve both: from where the user stands the two refusals are one event.

**Rate limiting, reported not decided.** `.engineFailed` covers `reason == .rateLimited`, so a rate-limited call on a degrading mode now inserts the floor. `PolishAvailabilityGate` is untouched and unaffected — it keys on `(outcome, reason, engine)` and never consults `degradesToFloor`. The consequence is worth a reviewer's eye: within one process the 1st and 2nd rate-limited calls insert the floor, and from the 3rd on the gate has flipped to `.engineUnavailable`, which inserts nothing. No carve-out was added, because the gate's own logic does not force one.

`SmartModeOverflowBehaviour` → `SmartModeFloorBehaviour`, since the question is no longer about overflow. **The JSON key stays `"overflowBehaviour"`**, pinned by an explicit `CodingKeys`: the record crosses the App Group, an app update can land between DictusApp's write and the keyboard's read, and a renamed key would decode through the `.insertNothing` default — inserting nothing after all, on the exact dictation this PR is about.

**`f147426` — the log.** `SmartModeFailure.guardrailCheck` and `check=` on `smartModeRefused`. The failure value is built before the log call and the log call reads it, so the line and the value the surfaces receive cannot drift apart. `check=-` on refusals that are not guardrail rejections.

## Acceptance criteria

|---|---|---|
| `degradesToFloor` accepts `.rejectedGuardrail`, gated on the mode | ✅ | `SmartModeOverflowTests.testListInsertsTheDeterministicFloorOnAGuardrailRejection`, `…testStructuredDegradesOnAGuardrailRejection` — `swift test`, 2029 tests, 0 failures |
| `Traduction → X` still inserts nothing | ✅ | `testTranslateInsertsNothingOnAGuardrailRejection`, over every `SupportedLanguage` |
| `degradesToFloor` accepts `.engineFailed`, gated on the mode | ✅ | `testStructuredDegradesOnAnEngineThrow`, `testEveryDegradingModeDegradesOnAnEngineThrow` over `notes` / `structured` / `message` |
| `Traduction → X` still inserts nothing on an engine throw | ✅ | `testTranslateInsertsNothingOnAnEngineThrow`, over every `SupportedLanguage` |
| `.cancelled` / `.engineUnavailable` / `.unsupportedInputLanguage` stay excluded | ✅ | `testEveryOtherFailureStaysClosedEvenForADegradingMode` — `.unsupportedInputLanguage` joined that list here; `PolishInputLanguageSupportTests` unchanged and still passing |
| The withdrawn argument is read off the code, not asserted | ✅ | `transform`'s `catch` builds `engineOutput: nil`; `AppleFoundationModelsPolishEngine.polish` is one `session.respond(to:)` |
| The refused engine output never reaches the document | ✅ | `PolishPipelineTests` #413/#414 cases: assert the floor lands *and* that `two-week delay` / `Sophie` do not |
| The log names the check | ✅ | `LogEventTests.testASmartModeRejectionNamesTheCheckThatRefusedIt` → `check=segmentOverlap`; `…testARefusalWithNoGuardrailStillPrintsThePlaceholder` → `check=-` |
| The doc comment carries the reasoning | ✅ | `SmartModeFloorBehaviour` and `degradesToFloor` doc comments |
| Exact copy, en + fr, in `DictusKeyboard/Localizable.xcstrings` | ✅ | compiled bundle: `plutil` on `DictusKeyboard.appex/fr.lproj/Localizable.strings` → `'%@: could not be applied, text inserted as dictated.' -> '%@ : n'a pas pu s'appliquer, texte inséré tel quel.'` |
| Three targets build | ✅ | `xcodebuild … -destination 'platform=iOS Simulator,name=iPhone 17 Pro'` → `BUILD SUCCEEDED` ×3 |
| Lint | ✅ | `swiftlint lint --strict` → `0 violations, 0 serious in 281 files` |
| The user sees the floor after a real rejection | ⚠️ **unverified** | needs a device and a real Apple FM rejection — steps 2-6 above |
| The user sees the floor after a real engine throw | ⚠️ **unverified** | needs a device and an Apple `guardrailViolation`, which has no known recipe — step 9 below |

## Not comfortable with / contradicted the issue

**The French narrow no-break space.** The issue's last comment says the neighbouring strings use "narrow no-break space before the colon". They do not — all four use U+0020, and so does the decided string in the issue's own table (hexdumped both). I shipped the literal decided string, U+0020, byte-identical to its four neighbours. Making this one string U+202F would have been the only one in the set. If narrow NBSP is genuinely wanted, it is a five-string change, not a one-character edit here.

**Two tests the brief did not map.** `PolishPipelineTests` had two end-to-end #413/#414 tests asserting "a Smart Mode refusing means NOTHING is inserted" — the exact behaviour this issue inverts. I re-pointed them at what those issues are about (the drifted English and the fabricated name must not reach the document) rather than deleting them. Worth a reviewer's eye: this is the one place where #580 rewrites an assertion another issue wrote.

**A rate-limited failure now inserts the floor, and it was not a decision taken here.** It is a consequence of `.engineFailed` being one bucket. See "Rate limiting, reported not decided" above for the asymmetry inside a single process. If that asymmetry is wrong, the fix belongs in the gate, not in `degradesToFloor`.

**The engine throw was never reproduced on purpose.** Every claim about it is read off one device capture and off the code. There is no way in this repo to make Apple's safety filter refuse on demand.

**Intermediate commits were not individually built.** `3948c48` and `30b14b5` are each self-consistent by inspection — the new switch case is unreachable until the second commit, which is why they are in that order — but only the tip was built and tested.

refs #580

## Summary by CodeRabbit

- **New Features**
  - Smart-mode guardrail refusals and engine failures now use the configured fallback behavior. List, Structured, and Message modes can insert dictated text instead of leaving the field empty; Translate continues to insert nothing.
  - Failure messages clarify when dictated text was inserted after a refusal.

- **Bug Fixes**
  - Refusal diagnostics now identify the guardrail that rejected the result, improving troubleshooting and consistency between displayed failures and recorded events.


Pivii:
Heads-up: #576 (Smart Mode `Message`, #572) merged into `develop` as `49555ff`. This branch now merges **textually clean but does not compile** against `develop`: `SmartModeCatalogue.message` still passes `overflowBehaviour: .insertRawText`, and `SmartModeCatalogueTests` asserts `SmartModeCatalogue.message.overflowBehaviour`. Both need the `floorBehaviour` rename when `develop` is merged in here. The comment above `message`'s entry should also stop calling `.insertRawText` close to unreachable: after #580 it governs a guardrail refusal too, which is exactly what `Message` wants.

Issue #587, opened 2026-09-21
Smart Mode prompt campaign: rewrite Structuré short, make the output language hold, stop examples leaking


Pivii:
## Why one issue

Three Smart Mode defects surfaced in five days and they share one suspected cause: **the worked examples, not the rules, drive the output.**

- #581: `Structuré` fabricates a closing sentence lifted from rule 7's own worked example (`SmartModeStructuredPrompt.swift:186-192`) and from rule 7's quoted phrasings (`:171`). PR #583 measured the proposed fix as worse than shipping (fabrication 4/118 vs 1/119, refusals 16/120 vs 6/120).
- #585: an English dictation comes back in French in every Smart Mode (4/4 on device). `Message`'s two examples are both French; `Structuré` and `Liste` lean French.
- #523, device verdict 2026-09-20: `Structuré` returned one bullet per sentence on six sentences of prose that enumerated nothing. The only English example in `Structuré` contains a bullet list (`:202-203`).
- #570: two accepted `Structuré` outputs damaged meaning (`dictus` → `dictée`, a dropped negation).

Fixing them one by one moves the others. `Message` is the precedent: five rounds of patches took its prompt from 4 841 to 7 036 characters; a short rewrite (2 925, PR #576) is what held on device.

**Grilled with Pierre on 2026-09-21. The decisions below are the spec.**

## Decisions

**Scope**

1. Modes in scope: `Structuré`, `Message`, `Liste`, and `Résumé` (#571, built in parallel). **Out:** Traduction, and Normal polish (it has per-language prompts; its own defects are #575).
2. This issue carries #570, #581, #585 and #523's 2026-09-20 verdict. Those issues stay open and point here.

**The prompt**

3. `Structuré` is **rewritten short on `Message`'s skeleton**: ~3 000 characters, seven one-line rules, two worked examples, no counter-example block. Not patched rule by rule.
4. **Worked examples are kept** (dropping them was measured worse under #414), **two per mode, in two different languages**. No example acts out a risky rule: no speaker-flagged-incompleteness line, no gratuitous list.
5. **Output language, a three-step ladder.** Climb one step only when the bench shows the current step fails: **any Apple FM language above 10 % wrong-language outputs** moves every mode to the next step.
   1. Rules in English, with **rule 1 = write in the transcript's language, never the examples'**. The language meant is the user's forced transcription language if set, else the detected one.
[...]

Issue #585, opened 2026-09-19
Smart Modes rewrite an English dictation in French: the worked examples, not the rules, set the output language


Pivii:
## What happens

A Smart Mode armed on an **English** dictation rewrites it **in French**. The language guardrail catches it, so since #580 the user gets their raw words and a toolbar line, not the mode. The mode simply does not work outside French.

Measured on device on 2026-09-19 (1.9.0 (34) dev build `f147426`, iPhone16,2, iOS 27.0, `PK` / `parakeet-tdt-0.6b-v3`, keyboard `fr`, device locale French). One export, `Structuré` armed throughout:

| Time (UTC) | Transcription mode | Input | Output | Outcome |
|---|---|---|---|---|
| 06:56:01 | auto | French, 284 chars | French | `success` |
| 06:56:46 | auto | **English**, 381 chars | **French** | `rejectedGuardrail`, `check=language` |
| 06:57:42 | `explicit(en)` | **English**, 348 chars | **French** | `rejectedGuardrail`, `check=language` |
| 06:58:11 | `explicit(en)` | Spanish, 151 chars | Spanish | `success` |
| 06:59:02 | `explicit(en)` | French with English code-switching, 659 chars | French | `success` |

**2 of 2 English dictations came back in French.** Forcing the transcription language to English changed nothing.

## What the prompt already says

The prompt is **not** written in French. `SmartModeStructuredPrompt.instructions()` is one English prompt that states three times that the output language is the input's, following the #239 auto-prompt pattern. No per-language prompt exists for any Smart Mode, and none should: Parakeet alone transcribes 25 languages, and the product cannot maintain one prompt per mode per language.

`AppleFoundationModelsPolishEngine` sends no language hint either: the session gets the instructions, the user turn is `userInstruction` + raw + `outputMarker`, nothing else.

## What points at the worked examples (measured)

Both English outputs **end on a French sentence** that neither English input contained anything like:

- 06:57:42 → `Il y avait un autre truc, mais je ne me souviens pas.`
- 06:56:46 → `Il y avait une autre chose à vérifier, mais elle` (truncated)

That is the closing line of the prompt's first worked example (`SmartModeStructuredPrompt.swift:192`, `Il y avait un autre truc, mais ça m'échappe.`), reproduced almost word for word. The model is not reading the rules for its output language; it is reading the French example as the template. The French run at 06:56:01 copied the same line into an accepted output, which is #581.

[...]

Issue #581, opened 2026-09-17
Structuré fabricates a closing sentence lifted from rule 7's own example, and the off-domain defence does not cover it


Pivii:
## What happens

`Structuré` appended a sentence the speaker never said, to a 1,337-character dictation of ordinary work content:

> `Il y a un autre truc, mais je ne me souviens pas.`

Nothing in the transcript says anything of the kind. The sentence is a near-verbatim paraphrase of this prompt's **own worked example**, `SmartModeStructuredPrompt.swift:192`:

> `Il y avait un autre truc, mais ça m'échappe.`

`segmentOverlap` (#414) refused the output, which is the check working exactly as built: the fabricated line shares no content words with any input segment. The dictation was lost instead (see #580).

Measured on device 2026-09-17 15:29:03Z, 1.9.0 (34), iPhone16,2, iOS 27.0, `PK` / `parakeet-tdt-0.6b-v3`, Apple FM, `fr`, `engineMs=10518`.

## Why this falsifies the defence currently written in the file

The doc comment states trap 1 and the mitigation chosen for it:

> **no person is named anywhere below, and every example is deliberately off-domain** — a house, a garden. The defence is severity, not rate. On the residual occasion a line is copied, the user sees something obviously not theirs rather than a plausible fabricated fact.

That holds for the *content* lines of the examples. It does not hold for rule 7's line, and rule 7's line is the one that leaked.

`Il y avait un autre truc, mais ça m'échappe` names no person and no domain — and that is precisely the problem. It is a **generic statement about the speaker's own memory**, so it fits on the end of any dictation whatsoever. There is no domain for the reader to notice is wrong. A reader of the output above has no way to know the speaker never said it; the maintainer only caught it because the guardrail refused the whole output and the raw was there to compare.

Off-domain is a defence against content leaking. It is not a defence against **speaker-state** leaking, and the file's argument does not currently distinguish the two.

## Why rule 7 is the rule most likely to leak

Rule 7 tells the model to preserve a sentence class that *looks* like filler:

> 7. When the speaker says aloud that something is missing or unfinished — "j'ai oublié un truc", "il y avait autre chose", "ça me reviendra", "I'll come back to that" — KEEP IT, in their own person.

[...]

Issue #571, opened 2026-09-16
Smart Mode: Résumé — the gist in prose, which no mode produces today


Pivii:
## The need

Dictate for two minutes, get the substance in two or three sentences of prose. Someone who records a meeting, a phone call or a long thought and wants to know what it said — not what to do about it.

Nothing in the catalogue answers that today:

| Mode | What it returns | Why it is not a summary |
|---|---|---|
| Normal | every word, cleaned | it is the opposite of a summary |
| `Structuré` | the same content as written prose | its prompt says **do not summarise**, and its band floors at 0.4 |
| `Liste` | infinitive tasks as bullets | it extracts **actions**, not substance. Band 0.1, `- ` items, and #523 decision 5 exists precisely to keep the two apart |

`Liste` is the closest and it is still the wrong shape: a user who wants the gist gets a to-do list, and a user who dictated something with no actions in it gets bullets invented out of statements.

## Why the #79 design session cut it, and what changed

`SmartModeCatalogue`'s header records the cut: *"SMS and Summary were cut in the design session … List already synthesises."* That was true of the **axis** and not of the **shape**. `Liste` synthesises into actions, in bullets, in the infinitive. The summary axis is the same compression with a different output contract, and the difference is visible to a user in one glance — which is bar B from #393, the bar Email failed and `Liste` passed outright.

## Why this is the cheapest mode in the catalogue to build

- The machinery is entirely in place: `SmartMode` row, `PolishAcceptanceContract`, the five guardrails, the fan, `SmartModeDisplayName` in both UI targets, the harness.
- **Compression is what a language model does best.** Measured on #523 round 10: the same model produces a list 2 times in 15 whatever the prompt says, but every single one of its outputs across every round compresses when asked to. This mode asks for the one behaviour that never had to be coaxed.

## Proposed contract, to be grilled

```
minimumLengthRatio: 0.1      // same floor as List — a good three-sentence summary
maximumLengthRatio: 0.6      // a summary that is 60 % of the input is not one
outputLanguage: .sameAsInput
requiresGroundedNames: true
requiresAlignedPrefix: false // it may open anywhere
overflowBehaviour: .insertRawText
```

## Bars, before any prompt is written

1. **Visibly different from `Liste` on the same dictation**, 5 of 5. Prose, no bullet, no infinitive.
2. **Zero invented facts, figures, dates or names.** #414's example-copying trap applies: every worked example off-domain, no person named.
[...]

PR #588, opened 2026-09-21
Structuré: rewrite the prompt short, one example set per Apple FM language, and refuse an invented incompleteness (refs #587)

Pivii:
refs #587 (PR 1 of decision 8). Also carries evidence for #570, #581, #585 and #523's 2026-09-20 verdict.

## What this was for

`Structuré` was failing Pierre in three ways at once, and #587 traced all three to the prompt's worked examples rather than its rules: an English dictation came back in French every time (#585, 4 of 4 on device); outputs closed on a sentence the speaker never said, `Il y avait un autre truc, mais ça m'échappe.`, copied from the prompt's own example (#581, 7 device outputs, 3 inserted); and six sentences of prose came back as six bullets (#523, 2026-09-20).

This PR rewrites the prompt short, gives it one example set per Apple FM language, adds a narrow check against the invented incompleteness sentence, and lands it on three benched rounds — 2 130 scored Mac outputs, bars committed before each round's first call.

## To test by hand — decision 12's device round

iPhone on iOS 27, this build installed, `Structuré` pinned, keyboard `fr`, transcription on auto-detect. **8 dictations:**

1. **Two long French dictations** (1 minute or more, rambling, one change of topic). Expected: French, real paragraphs, your first person kept, nothing you did not say, and no bullets unless you counted items off out loud.
2. **Two English dictations** (~20 s each). Expected: both come back **in English** and are accepted — not the raw text with the toolbar notice. This is #585's bar.
3. **Two short dictations** (1 or 2 sentences). Expected: returned nearly as said, nothing appended, and no stray last line (```` ``` ````, `---`, `-` or `...`).
4. **Two prose dictations that enumerate nothing** (6 sentences on one topic, like your 2026-09-20 recording-conditions test). Expected: **0 bullets**, paragraphs only.

**Bars: 0 language error, 0 fabricated sentence, 0 bullet on prose.** Flag every meaning distortion you see — a dropped negation, a changed name, a lost clause. Per decision 10, anything you flag opens #570's guardrail grilling, and zero takes #570 off the 2.0.0 gate.

Three extra checks, cheap while you are in there:

5. Dictate a French message that really ends on *"ah et il y avait un dernier truc, ça m'échappe"*. Expected: accepted, sentence kept (rule 7).
6. Dictate an ordinary French message with nothing about forgetting, several times. Expected: no output ever ends on a sentence about your memory. If one does, the field gets your raw words with `Structuré : n'a pas pu s'appliquer, texte inséré tel quel.` and the export shows `guardrailCheck=incompleteness`.
7. Arm `Liste` and `Message` once each. Expected: unchanged — this PR moves nothing in them (that is PR 2, #587 decision 9).

## What changed

**The prompt, landed.** 2 966 characters against 5 556: seven one-line rules, two worked examples, no counter-example block, rule 1 the language, rule 6 the list, rule 7 conditioned on the transcript and quoting no phrasing. The rules are one English text; only the two examples are translated, one set per Apple FM language (`SmartModeStructuredExamples`, 16 sets, agent-translated). The enumeration example comes first and the prose example last, which is a measurement rather than a preference (below). The largest dictation that fits the context rises from 3 972 to about 4 900 characters of speech.

**Three checks around it:**

- **`PolishIncompleteness`** (decision 6): refuses an output that reports the speaker's recall failing when the transcript carries none. Recall-failure phrasings only, in FR, EN, ES, DE, IT, PT; strict on the output, lenient on the transcript, so a spoken form rewritten into a written one is never refused. Its own `incompleteness` slug; a contract field off by default and on for `Structuré` alone.
- **`PolishPostpass.stripTrailingFenceLines`**: drops trailing lines made only of ```` ``` ````, `---`, `***`, `-`, `...` or `…`, before the guardrails. It cannot remove a word.
- **The language check treats `da`, `nb`, `no` and `sv` as one family.** Measured cause: `Jeg tror, jeg kan eksportere loggene, som de er.`, copied verbatim from a Danish transcript, reads as `nb` at 0.993, and the per-segment check (#413) refused the whole output twice. It is a whitelist of four codes, not a notion of "similar languages": Spanish/Portuguese and Simplified/Traditional stay refused, and an English chat reply on a Danish dictation stays refused.

**Also in the PR:** the per-language seam is in `develop` via #589 and only consumed here; the harness passes the transcript language, accepts a directory arm and records output language, list lines and the check's verdict; new fixtures (9 French and 5 English device dictations, plus 6 French dictations translated into 15 languages, valid for **output language only, not fidelity**); 40 new unit tests; and `docs/research/587-structured-rewrite/` with `bars.md`, the three arms, every capture and `summarise.py`, which reproduces every number below.

## The three rounds

| Arm | Worst language, refused on `check=language` | English | Bullets on prose | Fabricated incompleteness |
|---|---|---|---|---|
| The prompt in production | **29 %** (nl), 24 % (en), 17 % (da, de) | 22/33 English | **72 / 289** | 15 / 354 |
| C1 — one FR + one EN example | 28 % (da), 3 genuine translations | 33/33 | 1 / 288 | 0 |
| C2 — examples translated, list example last | 11 % (da), **0 genuine** | 33/33 | 11 / 290 | 0 |
| **C3 — landed** | **0 % in all 16 languages** | **33/33** | **5 / 290** | **0 / 356** |

Why the example order moved: with the list example last, bullets on prose went from 1 of 288 (untranslated examples) to 11 of 290; #571 measured the same position effect on the output language. Moving it off the end took it to 5 of 290.

| Bar (round 3) | Status | Evidence |
|---|---|---|
| B1a, 0 wrong-language accepted | ✅ | the only flags are `Vas-y let's go.` ×3 (your own 15 characters) and three correct Danish outputs the new family rule now accepts |
| B1b, ≤ 10 % refused on language per language | ✅ | 0 % in all 16 |
| B1c, English 100 % English | ✅ | 33/33, including your five #585 dictations |
| B2a, 0 fabricated incompleteness accepted | ✅ | 0/356. The check flags 15/354 on the old prompt, all genuine, and 0 false flags in 2 130 outputs |
| B2b, rule 7 kept 3/3 on `5-rambling` | ✅ | both arms |
| **B3a, 0 bullets on prose** | ❌ **5 / 290** | see below |
| B3b, a list on the enumerating fixture ≥ 2/3 | ❌ 0/3, both arms | declared in advance as a Mac limit; #523 measured the same |
| B4, not worse on #583's fidelity axes | ✅ | unrecalled 6→8 (inside the declared ±2), personLost 7→4, hedgeLost 5→5, hardened 1→0, fabricated 2→0 |

**The example's own content reached no output in round 3**, in either arm: the Norwegian leak of round 2 did not recur after the reorder. `PolishIncompleteness` would **not** have caught it — that sentence was about a hotel booking, not about the speaker's memory — and the check is blind to that by design.

## The one bar that fails, unclosed and recorded here

**5 of 290 prose outputs append a list that restates the paragraph above it**, two of them accepted. All five are on agent-translated fixtures, in Danish (3), Dutch (1) and Swedish (1) — synthetic text in languages nobody on this project reads, measured on a Mac that already understates the phone. It is **added content**, not a formatting preference, and no guardrail sees it: the lines reuse the speaker's own words, so `segmentOverlap` passes them.

It is **not fixed and not guarded** in this PR, deliberately. The numbers are in `docs/research/587-structured-rewrite/round3/` (`summarise.py round3`, section B3), and #587 carries a comment pointing at them. In French and English — the languages of the device round, and Pierre's — C3 produces **no** bullet on prose at all, which is why the decision was to land rather than spend a fourth round against a proxy.

## What contradicted the issue, and what I am not comfortable with

1. **Step 1 of decision 5's ladder does not hold for this mode.** English rules with a language rule were measured translating a Danish dictation into English 3 of 3. Step 2 is what works, and it is what ships.
2. **The Danish failure of round 2 was ours, not the model's.** Two refusals of correct Danish, by our own per-segment language check. That is now fixed, and its cost is unmeasured in the other direction: nothing here proves a genuinely Swedish output on a Danish dictation would be caught, because nothing produced one.
3. **14 of the 16 example sets are agent-translated** and have never been read by a native speaker. Two of the three languages that still produce the list defect are among them.
4. **The Mac barely rewrites** (median length ratio 1.00 on the French fixtures), so the fidelity numbers are partly that. The device round is the gate.
5. **B3b is out of reach on the Mac for every prompt tried**, which matches #523's own measurement. Only the device can say whether a genuine spoken enumeration becomes a list.

Verification after wiring: `swift test` **2126 tests, 0 failures** (1 skipped). `swiftlint lint --strict` **0 violations in 293 files**. `xcodebuild build` DictusApp with the keyboard on the iOS Simulator: **BUILD SUCCEEDED**. Nothing was validated on a physical iPhone.


Pivii:
## Corroborating evidence from #571 (`Résumé`), same Mac, same day — not re-measured here

The #571 agent reports, from its own runs (`docs/research/571-summary/` on `feature/571-resume-mode`):

- **Step 1 fails harder on a mode that genuinely rewrites.** Same shape as C1 (English rules, rule 1 = language, one EN + one FR example): 55 of 130 outputs came back in English across the 13 non-FR/EN Apple FM languages (de 10/10, ja 10/10, es/it/ko/pt/vi/zh ~5/10; da/nb/nl/sv ~0).
- **Example order matters.** With the English example last, 9/35 French dictations came back in English; with the French example last, 0/35 FR and 0/20 EN. **C1 puts the English example last**, and every C1 translation in this bench went to English (3× Danish `T3`, 1× French `F5`), which fits that effect.
- **Step 2 probe** (same rules, only the two examples translated into the transcript's language): de/es/ja/zh wrong-language 0/40 against 28/40 on step 1.
- Adding "never in English unless the text is English" to the language line did nothing (35/77 vs 31/78).

Read together with this PR's numbers, it supports climbing to step 2 rather than re-wording step 1. That decision remains Pierre's. These are the #571 agent's measurements; this PR did not reproduce them.

